PR #1495, opened 2015-08-01
Retrieve git commit id instead of subversion revision number

ctrlaltca:
Updated cmakelists.txt and the makedist.sh script


DarthGandalf:
I'd use something based on `git describe`, to show how different revisions compare to each other (by difference in numbers of commits since a tag)

That requires a tag though

PR #1504, opened 2015-08-02
Show WindowList With User Flag Option

Pan7:
To be able to hide the user flag(s) for channels in the window list. I.e. op or voice (@,+).


pragmaware:
I'd rename this to something like ShowUserFlagForChannelsInWindowList.

PR #1506, opened 2015-08-06
fix issue #1447 workaround buggy servers, complete downloads without corruption

### LINKED ISSUE #1447: Downloading 5 GB or larger file doesn't complete.  Stops at 100%.  End of file corrupted.
**Reported by Aaron on 24 Dec 2013 19:32:51 UTC**
I've ran into this multiple times.  The file will stop downloading at 100%, there are a few megabytes (give or take) missing from the end and Process Explorer shows KVIrc using around 24% CPU at this point.  It will sit like this for hours until I abort the download, and the file never completely downloads.  KVIrc doesn't lock up.  I can still use it while all of this is going on.  I've tried aborting and resuming multiple times and rarely I can get it to say complete but a hash check of the file shows it is corrupted, probably at the end.

I am using the Windows 64 bit build available at http://kvirc.d00p.de/

I found this site in the announce message on your IRC channel.

I am using KVIrc v4.3.1 r6366.

If I had a debug build, I might be able to help track it down to see what thread it is running circles in since it never actually crashes.

I have a pretty clean Windows 7 install and the file seems to download alright with mIRC but that client doesn't support SSL DCC transfers without a inefficient hack.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1447

un1versal:
This is a workaround buggy servers and fixes #1447

### DIFF
--- a/src/modules/dcc/DccFileTransfer.cpp
+++ b/src/modules/dcc/DccFileTransfer.cpp
@@ -403,7 +403,7 @@ void DccRecvThread::run()
 							}
 						} else {
 							// Must send the ack... the peer must close the connection
-							if(!sendAck(m_pFile->pos()))break;
+							// if(!sendAck(m_pFile->pos()))break; // comment out line - workaround allows fixing issue #1447
 						}
 
 						// now take care of short reads


pragmaware:
Not sending the ack is breaking the protocol.
We should _at least_ make this optional (default to disabled).

But probably there is a better fix for this.
Which error is reported from sendAck() when the bug happens?

pragmaware:
By the way, the "turbo" DCC extension does exactly that: it disables the acks. It's already implemented as an option.

pragmaware:
Please check if commit kvirc/KVIrc@dc6c4f5 fixes the issue.

un1versal:
> Not sending the ack is breaking the protocol.
> We should at least make this optional (default to disabled).

And is a perfectly good workaround, tested, tried and true. This other one, untested, no idea but trust your word for it.

> But probably there is a better fix for this.
> Which error is reported from sendAck() when the bug happens?

No error, it just hangs forever and a day at 100% which is not really 100% of anything as the result is corrupt missing a few 100's of bytes to completion enough to corrupt files 

> By the way, the "turbo" DCC extension does exactly that: it disables the acks. It's already implemented as an option.

By the way... doesnt work.

The only alternative was till this workaround to resume download (if supported) and redownload the remainder.

un1versal:
> Please check if commit kvirc/KVIrc@dc6c4f5 fixes the issue.

@pragmaware sorry for direct @ usage but I figured this has been in the cooler long enough to be buried and I cant test fully everyday with files large enough.

Actually after a solid testing dc6c4f5 doesnt seem to actually address the issue fully.

What happens is files larger than 5GB reach 99.97% and they dont complete 
Some actually reach 100% and are never stopped the connection is seen as active even though its been terminated.

The ones reported at 100% arent really 100% and always missing a few bytes so you end up with truncated archives or currupt files.
For both hard aborting and resuming where applicable seems to solve the hickups but is not practical. 

The patch I proposed here works simply and every single time without fail.

PR #1508, opened 2015-08-09
Appveyor: Build both 32 bit and 64 bit. Also downgrade to Qt4.

DarthGandalf:
Don't merge this yet.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,8 +8,10 @@ cache:
     - c:\enchant.7z
     - c:\spellcheck-dicts
     - c:\ProgramData\chocolatey
+    - c:\Qt-tversoft
 environment:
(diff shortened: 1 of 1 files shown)


pragmaware:
I haven't tried appveyor yet. But why downgrade to Qt4?

DarthGandalf:
Appveyor is continuous integration service for Windows. It's currently building 64bit windows installer for KVIrc with Qt5 on every push, and on every pull request.
But @wodim said that not every feature is implemented yet, e.g. https://github.com/kvirc/KVIrc/blob/d40801e287ea2cbefc8e43cfa6095f8c957b98bc/src/modules/objects/KvsObject_http.cpp#L28

pragmaware:
Ah, right, that's true.
...I'll have to look at it when back from holidays.

PR #1509, opened 2015-08-11
Add option to change the max keysize of mircryption keys

Cizzle:
Adds the uintBlowfishKeysize option (only settable through /option command) so users can choose what max keysize they allow for mircryption/blowfish keys without needing to adjust the sourcecode and compile kvirc themselves. The helptext listed when selecting "Mircryption" is also changed to explain this option (together with a whitespace-typo) and it's subsequent translations (only added as English text so translators know what needs to be translated and making the option visible for non-English users as well).


wodim:
Thank you. There is a " missing here:

https://github.com/Cizzle/KVIrc/commit/c45fedcee72106e77e066729728f9cf2f25a0688#diff-81d24cb2e89e3027ef266d92aebdb261R5081

Please fix it.

pragmaware:
I'm missing something here... why we need an user-settable limit to the blowfish key?

PR #1522, opened 2015-08-26
Change: Use new Info icon for the about Kvirc menu entry

un1versal:
~~kcs_mdihelp~~
#### Edit

Use Info icon for the menu Kvirc about entry instead -> kcs_info

I cleanup  alignment in the whitespaces in comments for `KviIconManager.cpp` and added a new icon type `Info` which for now is used specifically for this function.


pragmaware:
Hmmm.... why use "help" in the "about kvirc" entry?

un1versal:
Many softwares out there that uses icons for about menu entries such as this uses the ? or (i) perhaps it would be better to use (i) for information instead?

#### Edit:

we dont have a regular information icon.... mmmm

un1versal:
Ok pragma have a look at this change and confirm its correct, Im not experienced in c++ but seems about right the addition on the **new** info icon.

No longer using mdihelp icon now its a specific new Info icon (updated main post)

Hope this is more suitable now.

un1versal:
Thanks pragma, I take it the c++ changes were correct then...

pragmaware:
Yep :)

PR #1523, opened 2015-08-26
New - Help Graphics

un1versal:
Like title says. The ? icon matches the theme now. The Kvirc txt was just a quickfix


un1versal:
there should be a trigger not to autobuild every single PR ;)

un1versal:
Thank you.

DarthGandalf:
> there should be a trigger not to autobuild every single PR ;)

That's the point of continuous integration

un1versal:
Sure, does this actually build something and do these builds end up anywhere? If so can they be used?

Or is this like a PR check to make sure it compiles OK etc.?

Reason I mention is it seems wasteful to build each single PR such as things like this. I guess because Im not familiar with the system in place and tbh haven't done any research on it.

DarthGandalf:
Travis currently only checks compilation on linux.

AppVeyor checks compilation on windows and builds windows installer. It's available to download from AppVeyor website, if you click the link and open list of artifacts.
For pushes (not PRs) it's also uploaded to d00p's server, but currently there's no way to download that installer from d00p's server without knowing the filename. Soon it will be more available.

un1versal:
Thank you for that tidbit very useful to know.

PR #1531, opened 2015-08-30
Icons Part 2 = Work in progress fix popup icons

un1versal:
pragmaware this isn't for merging without careful review.

**Im only trying to fix the issue that the popup icons are 98% wrong with default file** while this PR fixes that Im unsure if anything breaks as a result.
The icons per functions/menu entries etc should display the correct icons with this change.

Please review changes carefully and make sure this PR fixes the popup icons and doesn't break anything in Kvirc.

### DIFF
--- a/data/defscript/popups.kvs
+++ b/data/defscript/popups.kvs
@@ -19,17 +19,17 @@ defpopup(highlight)
 		}
 	}
 
-	item($tr("Stop Highlighting (Local)","defscript")) (%:beingHlTmp && !%:beingHl)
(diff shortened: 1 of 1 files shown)


un1versal:
Nice thx.

Is there a way to correct the text icons associate with smilies etc? I see there's a user texticons.kvc but I dont see this in here.

pragmaware:
data/config/texticons.kvc

un1versal:
Thx Ill submit a fixed one

PR #1537, opened 2015-08-30
fix texticons

un1versal:
On top of this file it says.

```
# Default configuration of texticons.
# If you add any entry then please increase this number both here
# and in kvi_texticonmanager.cpp (they MUST match)
```

Cant find any kvi_texticonmanager.cpp but I did find a libkvitexticons.cpp however couldnt find any relation to the number increase there.

Is this no longer the case? And if so then can this notice be removed from this file?


pragmaware:
That should read:

...the please also increase the value of TEXTICONMANAGER_CURRENT_CONFIG_UPDATE in src/kvirc/kernel/KviTextIconManager.h (they must match)

un1versal:
Done updated the note on file and bumped as it requires. Please check its all OK.

pragmaware:
Great :+1:

PR #1538, opened 2015-08-31
Notifier - cleanup part 1

un1versal:
I would like to figure out what notifier graphics are still used aside for these being moved here some don't seem to be anymore in fact most seem to be ignored.

Does anyone know?

Also

When committed the new icons, it was just a copy and paste from my internal themes and as such some graphics were in wrong directories. mostly some of the notifier graphics I had redone.

Hence this is Part 1.

I will need to finish the notifier graphics but Ive no idea which still apply (aside the obvious ones) or which can be removed so if anyone can help there it would be great. I wouldn't like to wast time redoing graphics that are no longer needed.


pragmaware:
You can pop-up the notifier with "/notifier.message test" to see what is being used and what not.

pragmaware:
You can also take a look at the sources in src/modules/notifier/. Grep for *.png, for instance.

un1versal:
Thx, while /notifier.message test is a good tip doesnt really tell which actual png are used aside the ones I updated the other I grep *.png in that dir but did nothing in my test just hangs there.

Im looking to know which pngs are obsolete vs which are still used in this dialog. Again aside from the ones I committed

PR #1546, opened 2015-09-02
[kvirc] improvements to some icons

un1versal:
Just a few tweaks to some icons to better suit usage.

Are icq icons even used anymore these days?


pragmaware:
Don't think so..but the icons are numbered so for compatibility I'll leave them there.

un1versal:
I wouldnt remove them either and Im all for keeping, ideally leave as is and later on if there's a need to add 4 icons for something we could reuse these while also renaming them that way you dont screw with the number of icons in the grid which would break all popups icons again.

btw how do you find the new set so far?

pragmaware:
It's cool.
Keep up the good work!

Once you've created a complete and organic set we can talk about "this or that icon that could be better". But for now it's an extremely valuable work. Go ahead :)

un1versal:
Contributing to a project I and others in the family use everyday is only a small way to say thank you for KVIrc, so  from my part you are very welcome.

It also great this is now in Git.

I have a great deal more to add in a few other areas, but just needs polishing up a bit before Im happy to push it this way.

PR #1549, opened 2015-09-06
Various SVN -> Git changes and general README/docs updates

un1versal:
Please review.

Here I changed many SVN and updated say Git instead.

Corrected typos that I found on this voyage. (not comprehensive, these docs all need a once over)

Also updated the INSTALL doc so that its slightly easier to follow, contains all libraries that I had to hunt in debian packaging files and by trial and error since a great deal was missing.

Im not quite finished yet here but further changes can go in later.

### DIFF
--- a/doc/FAQ
+++ b/doc/FAQ
@@ -6,7 +6,7 @@
 
 2. COMPILATION
 
-2.1 I can't compile kvirc from svn, what's wrong?
+2.1 I can't compile KVIrc from Git, what's wrong?
 2.2 Internal Compiler Errors ?
 
 
@@ -27,7 +27,7 @@
 4.6 /me doesn't support the '$' character, what's wrong ?
 4.7 /me doesn't support the '"' (double quote) character, what's wrong ?
 4.8 /me can't contain the ';' character, what's wrong ?
-4.9 I try to do a /whois on nick blah\ but i get the reply for blah (without the slash)
+4.9 I try to do a /whois on nick blah\ but I get the reply for blah (without the slash)
 4.10 What is AVATAR protocol ? Is there any documentation about it ?
 4.11 The log files seem to be saved randomly, when I open a log
      I can see only a part of the text that actually should be there
@@ -63,7 +63,7 @@
 7. CRASHES
 
 7.1 What the heck is a gdb stack trace and how do I get it ?
-7.2 KVIrc hangs up, but doesn't crash. How do i report debug information ?
+7.2 KVIrc hangs up, but doesn't crash. How do I report debug information ?
 7.3 KVIrc exits by itself or simply "disappears" without any apparent reason.
 7.4 KVIrc dies when using alias(identify){ identify password; }
 
@@ -96,22 +96,23 @@
    the KVIrc sources with the KDE library changes we choose not to do it.
    If you wish to contribute some code that better integrates KVIrc with
    KDE you're welcome to do so: we will happily accept your patches as long
-   as you promise to mantain them :)
+   as you promise to maintain them :)
 
 
 --------------------------------------------------------------------------------
 2. COMPILATION
 --------------------------------------------------------------------------------
 
-2.1 I can't compile kvirc from svn, what's wrong?
+2.1 I can't compile KVIrc from Git, what's wrong?
 
    Have you checked the requirements ?
    *Carefully* read the INSTALL file that comes with the distribution:
-   it contains the information to solve 95% of the problems that
+   it contains the information which solves 95% of the problems that
    you may encounter. You may also try digging in our mailing list
    archive: many problems have been solved there.
-   If you still can't get it to work please either subscribe to the KVIrc
-   mailing list (see www.kvirc.net) and report the problem there or join #kvirc
+   If you still can't get it to work, please either subscribe to the KVIrc
+   mailing list (see www.kvirc.net) and report the problem there. Alternatively
+   open a issue at https://github.com/kvirc/KVIrc/issues or join #kvirc
    on Freenode and ask the channel members for a hint.
 
 
@@ -129,7 +130,7 @@
       out 5-6 of them daily. Usually it is solved by simply restarting the "make" command.
       If it doesn't go away or if it gets really frequent and annoying you might try to upgrade the compiler.
    2) A memory corruption: this is really frequent: one of your memory banks is spitting random bytes
-      sometimes. Usually the problems come out under heavy system load (kernel, KDE or kvirc compilation for example).
+      sometimes. Usually the problems come out under heavy system load (kernel, KDE or KVIrc compilation for example).
       You can check for memory corruption by trying different RAM modules (or removing them one at a time to discover
       the broken bank). You can also try a program called memtest and a kernel
       patch that can force the broken memory banks to be ignored by the kernel (allocated permanently to a dummy kernel task).
@@ -160,12 +161,12 @@
 4. IRC
 -------------------------------------------------------------------------------
 
-4.1 How do i join a channel ?
+4.1 How do I join a channel ?
 
   Type "/join #channelname"
 
 
-4.2 How do i autojoin channels?
+4.2 How do I autojoin channels?
 
   There are many ways to do it.
   For example:
@@ -180,7 +181,7 @@
   Now click OK and you're done: you have just created a handler for an event.
   Reconnect to a server and KVIrc will autojoin the channels that you have selected.
   Another option (nicer, if you use several servers) is to go to the servers options
-  page and click the button labeled 'Advanced' on the entry for the single server.
+  page and click the button labelled 'Advanced' on the entry for the single server.
   There is a server-specific "On connect execute..." event.
 
 
@@ -192,7 +193,7 @@
   "Advanced...". In the dialog that appears there is a NickServ tab
   on that you can add your identification rules. Adding a rule
   should be straightforward (if you stop the mouse over an item you will likely get a small
-  tool tip with the explaination). Each rule is composed of:
+  tool tip with the explanation). Each rule is composed of:
   - the nickname that is being identified: the rule will be applied only
       when your current nickname corresponds to this one.
   - a nick server user mask that must be matched by the NickServ user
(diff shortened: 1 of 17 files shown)


ctrlaltca:
:+1:

doc/FAQ
```diff
@@ -111,7 +111,8 @@
    you may encounter. You may also try digging in our mailing list
    archive: many problems have been solved there.
    If you still can't get it to work please either subscribe to the KVIrc
-   mailing list (see www.kvirc.net) and report the problem there or join #kvirc
+   mailing list (see www.kvirc.net) and report the problem there
```
DarthGandalf: `.` missing

doc/INSTALL
```diff
@@ -1,244 +1,215 @@
...
+  are distributions which are Ubuntu derivatives.
+    
+  Note: If anyone could find the time to test KVIrc with ICC please let us know!)
+
+-------------------------------------------------------------------------------
+  2. Dependencies and requirements:
+-------------------------------------------------------------------------------
+
+  KVIrc will require the development packages for its list of dependencies.
+  Dependencies marked as (Preferred) are a suggestion to improve your user experience.
```
DarthGandalf: And (optional) don't improve user experience? :)

doc/INSTALL
```diff
@@ -1,244 +1,215 @@
...
+    $ export QTDIR=/usr/lib/qt4
 
-    If you have found this document in a binary distribution then
-    KVIrc has been probably already installed by your favorite package
-    manager and maybe something is not working as expected. In this case
-    this document and the accompanying FAQ can help you in guessing
-    what's wrong.
+  For x86_64 e.g.
+  
+    $ export QTDIR=/usr/lib/qt4
```
DarthGandalf: which is exactly the same path as 2 lines above?

doc/INSTALL
```diff
@@ -1,244 +1,215 @@
...
+  uses qmake itself to determine the Qt4 library dir.
+  You can set your $PATH variable to have CMake using your preferred Qtx.x version:
 
-    If you have found this document in a source distribution or you
-    have downloaded it by using SVN then well...
-    this is a standard INSTALL file :)
+    $ PATH=/opt/qtx.x/bin/:$PATH cmake ..
+    
+  Note: The trailing dots are required or CMake will NOT create the right
+  environment.
```
DarthGandalf: They are important _if_ you do `mkdir build; cd build` in source tree.
Otherwise, it'll do something weird (most probably, just not find CMakeLists.txt)

doc/INSTALL
```diff
@@ -1,244 +1,215 @@
...
 
-###############################################################################
-# 2. Minimal requirements:
-###############################################################################
+  Note: Compiling with KDE4 support enabled will force KVIrc to
+  link against the Qt4 version that KDE4 is using. This may lead to problems
+  in linking. Disabling KDE4 support is the only known solution at the moment.
+  In this case override the Qt libraries used when running KVIrc:
```
DarthGandalf: I think this section belongs above, in discussion of Qt4 vs Qt5.
Or below, into KDE section

doc/INSTALL
```diff
@@ -1,244 +1,215 @@
...
 
-    This is usually included in nowadays distros.
-    To check which CMake you have, simply run:
+    $ msgfmt --version
 
-    # cmake --version
+  KVIrc will not complain if the command above is missing. It will just skip
+  the creation of the translation files. See the cmake configuration output.
+  If the command above fails, then you need to install the gettext package if
+  you want any other language than English.
```
DarthGandalf: 10 lines above you already mentioned that, no need to repeat

doc/INSTALL
```diff
@@ -246,353 +217,375 @@ General notes for the installation of the 4.2 release of KVIrc
     to "/usr/lib/libgsm.a", you might create the shared library manually
```
DarthGandalf, line 221: I don't think it's very good idea to modify /usr/lib by hand for this.
1. This won't work if the library was compiled without `-fPIE`.
2. Files in /usr/lib should really be managed by package manager of the distro. Whenever package containing libgsm.a upgrades, the manually created libgsm.so stays old.
3. Distro can provide a better way to install libgsm.so. E.g. in Gentoo many packages have `static-libs` USE-flag which tells the package to install not only `.so`, but also `.a` (which is kinda the reverse, but the point still stands)

doc/INSTALL
```diff
@@ -246,353 +217,375 @@ General notes for the installation of the 4.2 release of KVIrc
...
+
+  Since KVIrc 4.2 uses CMake as build system, we have to use it to create
+  the Makefile's rules for make program.
+  To do it, we encourage the "out-of-source" building: build all files
+  without dirtying the sources directory.
+
+    $ cd /home/setup
+    $ mkdir release
+    $ cd release
+    $ cmake [your options] ..
```
DarthGandalf: There was a different cmake line above...
Which one should be used?

That one can be just `export QTDIR=...` instead.

doc/INSTALL
```diff
@@ -246,353 +217,375 @@ General notes for the installation of the 4.2 release of KVIrc
...
+  -DCMAKE_INSTALL_PREFIX=PATH 0 (default: a system-dependant directory)
+    This will install all stuff under PATH directory.
+
+  -DLIB_SUFFIX=SUFFIX - (default: empty)
+    This will install libkvilib and KVIrc modules to
+    ${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}
+    directory. It defaults to an empty suffix.
+    Mainly used for x86_64 architectures.
+
+  -DMANDIR=DIRECTORY=  (default: $CMAKE_INSTALL_PREFIX/share/man)
```
DarthGandalf: Are you sure about `=` at the end?

doc/addon.howto.txt
```diff
@@ -304,14 +304,14 @@ The name of the classes refer to the ones described above.
   collisions with other scripts and addons.
 
 - Remember that your addon is going to be installed on different platforms (at
-  least linux, macosx and windows based). The poor windows' notepad has serious problems
+  least Linux, Macosx and Windows based). The poor windows' notepad has serious problems
```
DarthGandalf: I think it's Mac OS X

doc/addon.howto.txt
```diff
@@ -304,14 +304,14 @@ The name of the classes refer to the ones described above.
...
   with reading text files that contain only linefeeds as line separators. Keep it in mind...
 
 
 9. WHERE TO START
 
 It is a good idea to start on the KVIrc web site. There are surely several addons to
-look at. Pick one that seems simple and analyze its layout and code (wow... the free
+look at. Pick one that seems simple and analyse its layout and code (wow... the free
```
DarthGandalf: analyze is correct in American English, analyse is for British English

doc/INSTALL
un1versal: That information already was here, I just added "See cmake configuration output" Ill look at this again and improve it.

doc/FAQ
un1versal: This file is a mess, I need to check it properly, thers more than just a missing . :)

doc/INSTALL
un1versal: yes, of course it does Ill make it clearer.

doc/INSTALL
un1versal, line 221: No idea, I didnt touch this information, but I agree in principle people should either install distro provided libs or compile own libs properly should their distro not provided the prepackaged or specific version, whatever the case may be.

Ill try to make this clear.

doc/INSTALL
un1versal: What?

doc/INSTALL
un1versal: right.

doc/INSTALL
DarthGandalf: Line 122: `$ PATH=/opt/qtx.x/bin/:$PATH cmake ..`
I meant `export PATH=..`, without calling cmake at that line

doc/addon.howto.txt
un1versal: American/Canadian vs European English :) for a perfect nitpick.

doc/addon.howto.txt
un1versal: I think its rubbish. hahaha Apple sucks.

doc/addon.howto.txt
wodim: This is an European project so we should enforce British spelling.

doc/addon.howto.txt
wodim: This is not the place for childish remarks. Those should be at the IRC channel.

doc/INSTALL
un1versal: Read what it says.

  If you have multiple versions of Qtx.x installed and want to use a specific version,
  **exporting QTDIR won't suffice,** since the CMake module searching for Qtx.x
  uses qmake itself to determine the Qtx library dir.
  You can set your $PATH variable to have CMake using your preferred Qtx.x version:

   $ PATH=/opt/qtx.x/bin/:$PATH cmake ..

So essentially if you have more than one version of the same QT4.x then you need both export if your distro doesn't configure packages properly and you need    $ PATH=/opt/qtx.x/bin/:$PATH cmake .. to make sure the version you want is used if it differs from the exported one.

no? 

Again this is information that already existed and I didnt really touch it.aside from reformating.

doc/INSTALL
un1versal, line 221: Ya no idea what to do here, Ive just added your remarks as a note. hopefully people know how to use Linux and cover those situations else its a distro issue not a KVIrc issue anyway.

doc/INSTALL
DarthGandalf: Right now `cmake ..` is mentioned in 2 different sections. And the first one makes wild guess that user did read the second section first, and created a subdirectory.

I propose replacing `PATH=/opt/qtx.x/bin/:$PATH cmake ..` with `export PATH=/opt/qtx.x/bin/:$PATH` and telling `cmake ..` only here.

I'm not arguing about QTDIR or PATH or whatever, but about invocation of cmake.

un1versal: That info has always been like that mentioned like that exactly.

And as I understand it you need both since they cover very specific scenarios one that export blablablah wont work when you have two destinct Qt4 versions. Hell I dont understand exactly your comment give what the readme explains the reason for those entries existing.

export path is not enough (according to the explanation) if you want to use Qt4.8.5 and also have qt4.8.6 installed, thats what 
``PATH=/opt/qtx.x/bin/:$PATH cmake ..` is suppose to cover since export path wont cut it

If you mean condensing the usage down to one section so that its not repeated Ill have another look.

PR #1560, opened 2015-09-09
[coresmall] add Warning Icon and update notaway Icon

un1versal:
This is a two in one.
1) Added warning Icon for this specific menu entry

capture

2) The notaway icon in status bar needed something more inline with action in intends to represent. Could be better I suppose.


src/kvirc/kernel/KviIconManager.cpp
```diff
@@ -401,6 +401,7 @@ static const char * g_szIconNames[KviIconManager::IconCount] =
 	"sexircop",            // 315
 	"memoserv",            // 316
 	"info"                 // 317
```
pragmaware: A comma is missing here

un1versal:
done

PR #1561, opened 2015-09-09
Fix #1559 plus improve Option menu icon assignments to more suitable icons.

### LINKED ISSUE #1559: Wrong icon in settings menu
capture

Should be https://github.com/kvirc/KVIrc/blob/master/data/pics/coresmall/kcs_raw.png

I think I found the issue PR to fix shortly for review

un1versal:
This should fix #1559

Please review its untested here atm

Updated existing icon instead.

Edit

Also Fixed up some more icons in KVIrc Settings/Options menu that werent quite right and are better suited to these.


pragmaware:
You got it right, though it's still an arbitrary change. Maybe it's better if we add a new icon for the raw event editor window?

un1versal:
Thx, yes, I think Ill modify the existing icon to suit, thats seems more acceptable, and push the fix to here shortly.

un1versal:
Done please review icon changes

uNiversaI:
Actually hold a sec... one chnage too many.

uNiversaI:
OK, done for real this time.

pragmaware:
Wait. RawEvent is actually used in the RawEditorWindow.cpp.
We need all three: 
- KviIconManager::Raw which is used to display raw text from server, 
- KviIconManager::RawEvent which is used for raw event handler entries in RawEditorWindow.cpp
   (and which was previously used also in the menubar)
- a hypotetical KviIconManager::RawEventEditor which may be used for the menubar

un1versal:
> KviIconManager::Raw which is used to display raw text from server, 

We have that, :heavy_check_mark: 

Well... The **kvi_bigicon_raweditor.png**  which is in the toolbar the equivalent for coresmall **kcs_raw.png** 

That is used both in KviIconManager::Raw and the which is used to display raw text from server and 
For the menubar raw editor + toolbar raw editor

> KviIconManager::RawEvent (RawEditorWindow.cpp)

OK, this is this icon here right? :heavy_check_mark:  Where is that in the UI? Currently its in menu bar (settings menu) only which is in #1559

> a hypotetical KviIconManager::RawEventEditor

I can make this 3rd type but I dont know where that is supposed to be displayed, and I need some context in order to visualize if this is replacing I dont think I ever seen this.

un1versal:
Yep I see what you mean now, I actually looked properly, so yes, for here we need a new type.

un1versal:
Should be good to go

un1versal:
This is how it looks now

capture
capture1
capture2

Merging later on if there's no objections.

un1versal:
Merging now, I have more waiting on the burner (and want to get it gone before weekend). I rather avoid conflict resolution later on, if anything needs addressing, we can always do it after.

PR #1568, opened 2015-09-10
[Editors] Menu/Toolbar - Improve icon assignments

un1versal:
This will hopefully bring the menu bar and toolbar for Editors all inline.

Had to add a new icon type `KviIconManager::EventEditor` else the editors would looks off also assign it so its not forgotten.

It would be fantastic if anyone would volunteer to help with news / blog entry a sort of notice about the visual improvements and a sort of request to theme developers to get their themes inline with 4.3.2.

I presume at some stage some new release for stable will be made and Ide like to get the theme (icons/assignments) core ready sooner rather than later.

~~**This is a work in progress** Not for merging yet, as it turns out need to add editors (menubar) for all those icons here and this will serve to point out any errors on the code part.~~

Should be done soon.


un1versal:
Please review

Issue #1563, opened 2015-09-10
Optional "Set transparent or no background" when "Smart" Color nicknames is enabled


Zybyr:
Hi Guys,

The background colors behind nicknames should be an option when "Smart" color nicknames checkbox is enabled, In my opinion it does not look good at all, and also difficult to read. 

In addition, an option to have colored nicks in the list of users of the channel would be really nice to have. The nick colors in the list of channel users column, should match the color in the chat of the same user when users write to the channel. I think hexchat would be a an example of what I mean. 

![background](https://cloud.githubusercontent.com/assets/11053449/9778485/e3354962-573e-11e5-80c4-6034c4a83e8a.png)

un1versal:
> The background colors behind nicknames should be an option when "Smart" color nicknames checkbox is enabled, In my opinion it does not look good at all, and also difficult to read.

+1 here. I disable this feature on my themes precisely for this reason no matter if you have light or dark bg themes the issue is same.

> In addition, an option to have colored nicks in the list of users of the channel would be really nice to have. The nick colors in the list of channel users column, should match the color in the chat of the same user when users write to the channel. I think hexchat would be a an example of what I mean. 

That is already possible you can assign different colors to each user groups but only one color per each group is supported, So chan admin have one color, channel owners another, normal users another etc etc. and this works great in channels where theres varied user groups not so much where only 1 user type.

If you mean smart colors in user lists as well, I suppose this could be nice but to be honest I dont like the smart colors much so I cant really say much here.

And now Im really off to bed.

un1versal:
Duplicated of #353 (seems this is an old issue).

After reading that ticket I vote for closing that ticket and keeping this open, this one has a better proposed solution.

Guys?

pragmaware:
Disabling the background introduces a problem: what if the nickname color has bad contrast to the current user's background?

un1versal:
Same color on same color would be bad that's sure, Thinking aloud here, isn't that something which can be avoided? As is doesnt seem ideal.

pragmaware:
Well.. I remember writing that code the way it is explicitly to avoid the background problem.
The colors are in a table and are choosen to have reasonably good contrast (between fg and bg).

There is no safe way to exclude a foreground color (or a set of colors) that is too similar to the window background. .. or at least I couldn't think of one.

un1versal:
That screenshot doesnt make the issue justice, it gets quite unreadable and hurts brain :D

[...]

PR #1569, opened 2015-09-11
2015 - Update internal infernal themes - they're hot off the press

un1versal:
This updates internal themes to 4.3.2.
- Use the default core icons
- Color schemes have been slightly updated to improve readability
- Tweaked message colors
- Added theme icons
- Tweaked Aria's background text.
- Disabled smart nickname colors until https://github.com/kvirc/KVIrc/issues/1563 has some solution.
  This feature makes readability an issue and one shouldn't have to struggle. Usability takes priority over fancy features that dont behave well :smile: 

capture

md

aria

Could always improve in future a tweak here and there.


swarfega:
Looking good.

pragmaware:
Smart nickname colors should be on in the fancy theme.... it's fancy :)
#1563 is opinion based. Disabling the background introduces a problem that can't easily be solved.

un1versal:
I can enable it back no big deal, but (if Im honest) #1563 is a usability problem and the main reason why I started this themes icons thing in first place :) because built in themes were extremely hard on the more visually impaired users, like me, and before I could actually use KVIrc as replacement IRC client I had to make own theme

pragmaware:
Well... making the themes more readable is a good thing.
Though I think that the fancy theme should show all the available "fanciness" :D
One can still disable the smart nickname coloring after installing the fancy theme...

un1versal:
Do you think all users are aware they can disable it or even want to be that involved in IRC configuration outside the requirements to connect?

Ill enabled it no worries.

also while we are here...

Can we add more themes in by default so that they can be selected in theme manager? its like less that 50KB each. Not at setup time, thats for these here I was thinking simple light dark and black with softer color schemes.

un1versal:
This built OK (until I resync) and Ive enabled smart colors in Aria all the trims for fancy stuffs.

Ill merge now so that these things can be tested and we can take it from there if anything further needs adjusting.

pragmaware:
:+1:

un1versal:
Small problem, in Linux the 1.0.0 themes arent removed and you cant unistall KVIrc since its compiled.

pragmaware:
You're right. This is a well known limitation of source based installs.
It would be also a bad practice to try to remove the files programmatically as they may have been changed by the user (beside the fact that kvirc is usually installed in a system-wide directory but started with non-root privileges).

PR #1585, opened 2015-09-14
[ListWindow] use getSmallIcon() instead and fix assignments

un1versal:
Please review.

Im not sure if there's a reason why in ListWindow.cpp this toolbar is using `getBigIcon()` intead of `getSmallIcon()` 

Since its 16x16 only and `m_pStopListDownloadButton` is using `getSmallIcon()` I decided this change makes sense.
Let me know if this is right and merge if it is. Else Ill only assing the right icon to `m_pStopListDownloadButton` and discard the other changes.

### DIFF
--- a/src/kvirc/kernel/KviDefaultScript.cpp
+++ b/src/kvirc/kernel/KviDefaultScript.cpp
@@ -474,7 +474,7 @@ KviDefaultScriptDialog::KviDefaultScriptDialog()
 
 	m_pAdvanced->setLayout(pAdvLayout);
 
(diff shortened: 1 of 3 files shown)


un1versal:
The build failure in win32 is unrelated to the changes.

pragmaware:
It's probably simply because we had only big save and folder icons at the time that code was written.
Your change looks fine.

un1versal:
OK thanks just checking I haven't broken anything :)

With this PR assignments should now be all but complete.

If thers something Ive missed and you notice let me know.

Merging this now.

PR #1593, opened 2015-09-15
Add 3 basic simple themes for selection in theme changer with default installs

un1versal:
Adding 3 of my own themes in the Input/Output series, distributed by default with KVIrc

This does not in any way affect the internal default themes nor does it show up at setup time.

It occupies 132KB on disk :) this crazy bloatware eh? :grimacing: 

Now these are definitely not compatible or they are not designed with with "smart color nicks" in mind and never will be, they are designed for providing a simple and effective IRC experience without any fluff.

Personally I prefer the Dark and Black themes and only providing a light scheme for completeness of Series.

dark
black
light


un1versal:
I need to have a look at something in the light theme before merging.

PR #1596, opened 2015-09-15
Add tip of the day: Theme changing.

un1versal:
Is it wrong to assume translations will be handled from the English language updates?


pragmaware:
Tip translations have their own separate configuration files. If the translators are watching, they will translate the additional tip too.

un1versal:
Fingers crossed they are :)

PR #1616, opened 2015-09-20
Update numerics 344 and 345

staticfox:
`Currently 344 has no use, as the last time it was referenced
was in irc2.11.2 (Which hasn't been updated in about 23 years).
As for 345, it has since been changed to RPL_INVITED for GameSurge
as well as QuakeNet. I also believe InspIRCD at one point used it as
well.`

Documentations sited:
https://www.alien.net.au/irc/irc2numerics.html
http://www.rubydoc.info/gems/ircsupport/IRCSupport/Numerics
http://cpansearch.perl.org/src/HINRIK/IRC-Utils-0.12/lib/IRC/Utils.pm
http://www.irc.org/ftp/irc/server/irc2.11.2p3/common/numeric_def.h
https://wiki.inspircd.org/List_Of_Numerics

GameSurge's IRCd from Socket Spy

```
[9:15:43 AM] :*.GameSurge.net 345 #blindsight blind|web2 blind|web1 :blind|web2 has been invited by blind|web1
```

It is disabled by default on QuakeNet's IRC Network, however, here are references to their current source code
https://hg.quakenet.org/snircd/file/a93b82f3e4a5/include/numeric.h#l264
https://hg.quakenet.org/snircd/file/a93b82f3e4a5/ircd/m_invite.c#l295

### DIFF
--- a/po/core/kvirc/kvirc.pot
+++ b/po/core/kvirc/kvirc.pot
@@ -7160,10 +7160,6 @@ msgstr ""
 msgid "protected/admin list"
 msgstr ""
 
-#: src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp:797
-msgid "reop list"
-msgstr ""
-
 #: src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp:798
 msgid "spam filter list"
 msgstr ""
--- a/po/core/kvirc/kvirc.pot
+++ b/po/core/kvirc/kvirc.pot
@@ -7160,6 +7160,10 @@ msgstr ""
 msgid "protected/admin list"
 msgstr ""
 
+#: src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp:797
+msgid "reop list"
+msgstr ""
+
 #: src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp:798
 msgid "spam filter list"
 msgstr ""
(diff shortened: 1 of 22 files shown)


pragmaware:
Hm... it seems that IrcNet still uses 344/345 for the "reop list" +R channel mode
(CHANMODES=beIR,k,l,imnpstaqr)
An idea could be to:
- keep the old code
- rename parseNumericEndOfReopList to parseNumericEndOfReopListOrInvited (or something like this)
- add a branch inside parseNumericEndOfReopListOrInvited to handle either RPL_ENDOFREOPLIST or RPL_INVITED depending on the supported channel list modes.

staticfox:
Incidentally, IRCu uses +R for channel registered. To make life a bit easier, I've just made it check against IRCd version since the numeric for RPL_INVITED is only within IRCu and Snircd.

un1versal:
I think `msgid "reop list"` needs to be `msgid "Reop list"`

staticfox:
Hmm I kept it the way it was before and it seems fine. Unless you want me to capitalize in the po files and the numeric handler anyways.

un1versal:
If this string is the beginning of a sentence or its standalone string the it must start with a capital letter.

The whole language strings need reviewing IMO. Grammar is murdered fast and loose in KVIrc shouldn't we all make some effort to improve this?

staticfox:
It isn't, here's an example from /invite on GameSurge (Since I'm using a release client, I don't have this patch)

```
[1:38:16 PM] End of channel reop list for Gwen
```

If you would like, I could take a look at the language strings.

un1versal:
right combination strings then?

See https://github.com/blindirc/KVIrc/commit/7766c5102492010f2be198434a61bbb4547bbf73

many of the translations then are wrong starting with a capital letter.

Im not completely familiar with how strings are used in KVIrc, if they are standalone or combination strings, but seems to me this one is wrong in either main string or translation.

We can leave the string review to another more specific PR and if you are willing to make some dent into that, you also get my +1 :grinning:

PR #1625, opened 2015-09-23
Show spell checker suggestions in a submenu of popup menu.

DarthGandalf:


un1versal:
+1

Will this make the suggestions immediately available or do you still have to fiddle about?

DarthGandalf:
You still need to select the wanted fix from the list...

un1versal:
Yes sure, what I meant was, will this list be available immediatly or is there fiddling required? 
Currently you need to right click over error, once or twice sometimes even three times before the spellchecker suggestion(s) is available for selection. Otherwise its disabled.

DarthGandalf:
When do you need to click it 3 times? o_O
kvirc 4 3 2 aria - _217
kvirc 4 3 2 aria - _218

un1versal:
Ah I figure it out, if the cursor is not over the word error it doesn't become active.

**Not active** cursor by end of OK
capture1

capture

**Active** cursor at end of word or middle of word
capture2

capture3

So o_O indeed.

pragmaware:
Yes, the suggestions are for the word at cursor.
You can activate the suggestions popup with CTRL+G too.

un1versal:
> Yes, the suggestions are for the word at cursor.
> You can activate the suggestions popup with CTRL+G too.

A more intuitive manner, would be perhaps to be able to activate the suggestions by right clicking over the underlined word.

Issue #1626, opened 2015-09-26
New icons are blurry at Medium size (22x22)


DarthGandalf:
16:
![selection_220](https://cloud.githubusercontent.com/assets/325092/10116401/1d59013e-642c-11e5-8850-710775e50895.png)

22:
![selection_221](https://cloud.githubusercontent.com/assets/325092/10116402/1fd32110-642c-11e5-80f8-33bdfa472895.png)

32:
![selection_222](https://cloud.githubusercontent.com/assets/325092/10116404/21d1f91e-642c-11e5-8756-f0721c57577a.png)

48:
![selection_223](https://cloud.githubusercontent.com/assets/325092/10116405/23e6593e-642c-11e5-9da9-d1c272015ba8.png)

Also, why at 16 some icons are different?
At 48 icons look the same as at 32, but centered.

un1versal:
They are resized by program from the original at 32x32 so there's nothing to be done here from graphics side. Old design would have had same issue exactly.

If KVIrc can use a better rescaling algorithm like bicubic or better then this may improve something,

un1versal:
So Ill reassign to whoever can change or add a better rescalling algorithm Cheeky :)

DarthGandalf:
Hm. I see.
Old icons didn't feel as blurry at 22. Well, I've noticed it only with new icons.
Probably because old icons had less details...

un1versal:
I think the scaling is handled by Qt, no idea even, its just a guess since I cant find any scaling code

@pragmaware probably knows

Reading http://doc.qt.io/qt-5/qimage.html points to http://doc.qt.io/qt-5/qimage.html#transformed, I still see no specific algorithms being used.

DarthGandalf:
How does Qt rescale a rectangle with a key to a spiral fingerprint on 16x16?

un1versal:
No idea, probably the identity icon is using who, Ill look into that
sorry fat fingers closed by mistake.

un1versal:
> How does Qt rescale a rectangle with a key to a spiral fingerprint on 16x16?

Found the issue.

This is a general problem I solved with a few other icons already, the big icon filename and small icon filename dont match at all so end up causing issues like this.

Will submit a fix for this one soon enough.

uNiversaI:
mmm the whole 16x16 toolbar has this same issue and needs a little more work than 1 icon fix only but will fix nonetheless.

un1versal:
> Also, why at 16 some icons are different?

[...]

PR #1627, opened 2015-09-26
Improve various icons, mainly fix some comments in #1626

un1versal:
@DarthGandalf this PR addresses the 16x16 toolbar issues mentioned in https://github.com/kvirc/KVIrc/issues/1626

The 16x16 toolbar was a mix between rescaled and non rescaled icons this fixes that hopefully.

The naming conventions between large and small icons is still a problem, and tbh thats not fully addressed and Im trying to gain courage to look into that HUGE mess.


Stanzilla:
maybe start with folders? One for 16px and one for 32px, would be a good start to organize?

un1versal:
> maybe start with folders? One for 16px and one for 32px, would be a good start to organize?

@Stanzilla
I dont think you understand the issue at all. 

It already is separate so nothing to do there, you have all 16x16 in coresmall folder and all 32x32 outside, its always been like that afaik or used KVIrc

Naming convention is where a big icon name vs small icon name is so different that you cannot associate their function or use by looking at their filenames.

A secondary issue with naming is where icons that are used in a specific function are named in a manner where they are meaningless to that function or any function used in program currently and for a long time now.

Stanzilla:
Ah ok, I see.

PR #1631, opened 2015-09-27
Random collection of typos grammar and string naming consistency fixes

un1versal:
Please review


po/core/kvirc/kvirc.pot
```diff
@@ -1834,19 +1834,19 @@ msgstr ""
 
 #: src/modules/ident/libkviident.cpp:403
 #: src/modules/ident/libkviident.cpp:423
-msgid "Identd accepting connection"
+msgid "Ident accepting connection"
```
Dessa: starting here, it should really be identd or ident service as above, but that might be too long?

un1versal: No idea, why is it that we are mixing identd and ident to begin with. Its not a good move imo to mix terminology

pragmaware:
Hm.. we should be very careful with string changes.
In order to avoid breaking translations they should be done both in the source code and in all the *.po files. For instance, the "Success ?" -> "Success?" change is in all the *.po files but I couldn't find the source code change in the diff...

un1versal:
> In order to avoid breaking translations they should be done both in the source code and in all the *.po files. For instance, the "Success ?" -> "Success?" change is in all the *.po files but I couldn't find the source code change in the diff...

Yes, I didnt do the source code, and actually dont understand why the source code doesn't have a string number instead and you only mess with text in the pot/po files (but I am not going to do that at all)

I will slowly add the missing source code changes so you can consider this PR a Work in progress.

**EDIIT**
Apparently the source code changes that matter the po and pot files are generated.by `make messages-update`

un1versal:
@pragmaware @Dessa please have another look, I think I caught all core strings + po strings.

Also fixed Dessas and Stanzillas comments and then some, just not squashed all the way else its a huge bundle.

pragmaware:
Well... when you do a grammar/syntax fix in a string you can usually do it by search and replace.
Then do it on both all the sources and all the po files at the same time...

DarthGandalf:
There is a caveat.
If source contained inconsistency, the translated string probably has the
same inconsistency.
So it's better for translators to give a chance to check the new
translation without "search and replace"y marking all translation as
correct.

2015-09-28 12:22 GMT+01:00 Szymon Tomasz Stefanek notifications@github.com
:

pragmaware:
Search & replace will usually not catch the translated string since (i.e. the msgstr) it's in another language. But it will catch the msgid that should be exactly the same as the string in the sources.
This means that the translation will still continue working, even if the translated string is somewhat inconsistent (but it even depends.. the translation might NOT contain the typo or grammar issue at all).
Well... I don't say that you strictly have to use search & replace. I'm just saying that it's a good tool.
What _should_ be done is to change the string in the sources and all the msgid strings in the *.po files.
Whatever tool helps, it's OK.

DarthGandalf:
No. msgids strings in *.po files should be changed, but also be marked as
fuzzy, so that the translator can check the translation and fix it if
needed. Probably simple regeneration of .po files will be enough, as it
doesn't remove old entries, but compares them with new entries, matches,
and marks them fuzzy itself.

This is not about typos, which are obviously per-language, but about
inconsistencies like "identd" vs "ident service" vs "ident server".

2015-09-28 13:19 GMT+01:00 Szymon Tomasz Stefanek notifications@github.com
:

un1versal:
I think the source strings should be numbers and not actual text strings so they are immutable.

Source would be e.g.

```
        __tr2qs("1"),
```

and the po files would be

Master po

```
msgctxt "#1"
msgid "Some text"
msgstr ""
```

Some language po

```
msgctxt "#1"
msgid "Some text"
msgstr "Some translation"
```

That would be a sane translation system (as other softwares use) as opposed to this existing error prone method.

But since Im not going to do that in any case, whats the verdict for these now?

un1versal:
I can squash more if you guys want.

DarthGandalf:
> That would be a sane translation system (as other softwares use)

Lots of software uses gettext, not only KVIrc. Yes, design decisions made by gettext are sometimes questionable.

>  __tr2qs("1"),

If a translation misses some string, how user is supposed to guess that "1" means "Connect to...", and "2" means "Delay between attempts"? gettext defaults to source string if translation is missing. Mixed translation/English is unpleasant, but better than "1" which wouldn't be usable by end user at all.

> But since Im not going to do that in any case, whats the verdict for these now?

Just revert all changes of .po, then regenerate .pot and .po (`make messages-update`), and gettext will do the right thing. This will allow translators to check whether translation is still valid. Consider cases like:

``` diff
-msgid "Can't add a file offer for file %Q (huh ? file not readable ?)"
+msgid "Can't add a file offer for file %Q (Huh? File not readable?)"
 msgstr "Nelze přidat ke sdílení soubor %Q (huh ? soubor je nečitelný ?)"
```

``` diff
-msgid "Identd accepting connection"
-msgstr "Identd přijímá spojení"
+msgid "Ident service accepting connection"
+msgstr "Ident přijímá spojení"
```

Both changes obviously decrease quality of translation, this is copied from current changes.

DarthGandalf:
Btw, we have transifex, should it be somehow involved in generating .po instead of `make messages-update`? @ctrlaltca

un1versal:
> Just revert all changes of .po, then regenerate .pot and .po (make messages-update), and gettext will do the right thing. This will allow translators to check whether translation is still valid. Consider cases like:

Then close this, and ignore this PR, I wont bother then. its insane amount of work to revert this in the all the po files.

> Btw, we have transifex, should it be somehow involved in generating .po instead of make messages-update? @ctrlaltca

Thats depends on how its setup. but mmm, Im not going close to transifex though.

Sorry if I sound grumpy, its not my intention :D

DarthGandalf:
@uNiversaI why would you do it by hand? :) That would be insane.
Something like `find . -name '*.po' | xargs git checkout 1e1d2a93eb7 --` should work.

un1versal:
> @uNiversaI why would you do it by hand? :) That would be insane.
> Something like find . -name '*.po' | xargs git checkout 1e1d2a93eb7 -- should work.

Well some of us are not that clever. :smile: 

OK done... Hopefully this isnt too bad.

As for regeneration I can push that into here, but to be honest this is cleaner as is and I rather not mess with that.

Once this is ready, maybe someone can help with regenerating po files in a separate PR?
The machine where Im working on this doesn't have a setup environment and its not going to either, if Im on my own here Ill figure something out into a separate PR.

un1versal:
Im ankle deep (upside down) into these strings stuffs.

The (in)consistency of capitalization vs typos vs grammar is just too much to do a partial fix, which means Ill be on here for a while.
The issue really annoys me enough for me to go in head first.

Ill try to push things per pot file, but due to some strings being (duplicated) in more than one pot file this isnt always going to be possible.

So the top commits will be older vs newer revisions at bottom, then Ill squash the lot so that you guys reviewing are not reviewing two versions.

Expect me out in 20 to 1000 years. 
Thx for your patience in advance.

ctrlaltca:
Developers should be able to add and change tr("xxx") in cpp,h files without caring about translations.
Sometimes (like e.g. every X weeks, a month before a release, or anytime) a developer taking care of translations should follow this workflow:
- create a new temporary branch named eg. "translations"
- run `make messages-extract` to extract the .pot files from the source code
- push the pot files containing english strings to transifex using `tx push -s` (more info http://docs.transifex.com/client/, config: https://github.com/kvirc/KVIrc/blob/master/.tx/config )
- ideally, let translators do their work;
- pull the updated translations from transifex using `tx pull -a -f`
- commit the updated pot and po files and open a PR

Regarding this issue, typo fixes should take place only in the cpp files, and they will obviously break old translations. Having translations handled in an external site, any modification made to .po files would be overwritten the next time translations are updated from transifex.

ctrlaltca:
#1637 is a PR created following my previous comment.

pragmaware:
We have 18 languages.
So breaking a single translation _requires work of 18 persons_ to fix.
Additionally most translations are actually unmantained (i.e. there is
no person regularly fixing the *.po file): translators of a particular
language are actually rare.

IMO we should do everything possible to preserve the translations,
even if they aren't perfect.

If the message update procedure is able to mark the translation
as fuzzy but still keep it in the generated *.mo file, it's OK.

Well.. obviously keeping the translation working will not be possible
in every single case, but in many cases it's possible: the identd/ident
issue is an example. We fix it in english. It's better to have the string
translated with a minor glitch than have it untranslated.

pragmaware:
Btw, if translators for most languages show up on transifex, then we obviously
can break translations without caring too much.

uNiversaI:
Well so close this then. Simple.

un1versal:
I understand the point of view but, translations dont matter here which is not as bad as it sounds.

What happens is the old strings are used instead where applicable that wont make KVIrc unusable it just meant that until these translations are updated they are using the old outdated strings.

We dont have to kill the translations completely just push the updated stuff to transifex and wait. WHile in master the core is updated and the translations are the old stuff.

We should make some effort to tell people that we need translators, and if languages are severely unsupported and not maintained then that's another issue.

The only issue HERE is the current state of core strings which is very much less than ideal. And that looks unprofessional and amateurish IMO.
Consider this. Which came first, the chicken or the Egg? Was it core strings or the translations? The impression I get is this is being treated as if the translations came first and are most important when in fact they wont be unusable anyway unless we make them so.

If this is will cause such a big issue with translations you really rather avoid despite the obvious implications, you have some thinking to do and decide if I can go ahead or close this PR.

@ctrlaltca thoughts?

data/defscript/classes.kvs
```diff
@@ -15,19 +15,19 @@ class(installer,object)
 	{
 		if($0 == "")
 		{
-			echo $tr("[Installer] Error: the first argument of copyFiles must be a source directory","defscript")
+			echo $tr("[Installer] Error: The first argument of copyFiles must be a source directory","defscript")
```
pragmaware: After a colon/semicolon the letter should be small.. or I'm wrong?

data/defscript/popups.kvs
```diff
@@ -18,17 +18,17 @@ defpopup(highlight)
 		}
 	}
 
-	item($tr("Stop Highlighting (Local)","defscript"),41) (%:beingHlTmp && !%:beingHl)
+	item($tr("Stop highlighting (Local)","defscript"),41) (%:beingHlTmp && !%:beingHl)
```
pragmaware: Heh.. these capitalization issues always bothered me.
Several UI guides (including the KDE one) suggest to use Title Case for menus, buttons and other simple controls.
This is what I have done in most places. However I agree that it's strongly debatable.
If it wasn't for the guide I'd use normal capitalization (the one you suggest to change to).
Whatever we choose to do, though, must be done for all the menus. That is, we need
a standard to follow everywhere...

data/man/kvirc.en.1
```diff
@@ -64,14 +64,14 @@ Use this port for connection.
...
+An URL in the following form:
 
   irc[6]://<server>[:<port>][/<channel>[?<pass>]]
 
-For each url on the commandline create a new irc-context
+For each URL on the commandline create a new irc-context
 and attempt to connect to <server> on the specified <port>.
 The optional <channel> is joined just after the connection
-has been estabilished.
+has been established.
```
pragmaware, line 74: Errors like this (i.e, typos) are probably very common. They can be fixed massively with
search & replace on all the source + *.po files at once.

data/defscript/classes.kvs
un1versal: The colon isnt acting as a sentence expansion but as a delimiter, the sentence starts after the colon itself.

You may as well say

```
echo $tr("[Installer] Error! The first argument of copyFiles must be a source directory","defscript")
```

Its odd in most of these cases as they are error messages I also didnt feel like changing the whole syntax.

Examples:

```
Incorrect: I want: butter, sugar, and flour.

Correct:
I want the following: butter, sugar, and flour.
OR
I want butter, sugar, and flour.

Incorrect: I've seen the greats, including: Barrymore, Guinness, and Streep.

Correct: I've seen the greats, including Barrymore, Guinness, and Streep.
```

So what I wanna know is can I keep going with this PR or shall I stop?

data/man/kvirc.en.1
un1versal, line 74: There's no point in doing the po and pot files they will get overwritten.at regeneration.

See https://github.com/kvirc/KVIrc/pull/1637 and https://github.com/kvirc/KVIrc/pull/1631#issuecomment-144051147

Also it will make my job 100 times more difficult.

data/defscript/popups.kvs
un1versal: I agree they all should be the same, but camel casing imo in this usage is more wrong than right.

Book titles, Headlines, blog posts titles, I can go with even if its debatable like you say, but here I disagree with the usage.

wodim: https://techbase.kde.org/Projects/Usability/HIG/Capitalization

un1versal: > https://techbase.kde.org/Projects/Usability/HIG/Capitalization

Thx

un1versal:
I have broken down this HUGE PR into PR's with more easily digestible counterparts. So closing here for now.

PR #1637, opened 2015-09-29
extracted new strings and fetched updated translations from transfex

ctrlaltca:


un1versal:
I can appreciate this PR, but it will make my rebase and conflict resolution a nightmare regarding #1631

So if #1631 goes ahead then this PR can be based on that.

ctrlaltca:
I agree on holding back this PR, getting #1631 merged, and then rework this one.
I basically created this PR only as an example as written in #1631.
Anyway, in #1631 there should be no need to touch .po and .pot files, since they would be overwritten at the very next translation update anyway.

un1versal:
mmm right so what you are saying is, what I feared, the only files that need changing are the source right?

In that case it will make my job slightly easier and I can ignore the pot files from now on.
What a mess, being unfamiliar with this porject and way it uses po files is completely diffrent from what Im used to.

ctrlaltca:
Exactly. It will make your job easier, but translators will need to review (fix or accept as unchanged) all their translations for the modified strings.

un1versal:
> Exactly. It will make your job easier, but translators will need to review (fix or accept as unchanged) all their translations for the modified strings.

Sure, unfortunately the source strings are a bit all over the place, so thers going to be some to a lot of changes, however I see some translations are correct anyway. So hurrah to the translators who stuck to the **_right way**_

Thx btw your comment helped a lot since there was mention of broken translations and I was unsure why. Now I know.

un1versal:
I still think its easier to have a main language file for the source so dont have to sift through the code to find strings.

un1versal:
Can we leave the old po files in core until they are fixed in transifex?
, I mean if anyone complains they can go to transifex and fix it else they just use the outdated strings...

If languages can keep using old strings anyway even after language updates its even better.

ctrlaltca:
Closing for merge conflicts, will reopen a new one.

un1versal:
I still have a lot to do there, so maybe leave it till a bit later?

ctrlaltca:
ok

PR #1638, opened 2015-09-29
Add recent instructions for building on Win64

Cizzle:
Recent instructions for building on Win64, with some adjustments they probably work for Win32 as well but I haven't tested and thus didn't want to remove the original Win32 instructions.

### DIFF
--- a/doc/INSTALL-Win64.txt
+++ b/doc/INSTALL-Win64.txt
@@ -1,0 +1,178 @@
+-------------------------------------------------------------------------------
+KVIrc4 win64 compiling howto                                          by Cizzle
+ based on wodim's guide at https://thacid.wordpress.com/2010/10/18/compiling-kvirc-64-bits-from-svn-using-visual-studio-2008/
+ and DarthGandalf's appveyor buildinstructions
+-------------------------------------------------------------------------------
+
+This paper describes how to successfully compile a working copy of KVIrc 4.3.x
+under Microsoft Windows operating systems. For own compilations we assume a
+root-directory C:\kvirccomp\ being used.
+
+-------------------------------------------------------------------------------
+Prerequisites
+-------------------------------------------------------------------------------
+
+- A 64-bit Microsoft Windows operating system.
+	Should work at least on 7
+
+- Microsoft Visual Studio 2013 (Community Edition).
+	Get it from https://www.visualstudio.com/nl-nl/products/visual-studio-community-vs
+	Install in default location with default options, though you can slim it down a lot if wanted.
+
+- CMake.
+	Get it from http://www.cmake.org/download/
+	Install in default location with default options and choose "Add CMake to command path for all users".
+
+- GIT client. 
+	Get one from https://git-for-windows.github.io/
+	Git bash should be enough.
+
+- Qt.
+	Get it from https://www.qt.io/download-open-source/
+	Version 5 should work.
+	Get the online installer and be sure to mark MSVS 2013 64bit version.
+	Install in default location with default options.
+
+- Perl.
+	Get it from http://www.activestate.com/activeperl/downloads
+	Get the 64-bit version, use 5.20.x (newer versions don't have dmake yet in the repo).
+	Install in default location with default options.
+	
+	If you want scripting support, you will also have to do this:
+	Go to C:\Perl64\lib\CORE, make the file config.h not read-only and open it for editing.
+	Find the line that starts with "#define PERL_STATIC_INLINE" and put "/*" in front of it (so it's commented).
+	And of course save the file.
+	
+	Some perl releases miss dmake at install time; if you get a strange error at compilation
+	time, run the following command from cmdline: ppm install dmake
+
+- zlib.
+	Get it from http://zlib.net/
+	Get the source as we need to compile it ourselves (until a 64bit dll is provided).
+	Compiling (replace versions with what you have):
+		Unpack in C:\kvirccomp\; then open a VS x64 Native CMD Prompt and run:
+			cd C:\kvirccomp\zlib-1.2.8
+			"C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat" x64
+			nmake -f win32/Makefile.msc AS=ml64 LOC="-DASMV -DASMINF -I." OBJA="inffasx64.obj gvmat64.obj inffas8664.obj"
+
+- OpenSSL.
+	Get it from http://www.openssl.org/source/ to compile yourself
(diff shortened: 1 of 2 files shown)


doc/INSTALL-Win64.txt
```diff
@@ -0,0 +1,178 @@
...
+	Git bash should be enough.
+
+- Qt.
+	Get it from https://www.qt.io/download-open-source/
+	Version 5 should work.
+	Get the online installer and be sure to mark MSVS 2013 64bit version.
+	Install in default location with default options.
+
+- Perl.
+	Get it from http://www.activestate.com/activeperl/downloads
```
DarthGandalf, line 37: Oh. I tried with strawberry perl and active perl, until found that .h magic which worked with strawberry perl.
So with active perl it's similar? interesting.

doc/INSTALL-Win64.txt
```diff
@@ -0,0 +1,178 @@
...
+
+- Python.
+	Optional for scripting purposes
+	Get it from http://www.activestate.com/activepython/downloads/
+	Get the 64-bit 2.x version as 3.x is not backwards compatible and supported by now.
+	Install in default location with default options.
+
+- Enchant.
+	Optional for spellchecking support
+	Get it from https://ci.appveyor.com/api/buildjobs/kf2efxpaes6uugwg/artifacts/enchant.7z
```
DarthGandalf, line 88: That's cheating. Please also point at sources (I had to modify libenchant a bit regarding path where it searches for dictionaries)

doc/INSTALL-Win64.txt
```diff
@@ -0,0 +1,178 @@
...
+	 cd build
+	 cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=release -DWANT_KDE=0
+	 nmake
+	 nmake install
+ 
+	Your binaries are now located at C:\kvirccomp\kvirc\build\release.
+	But you can’t start kvirc.exe as there're some libraries missing.
+	You can either copy these to the system dir, add their locations to %PATH% or copy them to the KVIrc-release dir.
+	The latter is preferred for portability reasons.
+	 robocopy C:\Qt\5.4\msvc2013_64\bin C:\kvirccomp\kvirc\build\release\ Qt5Network.dll Qt5Core.dll Qt5Widgets.dll Qt5Gui.dll Qt5WebKitWidgets.dll Qt5WebKit.dll Qt5Sensors.dll Qt5Positioning.dll Qt5Quick.dll Qt5Qml.dll Qt5Multimedia.dll Qt5WebChannel.dll Qt5Sql.dll Qt5MultimediaWidgets.dll Qt5OpenGL.dll Qt5PrintSupport.dll Qt5WinExtras.dll icuin53.dll icuuc53.dll icudt53.dll
```
DarthGandalf, line 164: If you use Qt5, it has a special program `windeployqt` which looks at kvirc.exe and copies required .dll files (including plugins).
It's used in AppVeyor build.

doc/INSTALL-Win64.txt
```diff
@@ -0,0 +1,178 @@
...
+	 robocopy C:\kvirccomp\openssl-1.0.2d\out32dll C:\kvirccomp\kvirc\build\release\ libeay32.dll ssleay32.dll
+	 copy C:\kvirccomp\zlib-1.2.8\zlib1.dll C:\kvirccomp\kvirc\build\release\
+	 copy C:\Perl64\bin\perl520.dll C:\kvirccomp\kvirc\build\release\
+	Optionally copy:
+	 copy C:\Windows\System32\python27.dll C:\kvirccomp\kvirc\build\release\
+
+- Add extra plugins
+	Qt provides extra plugins which can be used, these have to be placed in a subdirectory named "qt-plugins".
```
DarthGandalf, line 172: There is a caveat for platform/ plugins in Qt5, described in .appveyor.yml

doc/INSTALL-Win64.txt
```diff
@@ -0,0 +1,178 @@
...
+ based on wodim's guide at https://thacid.wordpress.com/2010/10/18/compiling-kvirc-64-bits-from-svn-using-visual-studio-2008/
+ and DarthGandalf's appveyor buildinstructions
+-------------------------------------------------------------------------------
+
+This paper describes how to successfully compile a working copy of KVIrc 4.3.x
+under Microsoft Windows operating systems. For own compilations we assume a
+root-directory C:\kvirccomp\ being used.
+
+-------------------------------------------------------------------------------
+Prerequisites
```
DarthGandalf, line 12: Many of them are easier to be installed using chocolatey and nuget, though manually downloading them and running setup.exe (or whatever) works too, of course.

doc/INSTALL-Win64.txt
```diff
@@ -0,0 +1,178 @@
...
+-------------------------------------------------------------------------------
+
+- A 64-bit Microsoft Windows operating system.
```
DarthGandalf, line 15: I think you can cross-compile from 32 bits to 64 bits. I didn't test it for kvirc though, and the only possible problem I foresee, is perl support, because perl is used for both docs and scripting, so it needs to run and to link.

doc/INSTALL.txt
```diff
@@ -618,11 +618,11 @@
...
 
 -------------------------------------------------------------------------------
-  9. Compiling on Win32
+  9. Compiling on Windows
 -------------------------------------------------------------------------------
 
-  There is a detailed compilation and installation HOWTO for Win32 systems
-  located in the doc directory. It's named INSTALL-Win32.txt.
+  There is a detailed compilation and installation HOWTO for Win32 and Win64 systems
+  located in the doc directory. They're named INSTALL-Win32.txt and INSTALL-Win64.txt.
```
DarthGandalf, line 625: Are they very different? I'd prefer deduplicating the information...

un1versal, line 625: +1 for **not** duplicating information,

Make a readme named INSTALL-Windows.txt wack everything in there and document any differences.

bare in mind you need to also add the document and any document name changes to https://github.com/kvirc/KVIrc/blob/master/doc/CMakeLists.txt

doc/INSTALL-Win64.txt
Cizzle, line 88: It's no more cheating than using a precompiled package for openssl? I was going to try compiling it myself, but then I saw the instructions and thought: this is an install guide for KVIrc, not libenchant. And I know there's instructions in here for compiling some packages oneself. Anyway, I'll add links to your github fork and the instructions there.

doc/INSTALL.txt
Cizzle, line 625: I'll give that a try later on.

PR #1640, opened 2015-09-30
fix #1591 add custom cursor thickness size

### LINKED ISSUE #1591: Make KviInputEditor cursor width configurable
It would be nicer not to have to wonder so much where position of cursor is, even with higher contrast colors I still find it difficult to spot that bulimic thin line.

the spellchecker relies on cursor position for instance  and this is an issue when you have to hunt down the cursor position.

Also, not everyone is young or have perfect vision or even dont have any visual impairment issues, like color blindness. 

---

Edit

Reopened because Linux users since https://github.com/kvirc/KVIrc/commit/a506f83cf7861f105546fc72065838e6fee1efd0 are being left out and Im a Linux user.

**EDIT**
What would be ideal is optional cursors types, Windows cmd supports it, Linux supports it and would solve this request also.

capture
capture
capture2

un1versal:
By default it will look like this (with this PR) in Linux

If I understand https://github.com/kvirc/KVIrc/commit/a506f83cf7861f105546fc72065838e6fee1efd0 correctly, Windows users get the Windows default or what their accessibility options defines.

That would be ideal in Linux counterparts but Ive no idea how to achieve that.

**Before**
capture

**After**
capture1

We can either make the cursor 2 pixels wide or better, this one uses 8 since its my preferred cursor type and what Im use to in all terminals.

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -333,7 +333,8 @@ KviBoolOption g_boolOptionsTable[KVI_NUM_BOOL_OPTIONS]=
 	BOOL_OPTION("ShowDialogOnChannelCtcpPage",false,KviOption_sectFlagCtcp),
 	BOOL_OPTION("PopupNotifierOnNewNotices",true,KviOption_sectFlagFrame),
 	BOOL_OPTION("UserListViewUseAwayColor",true,KviOption_sectFlagUserListView | KviOption_resetUpdateGui),
-	BOOL_OPTION("ShowUserFlagForChannelsInWindowList",true,KviOption_sectFlagWindowList | KviOption_resetUpdateGui)
+	BOOL_OPTION("ShowUserFlagForChannelsInWindowList",true,KviOption_sectFlagWindowList | KviOption_resetUpdateGui),
+	BOOL_OPTION("EnableCustomCursorWidth",false,KviOption_resetUpdateGui)
 };
 
 #define STRING_OPTION(_txt,_val,_flags) KviStringOption(KVI_STRING_OPTIONS_PREFIX _txt,_val,_flags)
@@ -669,7 +670,8 @@ KviUIntOption g_uintOptionsTable[KVI_NUM_UINT_OPTIONS]=
 	UINT_OPTION("OnJoinRequestsDelay",1,KviOption_sectFlagConnection), // FIXME: Wouldn't this be nicer in msecs defaulting to 100-200 ?
 	UINT_OPTION("ToolBarIconSize",22,KviOption_groupTheme | KviOption_resetReloadImages),
 	UINT_OPTION("ToolBarButtonStyle",0,KviOption_groupTheme ), // 0 = Qt::ToolButtonIconOnly
-	UINT_OPTION("MaximumBlowFishKeySize",56,KviOption_sectFlagNone)
+	UINT_OPTION("MaximumBlowFishKeySize",56,KviOption_sectFlagNone),
+	UINT_OPTION("CustomCursorWidth",1,KviOption_resetUpdateGui)
 };
 
 #define FONT_OPTION(_name,_face,_size,_flags) \
--- a/src/kvirc/kernel/KviOptions.h
+++ b/src/kvirc/kernel/KviOptions.h
@@ -353,8 +353,9 @@ DECLARE_OPTION_STRUCT(KviStringListOption,QStringList)
 #define KviOption_boolPopupNotifierOnNewNotices 258                      /* query */
 #define KviOption_boolUserListViewUseAwayColor 259                       /* userlist */
 #define KviOption_boolShowUserFlagForChannelsInWindowList 260
+#define KviOption_boolEnableCustomCursorWidth 261                          /* interface */
 
-#define KVI_NUM_BOOL_OPTIONS 261
+#define KVI_NUM_BOOL_OPTIONS 262
 
 
 #define KVI_STRING_OPTIONS_PREFIX "string"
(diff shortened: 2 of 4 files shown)


ctrlaltca:
what about adding an option?
- use system cursor thickness
- use custom thickness: [ 8] px

un1versal:
> what about adding an option?
> 
> ```
> use system cursor thickness
> use custom thickness: [ 8] px
> ```

Sure, that would be ideal, I dont know how though. maybe I would make the value user selectable with a pre-selected value of 8

But again no idea how.

wodim:
Make this an option, because having a cursor of 8 pixels is not an "improvement", especially taking into account how it makes things inconsistent with the rest of the system.

un1versal:
Im all for making it optional, but as is idk how.

So some help needed there.

wodim:
The options are added to `\src\kvirc\kernel\KviOptions.h` (two options it would be: one for the radio button, other for the input box).

The UI for the options is added to a file in `/src/modules/options/`. Every file is a widget that represents a different section of the options dialog. Search for the appropriate one.

Then those options can be read later from code (in this case the `KviInputEditor` constructor) using macros such as `KVI_OPTION_BOOL`, `KVI_OPTION_UINT`...

un1versal:
That just made my head explode

un1versal:
OK Good news bad news...

**Good News**
This is how this looks in options.

capture

And the Custom Thickness size works 100%.

**The bad news...**
- You must quit and rejoin channel before the width is applied.  
- ~~Also the checkbox while unchecked should mean that the width selector is greyed out but its not.~~

Help?

ctrlaltca:
> Also the checkbox while unchecked should mean that the width selector is greyed out but its not.

```
KviBoolSelector *b = addBoolSelector(....)
KviUIntSelector *u = addUIntSelector(....)
connect(b,SIGNAL(toggled(bool)),u,SLOT(setEnabled(bool)));
```

ctrlaltca:
> You must quit and rejoin channel before the width is applied.

This one is a bit harder; look in the code for "KviApplication::optionResetUpdate"; you'll need to add a new flag "KviOption_resetInputCursor" and implement it

wodim:
That's very good! Do as @ctrlaltca said and if you need further assistance I'll help you later tonight.

un1versal:
> KviBoolSelector *b = addBoolSelector(....)
> KviUIntSelector *u = addUIntSelector(....)
> connect(b,SIGNAL(toggled(bool)),u,SLOT(setEnabled(bool)));

Yep that works... Obliviously :tongue:  Because Im oblivious and its obvious. Should have looked up at the previous option.

capture

> This one is a bit harder; look in the code for "KviApplication::optionResetUpdate"; you'll need to add a new flag "KviOption_resetInputCursor" and implement it

mmm sure Ill dig into that and see what damage I can make in there.

Thx guys.

pragmaware:
Hm.. probably KviOption_resetUpdateGui is enough for the cursor thickness update.
It will trigger "applyOptions" which should be the place where you determine the
cursor width to use.

un1versal:
> Hm.. probably KviOption_resetUpdateGui is enough for the cursor thickness update.
> It will trigger "applyOptions" which should be the place where you determine the
> cursor width to use.

Can you elaborate on that? Its already using **`KviOption_resetUpdateGui`**

https://github.com/uNiversaI/KVIrc/blob/7409e480e4738b6462e168578c898b5fa939c7e2/src/kvirc/kernel/KviOptions.cpp#L337
https://github.com/uNiversaI/KVIrc/blob/7409e480e4738b6462e168578c898b5fa939c7e2/src/kvirc/kernel/KviOptions.cpp#L674

Its **`KviOption_sectFlagInput | KviOption_resetUpdateGui`** do you mean drop **`KviOption_sectFlagInput`**?

**Edit** 

Using just **`KviOption_resetUpdateGui`** does exactly the same, need to quit channel and rejoin or disconnect and reconnect (if using autojoin channels)

unless **`KviOption_resetUpdateGui`** is buggy? Or am I missing something?

ctrlaltca:
@pragmaware is right; you should just need to add something like

```
m_iCursorWidth = KVI_OPTION_UINT(KviOption_uintCustomCursorThicknessSize);
```

inside this method: https://github.com/kvirc/KVIrc/blob/master/src/kvirc/ui/KviInputEditor.cpp#L187

un1versal:
> @pragmaware is right; you should just need to add something like
> 
> m_iCursorWidth = KVI_OPTION_UINT(KviOption_uintCustomCursorThicknessSize);
> 
> inside this method: https://github.com/kvirc/KVIrc/blob/master/src/kvirc/ui/KviInputEditor.cpp#L187

Ive no idea how to do that correctly Im looking at that and im blank its really not obvious to me what to do in there.

Im trying just to wack that below https://github.com/kvirc/KVIrc/blob/master/src/kvirc/ui/KviInputEditor.cpp#L193 and compiling to see how that goes

ctrlaltca:
I'm looking at it; also, lol at the branch name

un1versal:
> lol at the branch name

:smile:

I spotted a couple of pebcak typos, lets see... but even with the typos 

> m_iCursorWidth = KVI_OPTION_UINT(KviOption_uintCustomCursorThicknessSize);
> 
> inside this method: https://github.com/kvirc/KVIrc/blob/master/src/kvirc/ui/KviInputEditor.cpp#L187

worked

So lets rework that and retest...

un1versal:
@ctrlaltca this as is now seems to work fine. :+1: 

You can now set any cursor size and its automatically applied when you click apply.

The but lies in the bool checkbox when you first navigate to the interface > Input area is unchecked by default which is OK but the width selector is active and only becomes inactive if you check/uncheck the box.

The width selector should be inactive even if its never been set right?

The code seems right to me I cant find any typos or issues

So Im still missing something somewhere what idk.

Im wondering if

```
KviUIntSelector *f = addUIntSelector(0,7,0,7,__tr2qs_ctx("Define custom cursor width - in pixels","options"),KviOption_uintUseCustomCursorWidth,1,24,8,true);
```

Should be

```
KviUIntSelector *f = addUIntSelector(0,7,0,7,__tr2qs_ctx("Define custom cursor width - in pixels","options"),KviOption_uintUseCustomCursorWidth,1,24,8,false);
```

Or is this wrong assumption? Seems worng to me due to conflict.

ctrlaltca:
Enabling or disabling the width selector when it's created should be dependent to the value of the previous option:

```
KviUIntSelector *f = addUIntSelector(0,7,0,7,__tr2qs_ctx("Define custom cursor width - in pixels","options"),KviOption_uintUseCustomCursorWidth,1,24,8,KVI_OPTION_BOOL(KviOption_boolUseCustomCursorWidth));
```

Also, in KviInputEditor.cpp you need to check if KVI_OPTION_BOOL(KviOption_boolUseCustomCursorWidth) is enabled before applying the custom cursor width.

un1versal:
> Also, in KviInputEditor.cpp you need to check if KVI_OPTION_BOOL(KviOption_boolUseCustomCursorWidth) is enabled before applying the custom cursor width.

Is neither if you look at it! How do I check lol...

wodim:
:shipit:

un1versal:
Well this works fully on normal input.

It wont work on any editor or multiline input box...

I guess wining half the battle wasn't too bad.

PR #1642, opened 2015-09-30
build: Only display GTKSTYLE warning if on Qt4

staticfox:
Because GTKSTYLE is now mandatory in Qt5, the warning should be
silenced for Qt5 builds as there is no physical way to disable
it (and at this point, no legitamate reason to as Qt5+KVIrc is
non functional without it)


un1versal:
Was just thinking about that, now need to update docs that mention this.

staticfox:
Just the INSTALL.txt file references it, I can update it if you want.

un1versal:
Sure, that would be nice. I remember vaguely reading that elsewhere as well but for the life of me cant find it.

PR #1644, opened 2015-09-30
Fix the minimum width bug for KviTalListWidget

wodim:
`KviTalListWidget` has a bug, easily reproducible:
1. Open the Theme Manager dialog.
2. Make it wider.
3. Close it.
4. Open it again.
5. Make it thinner.

A scrollbar appears. That's a bug. This PR fixes it, but there's another bug I have been unable to fix: even though the `KviTalIconAndRichTextItemDelegate`s are made taller or smaller depending on the contents of them (can be checked by clicking on them), their position inside the parent `KviTalListWidget` does not change, so they are floating or overlap if their sizes change dramatically. Make the dialog as wide as you can, close and open it, then make it thinner and you'll see what I mean.

I have read I need to `emit sizeHintChanged()` from somewhere, but I just have no clues or energy left to keep trying.


wodim:
Also: what does `Tal` mean?

ctrlaltca:
Toolkit abstraction layer, ore something like that. The idea is to create an additional layer between libraries (qt) and the interfases used by kvirc code. IIRC this was implemented following qt's habit of breaking its own interfaces between major versions (e.g. qt3 to qt4)

un1versal:
+1 for this especially if it solves the horizontal scrollbar visible by default.

wodim:
@pragmaware You closing this means you will take a look at the problem I described? Because this fix was half done.

pragmaware:
Done, probably.

Issue #1645, opened 2015-09-30
Broken icons on the Manage Themes dialog


Stanzilla:
Picture says it all:

![manage_themes_-_kvirc_2015-09-30_22-04-52](https://cloud.githubusercontent.com/assets/75278/10205049/5ade1260-67bf-11e5-8fef-5e3444056d20.png)

wodim:
#1314 #1309

un1versal:
They not "broken" per say they are corrupted.

For reference sake the bug report that remains open is #1309 
#1314 and this one are closed as we dont need 3 reports for same issue.

Also Icon corruption bug only happens to the **big icons**. the small icons survive whatever #1309, #1314, and this reports cause is.

Issue #1643, opened 2015-09-30
Can't save themes


Stanzilla:
It always fails with the message that it can't save an image to the theme. Unchecking `save icons`is a workaround.

![save_current_theme_-_kvirc_2015-09-30_22-06-20](https://cloud.githubusercontent.com/assets/75278/10205098/90b99dfa-67bf-11e5-88b5-648090ff8a0b.png)

```
KVIrc 4.3.2 'Aria'

Runtime Info:
System Name: Windows Release : 6.2.9200
System Version: Windows 8 Professional (x64) (Build 9200)
Architecture: Intel64 Family 6 Model 58 Stepping 9

Build Info:
Build Date: 2015-09-30 01:21:42 UTC
Sources Date: 20150926
Revision Number: git-5439-g8ca6791
System Name: Windows-6.2
CPU Name: AMD64
Build Command: C:/Program Files (x86)/CMake/bin/cmake.exe
Build Flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=C:/projects/kvirc/build/release
   Threads=Win32
Compiler Name: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/amd64/cl.exe
Compiler Flags: /DWIN32 /D_WINDOWS /W3 /GR /EHsc
```

un1versal:
~~If they got corrupted you wont be able to.~~

~~You just reported https://github.com/kvirc/KVIrc/issues/1645 a triplicate report, now it makes sense to me, that if the icons are corrupted because of that bug, they cannot be saved into a theme, no?~~

In any case I dont recommend anyone be saving any icons into themes at this stage, this is why I requested this feature to optionally save icons to be added.
The rationale is, If I update any icons , change their names in core or add more specific icons the theme is already outdated and possibly broken.

un1versal:
Well I just tested this and it does pop up this error on current master f5dc3c4baf6a30b2cbcc9a50c129a8564b22fa13 which confirms this report.

---

Maintaining the advisory notice on saving icons into themes. its not really a good idea tbh, since any themes that contain any icons overwrite the default ones and ignores any updates to core icons name or image updates like splashes and other non icons images.

Where people want to package different icons with their themes, they only need to include the icons they wish to replace and if they wish to replace all then this makes even less sense, but none-the-less it is helpful to the process as you need to know what you are replacing.

---

IMO this option should be off by default and carry some informational text. **EDIT** see #1647

[...]

Issue #1314, opened 2015-07-26
Configure Theme - Text - Message colors - Icon change menu corrupt


DarthGandalf:
**Reported by tedych on 18 Aug 2012 11:42:01 UTC**
Windows 7 Ent. 64 bit. KVirc 4.2.0.
I just noticed that there was no icon for the channel message when my nick was highilghted and decided to assign an icon to that.
Settings - Configure theme... - Text - Message colors. There I select a line and try to change the icon from the small button down.
A corrupted menu with one and a half choice appears. The half is on the same line half visible, numbered 1 and 2..
I'm not sure if this is a latent bug because this is the first time I've been trying to change icons.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1314

DarthGandalf:
**Modified by tedych on 18 Aug 2012 11:44:05**
Summary: Configure Theme - Text - Message colors - Icon change menu corrupted → Configure Theme - Text - Message colors - Icon change menu corrupt

DarthGandalf:
**Modified by Visitor_099 on 19 Jan 2013 13:44:24**
Component: Core → GUI

un1versal:
I may have accidentally seen this issue for the first time yesterday, current master compile, more testing required.

heinrich5991:
Perhaps I have a similar bug. When I open the icon menu mentioned above, it shows me exactly zero choices, although the messages in the various channels I'm in _do_ have icons.

un1versal:
Duplicate reports #1309 #1645 closing this.

Issue #1309, opened 2015-07-26
Icons in toolbar gone, and some icons corrupted


DarthGandalf:
**Reported by waser on 26 Jul 2012 02:03:11 UTC**
I closed kvirc one night, it was fine, reopened it the next day, and all icons are a mess. The Toolbar still is set at "icons" (and not text). I recompiled, and the problem still persist.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1309

---

**EDIT by team** Adding a screenshot from duplicate report #1645
![manage_themes_-_kvirc_2015-09-30_22-04-52](https://cloud.githubusercontent.com/assets/75278/10205049/5ade1260-67bf-11e5-8fef-5e3444056d20.png)

DarthGandalf:
**Modified by waser on 27 Jul 2012 01:35:53**
Component: Core → GUI

DarthGandalf:
**Modified by waser on 11 Sep 2012 22:31:48**
Summary: Icons in toobar gone, and some icons corrupted → Icons in toolbar gone, and some icons corrupted

DarthGandalf:
**Comment by waser on 15 Oct 2012 21:32:42**
As it turns out, I've installed kvirc on a Windows machine and imported my configurations from the buggy linux compilation, and the bug is still present, so I suspect a configuration problem causing all this. I have yet to find what is the source of the corrupted images.

un1versal:
Ive experienced this once or twice with 4.3.1 some SVN build and up to 4.3.2 so this isnt 4.3.2 specific.

What happened here last time this issue appeared was the large icons in the active theme somehow got corrupted. The coresmall icons remained OK though, weird it was only the toolbar ones.

Why that happened Ive no idea. At the time, I was testing recent theme changing fixes **after** #1542 and #1541 were merged had to restart KVIrc after changing themes a great deal of times.

Always Linux though.

un1versal:
This happened again yesterday, seems if you change themes a few times in a row seems to trigger this, I should have grabbed a screenshot of it last night...

un1versal:
#1314  #1645 where closed since this is the original report and still applies. So we keep this and close the others.

un1versal:
Also **Icon corruption bug** only happens to the **big icons**. the small icons survive whatever #1645, #1314, and this reports cause is.

:warning: Changed priority of report due to multiple reports and how many times this has happened to myself alone. It renders KVIrc unusable so this is beyond normal.

[...]

PR #1647, opened 2015-10-01
[Theme save options] Save icons with themes + add small description

un1versal:
There's room for improvement on messages, this will hopefully make it clear what the option does and when should it be used.

By default option is should be **off** which is the recommended state.

**Before**
before

**After**
capture

### DIFF
--- a/src/modules/theme/SaveThemeDialog.cpp
+++ b/src/modules/theme/SaveThemeDialog.cpp
@@ -152,13 +152,23 @@ SaveThemeDialog::SaveThemeDialog(QWidget * pParent)
 	pPage = new QWidget(this);
 	pLayout = new QGridLayout(pPage);
 
-	m_pSaveIconsCheckBox = new QCheckBox(__tr2qs_ctx("Save icons","theme"),this);
+	m_pSaveIconsCheckBox = new QCheckBox(__tr2qs_ctx("Save icons and images with theme","theme"),this);
 	m_pSaveIconsCheckBox->setChecked(true);
 
-	pLayout->addWidget(m_pSaveIconsCheckBox,0,0);
-	pLayout->setRowStretch(1,1);
+	pLabel = new QLabel(pPage);
+	pLabel->setWordWrap(true);
+	QString ozText = "<p>";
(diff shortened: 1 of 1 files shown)


staticfox:
Is there a point to keeping the option at all for now? I tried this on a vanilla build and it still breaks. IMO I don't think there's a feasible way to get the option to work at all.

wodim:
If there is "room for improvement in messages", I'd say that after this commit there is even more room for improvement.

staticfox:
In my personal opinion, https://github.com/blindirc/KVIrc/commit/d0ca42300c18ec4f345a77981a7624e5041b6bfe would make a lot more sense since there is no chance of this working plus it gives us more time to focus on fixing the actual issue rather than dabble with long dialog messages and possible duplicate reports for people who smash next through option menus.

un1versal:
> In my personal opinion, blindirc@d0ca423 would make a lot more sense since there is no chance of this working plus it gives us more time to focus on fixing the actual issue rather than dabble with long dialog messages and possible duplicate reports for people who smash next through option menus.

This PR has nothing to do with the bug report #1643, or because this option is broken.

And why do you think that #1645, #1314, and #1309 are related to #1643 imo its not as Ive experienced this icon corruption bug 3 times and none of those times was in conjunction or anywhere in the neighborhood of this, just hammering on theme changing while testing other issue can cause the icons to die.

Icon corruption bug is only the **big icons**. the small icons survive whatever  #1645, #1314, and #1309 cause is.

None the less. your PR makes sense temporarily So Ill support that with some minor changes to the notes.

After when #1643 is fixed  this one can go in as the reason why I requested this feature and why adding this message is due to reason explained on dialog messages and not #1643

To be honest the only reason why saving icons makes sense is if you want to replace them completely. But...

pragmaware:
TL;DR :laughing: 
This text probably belongs to a "theme HOWTO" instead of the GUI. A shortened version could eventually end up in a tooltip for the option.

un1versal:
This PR is already outdated, My local version is much shorter. but the issue is the text goes to middle of box. instead of just bellow the option.

un1versal:
OK this is as ready as it will ever be.

capture
The text has been capped to minimum, the rest we can add to docs maybe when I figure how to do that and or tooltips

@pragmaware your button for a verdict.

PR #1648, opened 2015-10-02
savethemedialog: Disable saving theme's icons

staticfox:
Disable this due to the function being broken. We can always re-enable it once it is fixed again.

Edit

Bug report #1643


un1versal:
@pragmaware your button here, Im unsure if the save theme ever worked since I dont use it. And a fix may come instead of this PR, but if not this should go in as a remedy.

Once the underlying cause is fixed this option should be unselected by default and also have some message on explaining when to use this option. See #1647

pragmaware:
We shouldn't disable a feature just because it doesn't work on one platform.
The right thing is to fix issue #1643 .

pragmaware:
#1643 is fixed so this pull is no longer necessary.

Issue #1566, opened 2015-09-10
Get rid of KVI_SOURCES_DATE and have a version handle the sources date - Merge #1566 and #1548: 


un1versal:
The value of 
# define KVI_SOURCES_DATE "20150910"
# define KVI_SOURCES_DATE_NUMERIC 0x20150910

Should be something Dynamic for instance in Linux you can do:

Reference: 5fa4ce8d882cb03e6a2a79cbac45e5e2f2870d17
#### Edit Sat Oct  3 11:59:38 BST 2015

In reference to https://github.com/kvirc/KVIrc/pull/1649#issuecomment-145236664

It makes no sense to manually update these fields and also since discussion in https://github.com/kvirc/KVIrc/pull/1649 I think that things could be handled better automatically and in a saner manner.

Im not proposing that we ditch git describe just that we ditch #define KVI_SOURCES_DATE and improve kvirc versioning to include that automatically.

So I submit to your consideration revising the program version number in a way that doesn't break git describe and handles this sources date.

Suggestions at 

https://github.com/kvirc/KVIrc/pull/1649#issuecomment-145211207
https://github.com/kvirc/KVIrc/pull/1649#issuecomment-145211569
https://github.com/kvirc/KVIrc/pull/1649#issuecomment-145211681

The date is extracted as described by **ctrlaltca** in  https://github.com/kvirc/KVIrc/issues/1566#issuecomment-139749165

Previously suggested formats:

```
2015-10-02-7d8be0f 
2015-10-02-g7d8be0f  (if that works with git describe) 
2015-10-02-5484-g7d8be0f 
```

Or any other better suggestions?

pragmaware:
That defines are meant to reflect the "release" date of the source code, not compilation date...
Or I'm missing something?

un1versal:
idk, it seems weird that we have to manually edit this every time theres a release. hence the thought above as is that field is no longer correct today.

pragmaware:
Ideally it should be changed to the late of the last commit... but well, I guess that it's not very easy to do.

un1versal:
stat -c %y .git/FETCH_HEAD does that

pragmaware:
On my system it returns a different date every time I git pull.
The sources date should refer to the date of the last commit, not the last pull.

un1versal:
Yes, probably not simple to achieve automatically.

[...]

PR #1649, opened 2015-10-02
Fix #1548 git version mangled by git describe

### LINKED ISSUE #1548: Git hash is being mangled in build revision
Essentially g is being appended to all the hashes.

First noticed at d00p while checking the progress of the site.

capture

looking at https://github.com/kvirc/KVIrc/pull/1546 you can see the resulting builds all have a **g** appended to hash https://ci.appveyor.com/project/DarthGandalf/kvirc/build/0.0.0.0.1-branch-master-build-315/job/4sq89ayh47vgswc3/artifacts

I presume we are stripping g**it** and leaving the g behind which is OK for filenames but when you follow the hash it results into a 404

Its also is being added to version

```
-- Build Date                     : 2015-09-04 14:23:23 UTC
-- Build Version                  : 4.3.2
-- Build Codename                 : Aria
-- Build Revision                 : git-5275-g7d14b49
```

On a side note this should read 

-- Build Revision                 : 5275-git-7d14b49

since 5275 is NOT a git hash

staticfox:
real quick since im half way out the door so im just copy pasting from irc

[12:00:21 PM] <BlindSight> un1versal:  ping
[12:02:18 PM] uNmowed|2 [~Kaede@cli-nw.71.184.helios-nw.ru] has quit IRC: Read error: Connection reset by peer
[12:03:36 PM] alexzulu[away] [~alex@85.115.248.197] has joined #KVIrc
[12:04:13 PM] <BlindSight> well whenever you're around, https://github.com/blindirc/KVIrc/commit/74c6dc1cfabf9fba64013afdb25d994319f0fb3a its a hack but it cant really fail since "git" will never change, [0-9] will never match, and the -g is always gonna be a part of the revision and the revision itself is always [a-zA-Z0-9] so it cant contain any symbols


staticfox:
done and updated commit message

un1versal:
idk if the fix #1548 need to be in beginning or just anywhere will do, I guess we will find out.

thx again. :+1:  :+1:  :+1:

DarthGandalf:
> since "git" will never change

What made you think so?

The format of `git describe`'s output is `<tag>-<number_of_commits_since_tag>-g<hash>`

To make `git describe` working at all after migration to git, I created a tag at the very beginning of history and named it `git` (I didn't come up with any better name, `start` would be even worse). When 4.4.0 is released, commits after it will look like `v4.4.0-12-g12345` (depending on how the tag will be named)

staticfox:
Fair point, didn't think about that. Either way though I can come up with
significantly better regex, however for the now thing both scenarios you
mentioned won't interfier with the replace.

DarthGandalf:
@blindirc see https://github.com/kvirc/KVIrc/issues/1548#issuecomment-145122787

I think such replace shouldn't happen at all.

staticfox:
Nor do I, I only brought this up because @univeral mentioned the ticket was
open and it continued to break support for some things. In any event you'd
have to ask him since I submitted this based on what he said and the fact
the ticket was open.

DarthGandalf:
@uNiversaI which exact things are broken?

un1versal:
Why do you think its acceptable to mangle the hash and make it invalid to even git itself?

d00ps site is an example of where this broke it and if any user looks for that same hash manually they get the same issue.

Not only is the hash made invalid the g is duplicated meaning and you end up with two gits on version string..

see https://github.com/kvirc/KVIrc/issues/1548#issuecomment-137751431

IMO this is a bug and should be fixed. which blindirc was kind enough to submit a fix for I dont understand why the resistance.

As a side note the whole version string could do with better handling but thats another matter,

DarthGandalf:
>  Why do you think its acceptable to mangle the hash and make it invalid to even git itself?

This is output of `git describe` as is. It _is_ valid for git itself. Run `git show git-5434-g1e1d2a9` and see yourself that git recognizes it.
Substring of that string is not valid, right. Why would it be valid? `iversal` is not your valid nick either, but only substring of it.

> d00ps site is an example of where this broke it

d00p chose to parse the filename and did it incorrectly. Now he's not parsing it at all, but showing the correct git hash from a side channel. The whole `git-5434-g1e1d2a9` would work in the link to github too though.

> you end up with two gits on version string..

How do you propose to name the tag in the beginning of history? I chose to name it `git`, but it's not a very good name.

> As a side note the whole version string could do with better handling but thats another matter,

May you explain what do you mean?

un1versal:
I have a feeling this its a exercise in futility and frustration and Im slowly loosing will to live hahahaha.

> This is output of git describe as is. It is valid for git itself. Run git show git-5434-g1e1d2a9 and see yourself that git recognizes it.

Run `git show 1e1d2a9` is same result and universally recognizably hash. typing 1e1d2a9 and typing git-5434-g1e1d2a9 one works the other doesnt right here on comment box.

That's OK, if it was my baby I would fight for it too even if its an ugly baby. :smile: 

Lets look at the remainder.

> As a side note the whole version string could do with better handling but thats another matter,

---

> May you explain what do you mean?

Sure, just promise me you arent asking a rhetorical question. because none of my explanations are acceptable to you..

currently we have -> git-5459-g7d8be0f I think the version should be simple.

```
KVIrc 4.3.2 'Aria'

Runtime Info:
System Name: Linux 3.13.0-62-generic
System Version: #102-Ubuntu SMP Tue Aug 11 14:29:36 UTC 2015
Architecture: x86_64

Build Info:
Build Date: 2015-10-02 11:55:40 UTC
Sources Date: 20150926
Revision Number: git-5459-g7d8be0f
System Name: Linux-3.13.0-62-generic
CPU Name: x86_64
Build Command: /usr/bin/cmake
Build Flags: 
   MANDIR=man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler Name: /usr/bin/c++
Compiler Flags: N/A
```

Its presented as **Revision Number: git-5459-g7d8be0f**

> How do you propose to name the tag in the beginning of history? I chose to name it git, but it's not a very good name.

Do you really need any? lets leave that one for later...

Why 5434 or any other SVN (I presume) related legacy version number? Do we really need to keep that?

IMO we should merge #1566 and #1548 and make it 

So instead of **git**, we have **kvirc**

instead of a mangled hash partially or not its recognizable as a valid hash.

the date from @ctrlaltca comment https://github.com/kvirc/KVIrc/issues/1566#issuecomment-139749165

**Revision: kvirc-7d8be0f-2015-10-02**

and drop the rest, nice clean and simple....

and the the sources date field can be removed also form the about box and we have less code to maintain and dont have to manually bump 

#define KVI_SOURCES_DATE "20150910"
#define KVI_SOURCES_DATE_NUMERIC 0x20150910

which btw also doing that manually is lame.

this is how it would look.

```

KVIrc 4.3.2 'Aria'

Runtime Info:
System Name: Linux 3.13.0-62-generic
System Version: #102-Ubuntu SMP Tue Aug 11 14:29:36 UTC 2015
Architecture: x86_64

Build Info:
Build Date: 2015-10-02 11:55:40 UTC
Revision: kvirc-7d8be0f-2015-10-02**
System Name: Linux-3.13.0-62-generic
CPU Name: x86_64
Build Command: /usr/bin/cmake
Build Flags: 
   MANDIR=man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler Name: /usr/bin/c++
Compiler Flags: N/A
```

DarthGandalf:
> > How do you propose to name the tag in the beginning of history? I chose to name it git, but it's not a very good name.
> 
> Do you really need any? lets leave that one for later... Why 5434 or any other SVN (I presume) related legacy version number?

Because `git describe` was used, and it outputs such number. `git describe` is a widely recognizable way to show the commit in a nice format.

> Do we really need to keep that?

Not really.

> Revision: kvirc-7d8be0f-2015-10-02
1. This is not "revision" anymore, because git doesn't recognize this format.
2. I'd prefer `kvirc-2015-10-02-7d8be0f`, so that lexicographic sort sorts them chronologically.
3. Yes, it's clean. And it's more correct than `git-5434-7d8be0f` proposed in #1548 because it doesn't look similar to output of `git describe` anymore.

un1versal:
> This is not "revision" anymore, because git doesn't recognize this format.
> I'd prefer kvirc-2015-10-02-7d8be0f, so that lexicographic sort sorts them chronologically.

So lets go with this? and call it **Version**

> Yes, it's clean. And it's more correct than git-5434-7d8be0f proposed in #1548 because it doesn't look similar to output of git describe anymore.

git describe should die in a fire.

DarthGandalf:
Or just `2015-10-02-7d8be0f`

un1versal:
> Or just 2015-10-02-7d8be0f

Well that was one of the first things that came to mind.

:+1: 

killing the 

#define KVI_SOURCES_DATE "20150910"
#define KVI_SOURCES_DATE_NUMERIC 0x20150910

is a must in this move though.

and using https://github.com/kvirc/KVIrc/issues/1566#issuecomment-139749165 would be great since its accurate replacement for KVI_SOURCES_DATE

un1versal:
@blindirc are you up for the challenge of turning this PR into the new proposed format and killing #define KVI_SOURCES_DATE?

wodim:
Why change things for the sake of it? `git describe` output is standard, contains a revision number, and will contain the version string in the future. Why fuck it up?

un1versal:
git describe will output whatever it is defined afaik.

if you tell it the version is santa-claus-v199928 thats what it will output

edit__ for sake of reference

I just tested this PR and indeed fixes https://github.com/kvirc/KVIrc/issues/1548 100% and breaks nothing I can see. except running git describe manually  still adds the g because git describe is stupid.

wodim:
You have already been told of the syntax:

`current tag`-`number of commits since that tag`-`g``hash`

What exactly is wrong with it?

un1versal:
> What exactly is wrong with it?

read up, and read https://github.com/kvirc/KVIrc/issues/1548

wodim:
All of them are problems that are only problems inside your head.

un1versal:
updated #1566 with https://github.com/kvirc/KVIrc/issues/1566#issue-105849834

#1548 has been closed, git describe behavior is besides real point of this, so working around git describe is wrong imo  and if it has issues should be fixed upstream.

We can discuss better ways forward in #1566 and forget the non productive noise and concentrate on more positive things.

Ill close this and hope we can brainstorm in #1566

PR #1654, opened 2015-10-04
Update all applicable Subversion references to Git

staticfox:
Went through and updated all Subversion references and links to Git, including tickets, comments, etc. Left the changelog alone though.


staticfox:
Seems like it's failing to build? I'll check when I get home. Just had to pr then run. ~1 hour

PR #1657, opened 2015-10-06
cmake dir and admin dir - various files minor fixes and updates

un1versal:
Its extracted from #1631 as this isnt related to the strings work and can be submitted independently.

Please review

### DIFF
--- a/admin/makedist.sh
+++ b/admin/makedist.sh
@@ -45,7 +45,7 @@ for i in $*; do
 				DOMD5SUM="yes"
 			;;
 		--*)
-			echo "ERROR: Unrecognized option $i"
+			echo "ERROR! Unrecognized option $i"
 			usage
 			exit
 			;;
@@ -60,19 +60,19 @@ for i in $*; do
 done
 
 if [ -z "$SOURCETREEDIR" ]; then
-	echo "ERROR: Source tree directory missing"
+	echo "ERROR! Source tree directory missing"
 	usage
 	exit
 fi
 
 if [ -z "$KVIRCVERSION" ]; then
-	echo "ERROR: KVIrc version missing"
+	echo "ERROR! KVIrc version missing"
 	usage
 	exit
 fi
 
 if [ ! -d $SOURCETREEDIR ]; then
-	echo "ERROR: The source tree parameter doesn't identify a directory"
+	echo "ERROR! The source tree parameter doesn't identify a directory"
 	usage
 	exit
 fi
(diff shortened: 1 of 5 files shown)


admin/makedist.sh
```diff
@@ -45,7 +45,7 @@ for i in $*; do
 				DOMD5SUM="yes"
 			;;
 		--*)
-			echo "ERROR: Unrecognized option $i"
+			echo "ERROR! Unrecognized option $i"
```
DarthGandalf, line 48: Why?
"error: foo" is a common pattern

cmake/KVIrc.nsi.cmake
```diff
@@ -16,9 +16,9 @@ Name "@NICENAME@"
 !define URL_UPDATE 'http://www.kvirc.net/'
 !define PUBLISHER 'Szymon Stefanek and The KVIrc Development Team'
 
-; Git release, eg: KVIrc-4.0.0-rc3-git-4423-5ac91f3.exe
+; Git release, e.g. KVIrc-4.3.2-dev-git-5465-ge352609-x86_64.exe
```
DarthGandalf: -x86_64 doesn't belong here. Do you see it added below?
In appveyor config, it's appended to exe name in the very end of build.

admin/makedist.sh
wodim, line 48: And this is just an internal script, I don't understand why would you change things just for the sake of it... again.

un1versal, line 48: > Why?
> "error: foo" is a common pattern

its either `Error: unrecognized option $i` lowercase u -  or as I put it.

> And this is just an internal script, I don't understand why would you change things just for the sake of it... again.

Does matter. In your opinion, its OK to leave everything in an inconsistent state even if its wrong, just because you dont care about it doesnt make it right.

Bad habits translate to other areas of program.

cmake/KVIrc.nsi.cmake
un1versal: OK removing. twas a c&p monster issue.

pragmaware:
OK. Though maybe it's nicer to prefer the ERROR: <small_letter> pattern though
(as it's spread everywhere around the sources and ERROR! looks too "dramatic" :D)

un1versal:
> OK. Though maybe it's nicer to prefer the ERROR: pattern though
> (as it's spread everywhere around the sources and ERROR! looks too "dramatic" :D)

Next time review with that and Ill change it before merge, as I agree it should be consistent.

Now fixed in 73a5aeaf61f8d7c47b9e89f81c200c44c09e882f

PR #1658, opened 2015-10-06
Random aesthetic cleanups

staticfox:


wodim:
As much as I appreciate your contributions, I don't understand why would we pollute blame with such changes.

un1versal:
So we should leave things all over the place instead?

PR #1659, opened 2015-10-07
Update IRCd protocols

staticfox:
Add IRCds Unreal 3.2.x and ircd-hybrid
Base ircd-darenet, nemsis, ircd-seven, oftc-hybrid, snircd, and plexus off of their upstream forks.

If you want me to squash all of this in to one commit, please let me know.

### DIFF
--- a/src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
+++ b/src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
@@ -230,7 +230,9 @@ kvi_u32_t KviIrcConnectionServerInfo::modeFlagFromModeChar(QChar c)
 void KviIrcConnectionServerInfo::setServerVersion(const QString & version)
 {
 	if(m_pServInfo) delete m_pServInfo;
-	if(version.contains("unreal",Qt::CaseInsensitive))
+	if(version.contains("unreal3.2",Qt::CaseInsensitive))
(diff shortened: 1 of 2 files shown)


un1versal:
~~If Im looking at these right, freenode protocol could use an update       switch(mode)  looks a bit outdated to me, https://freenode.net/using_the_network.shtml~~

Or maybe Im confused idk. I swear I seen good modes show up as unrecognized at some stage.

staticfox:
> I was more inclined to make these all capitals like AUTH instead of auth.
> 
> Thoughts?

Why?...

un1versal:
> Why?...

https://www.quakenet.org/help/q-commands/auth
http://archives.theonering.net/barlimans/docs/4.html

isnt that it?

staticfox:
Not really, auth in this case just means to be identified with services. While I guess it would be more correct to read as `Must be identified with services to change nick or speak`, just saying `Need auth` gets the point across in way less words.

un1versal:
I was reading things like 

https://www.quakenet.org/help/q-commands/auth
http://archives.theonering.net/barlimans/docs/4.html

They mention auth as AUTH, which I guess is the command, so my question is in other areas where command is meant does it mean we should use AUTH or just mix two auth and AUTH as -> auth interchangeably? the latter being wrong IMO.

@pragmaware thoughts?

PR #1666, opened 2015-10-09
Add Irchighway server

Havokdan:
Add Irchighway server


un1versal:
Hope #1667 will serve you just as well and welcome to Opensource.
Closed in favor of the above/

2 reasons

1) IMO adding only SSL port for a network in a client that may very be used in a country where using encryption may not be legal, is not ideal

So we should add at least a non SSL option if Network supports it.

2) Alphabetical order, and no messy multiple commits fixing the main commit and the the copy and paste monsters for a rush job.

PR #1668, opened 2015-10-10
client: Revamp user mode support

staticfox:
serverinfo: Add all user modes with descriptions to each IRCd's applicable
class. Create tables for user mode requirements to determine whether or not
the user mode can be applied to the user.
coreactions: Display information on each user mode in the drop down menu
for user mode selection. Check if the user mode can be set with the users
current permissions, if it cannot then gray it out.

---

Alright so I'll start this off with a pure before & after:
### +s check

You will notice that I have a check within core actions to make sure user mode +s is supported by the IRCd. Some IRCds do not support +s for whatever reason, so for that we want to make sure it exists. The same could be done for +i and +w, however, I have yet to find a server that _doesn't_ support those modes.
### User mode enabled

Most user modes will be grayed out. This is due to that fact that user modes in general are more geared towards IRC Operators for server information. I have created some functions that determine whether or not the mode can be set based on the IRCd's programming. This is also to avoid having to rewrite "(staff only)" or "(server only)" within each string. Cut and dry, if it's grayed out then you can't use it. A further explanation can be determined with the mode's description.
### Rewriting "indexOf('+');"

This broke support for the 3 core modes that KVIrc appears to favor (+s, +i, and +w). I have gone ahead and standardized the strings to follow the same format as the channel mode window ("modechar: Description here"). This makes it simpler for the user to determine what each mode does, as the + might confuse users whether or not it is set. Why we have the sanity check there in the first place is beyond me as this is hard coded data calling hard coded data.
### getUserModeDescription()

This has been added to all existing IRCds that KVIrc currently supports. I have gone though each IRCd's help files and source code to verify accuracy.
### Add getUserModeRequirement()

This has been added to determine whether or not the user mode can be set. The user mode gets passed to the function and returns the user mode (if any) required to set. If it returns 1, then it cannot be changed (server managed only). If it returns 0 then it can be set. If it returns a QChar then it checks to see if the user has it set. If they do then the option is enabled. If not then it is grayed out.
### Add getNeedsOperToSetS()

Some IRCds require Oper to set user mode +s. I have added this bool function to determine whether or not the IRCd requires +s or not. If the function returns true then core actions checks for user mode +o. If the user does not have +o then the mode is grayed out.

### DIFF
--- a/src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
+++ b/src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
@@ -63,7 +63,6 @@ bool KviIrcConnectionServerInfo::isSupportedChannelType(QChar c)
 	return m_szSupportedChannelTypes.contains(c);
 }
 
-
 void KviIrcConnectionServerInfo::addSupportedCaps(const QString &szCapList)
 {
 	m_bSupportsCap = true;
@@ -264,7 +263,6 @@ void KviIrcConnectionServerInfo::setServerVersion(const QString & version)
 		m_pServInfo = new KviBasicIrcServerInfo(this, version);
 }
 
-
 KviBasicIrcServerInfo::KviBasicIrcServerInfo(KviIrcConnectionServerInfo * pParent, const QString & version)
 {
 	m_szServerVersion = version;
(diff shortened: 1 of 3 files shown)


un1versal:
I always found this somewhat lacking, and am glad to see someone taking a more detailed approach to extending making it _smart detect applicable modes_ and _label mode description_

my big huge fat  +1 for this type improvement. - Priority for this should be Yesterday.

I cant comment on code, but we discussed in IRC, my feedback on the strings is being done.

Ill make some comments inline mostly to the consistency of capitalization of same terms where you should pick one and apply throughout the exact same.
Mixing ANY descriptions sometimes more than 2 different ways for same word or term is a huge nono for me, its confusing like hell for noobs and all users and probably looks unprofessional.

(capitalization not withstanding as doesnt apply in these examples as its not determined by descriptions choices)
Fine examples are, administrator, admin and op, oper, operator

I think this PR is a must with those minors resolved and consistency issues addressed.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+	{
+		case 'O': return __tr2qs("O: Local IRC operator (LOCOP)"); break;
+		case 'a': return __tr2qs("a: Server administrator"); break;
+		case 'c': return __tr2qs("c: Recipient for cconn messages"); break;
+		case 'd': return __tr2qs("d: Recipient for server debug notices"); break;
+		case 'f': return __tr2qs("f: Recipient for full server notices"); break;
+		case 'i': return __tr2qs("i: Invisible"); break;
+		case 'k': return __tr2qs("k: Recipient for server KILL messages"); break;
+		case 'n': return __tr2qs("n: Recipient for nick changes"); break;
+		case 'o': return __tr2qs("o: IRC operator (OPER)"); break;
```
un1versal: I like that focus is brought to OPER, not entirely sure if we can get away with OP only but I like the capitalization.

I probably wouldnt use brackets anywhere instead separate with hyphen `IRC operator - OPER` simply because its cleaner and more readable. but not a show stopper, Ill leave that to you.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
 
+////////////////
+// User modes
+////////////////
+
+const QString & KviBasicIrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'O': return __tr2qs("O: Local IRC operator (LOCOP)"); break;
```
un1versal: See below comment about brackets.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+const QString & KviIrcdSevenIrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'O': return __tr2qs("O: IRC help operator"); break;
+		case 'Q': return __tr2qs("Q: Prevents you from being affected by channel forwarding"); break;
+		case 'S': return __tr2qs("S: Network service"); break;
+		case 'Z': return __tr2qs("Z: Connected over SSL"); break;
+		case 'h': return __tr2qs("h: Marks you as a helper in /stats p"); break;
+		case 'p': return __tr2qs("p: Enable oper overrides"); break;
```
un1versal, line 381: Above you chose to capitalize OPER, this should be carried over consistently. either you capitalize all instances of the same or you dont.

My preference is the capitalized version, but again I leave that to you, as long as its consistent.

A server message for instance ....

[sinisalo.freenode.net]: **\* You need oper and helpop flag for +h

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+	}
+	return KviHybridServerInfo::getUserModeDescription(mode);
+}
+
+const QString & KviBahamutIrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'A': return __tr2qs("A: Server administrator"); break;
+		case 'C': return __tr2qs("C: Only accept private messages from users who share common channels"); break;
```
un1versal: `"C: Only accept private messages from users who share common channels`

thats a mouth full, I wonder if we can say this with less words.

Perhaps a feeble attempt - need way more coffee.

`"C: Only accept private messages from users in common channels`

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'F': return __tr2qs("F: Can bypass the ircd's recvq throttling"); break;
+		case 'I': return __tr2qs("I: Hide oper status"); break;
+		case 'K': return __tr2qs("K: Can see U:lined kill messages"); break;
+		case 'R': return __tr2qs("R: Receive private messages from registered nicks only"); break;
```
un1versal: Other likeminded sentences start with Only, again same consistency should be observed.

`Only receive private messages from registered nicks`

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'f': return __tr2qs("f: Recipient for flood warnings"); break;
+		case 'g': return __tr2qs("g: Recipient for GLOBOPS notices"); break;
+		case 'h': return __tr2qs("h: Available for help (HelpOp)"); break;
+		case 'j': return __tr2qs("j: Recipient for rejected client notices"); break;
+		case 'm': return __tr2qs("m: Recipient for spambot notices"); break;
+		case 'n': return __tr2qs("n: Recipient for routing notices"); break;
+		case 'r': return __tr2qs("r: Registered"); break;
+		case 's': return __tr2qs("s: Recipient for server KILL messages"); break;
+		case 'x': return __tr2qs("x: Squelch with notice"); break;
+		case 'y': return __tr2qs("y: Can see certain information requests (e.g. /stats)"); break;
```
un1versal, line 439: Finally someone who can correctly use a **proper** abbreviation of  e.g.

:+1:  hahaha

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'C': return __tr2qs("C: Co-Admin"); break;
+		case 'F': return __tr2qs("F: Override filter settings"); break;
+		case 'G': return __tr2qs("G: Filter out bad words"); break;
+		case 'H': return __tr2qs("H: Hide oper status"); break;
+		case 'I': return __tr2qs("I: Idle time hidden from WHOIS"); break;
+		case 'N': return __tr2qs("N: Network administrator"); break;
+		case 'R': return __tr2qs("R: Receive private messages from registered nicks only"); break;
+		case 'S': return __tr2qs("S: Network service"); break;
+		case 'T': return __tr2qs("T: Prevent CTCPs"); break;
+		case 'V': return __tr2qs("V: Connected over WebTV"); break;
```
un1versal, line 459: whats WebTV?

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'W': return __tr2qs("W: Recipient for notices on WHOIS"); break;
+		case 'a': return __tr2qs("a: Services administrator"); break;
+		case 'd': return __tr2qs("d: Deaf"); break;
+		case 'g': return __tr2qs("g: Can send & read LOCOPS and GLOBOPS"); break;
+		case 'h': return __tr2qs("h: Available for help (HelpOp)"); break;
```
un1versal: Ive never been a fan of **glueing** words together like Halfop but here we start coming into that capitalization of OPER https://github.com/kvirc/KVIrc/pull/1668/files#r41694313 I asked if we could maybe use OP.

Should we maybe consider that this could be Help-OP or Help-Op at very least and use OP or Op and OPER or Oper and not mix and match in any other descriptions?

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'p': return __tr2qs("p: Channels hidden from WHOIS"); break;
+		case 'q': return __tr2qs("q: Immune to kicks (except U-Line)"); break;
+		case 'r': return __tr2qs("r: Registered"); break;
+		case 't': return __tr2qs("t: Using a vhost"); break;
+		case 'v': return __tr2qs("v: Recipient for infected DCC Send Rejection notices"); break;
```
un1versal: `Recipient for infected DCC Send Rejection notices`

was this meant to be

`Recipient for infected DCC send rejection notices`?

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+}
+
+const QString & KviInspIRCdIrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'c': return __tr2qs("c: Prevent CTCPs"); break;
+		case 'd': return __tr2qs("d: Deaf"); break;
+		case 'g': return __tr2qs("g: Server side ignore"); break;
+		case 'h': return __tr2qs("h: Available for help (HelpOp)"); break;
```
un1versal: https://github.com/kvirc/KVIrc/pull/1668/files#r41694517

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+	switch(mode.unicode())
+	{
+		case 'C': return __tr2qs("C: Can see CHATOPS notices"); break;
+		case 'D': return __tr2qs("D: Has seen DCCALLOW warning message"); break;
+		case 'F': return __tr2qs("F: Recipient for far connect/quit notices"); break;
+		case 'P': return __tr2qs("P: Services administrator"); break;
+		case 'S': return __tr2qs("S: Network service"); break;
+		case 'W': return __tr2qs("W: Recipient for notices on WHOIS"); break;
+		case 'Z': return __tr2qs("Z: Services root administrator"); break;
+		case 'a': return __tr2qs("a: Services operator"); break;
```
un1versal, line 510: Im split here, OPER, op and now operator... argh,,, how many different ways to say the same thing... Consistency 0 - Anarchy 1

me I would stick to OP, because anyone who reads IRC introductions docs or has used IRC for any time, knows or should know what OP stands for, and only lazy people who dont bother with anything wont know, and if they complain they should be shot for being lazy.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'e': return __tr2qs("e: Recipient for blocked DCC notices"); break;
+		case 'f': return __tr2qs("f: Recipient for exess flood notices"); break;
+		case 'g': return __tr2qs("g: Recipient for GLOBOPS notices"); break;
+		case 'h': return __tr2qs("h: Available for help (HelpOp)"); break;
```
un1versal: https://github.com/kvirc/KVIrc/pull/1668/files#r41694517

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'm': return __tr2qs("m: Recipient for spam warning notices"); break;
+		case 'n': return __tr2qs("n: Recipient for NETINFO and routing notices"); break;
+		case 'p': return __tr2qs("p: Protected irc operator"); break;
```
un1versal: https://github.com/kvirc/KVIrc/pull/1668/files#r41694558

and its IRC not irc

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+}
+
+const QString & KviIrcdRatboxIrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'C': return __tr2qs("C: Prevent CTCPs"); break;
+		case 'D': return __tr2qs("D: Deaf"); break;
+		case 'Z': return __tr2qs("Z: Recipient for oper spy notices"); break;
+		case 'a': return __tr2qs("a: Server administrator"); break;
```
un1versal, line 531: Admin, admin or Administrator administrator, mixing both is argh.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'Z': return __tr2qs("Z: Allows the ability to send OPERWALL messages"); break;
+		case 'a': return __tr2qs("a: Can see all users and channels"); break;
+		case 'b': return __tr2qs("b: Recipient for spam warning notices"); break;
+		case 'c': return __tr2qs("c: Recipient for connect and netjoin notices"); break;
+		case 'e': return __tr2qs("e: Registered"); break;
+		case 'f': return __tr2qs("f: Recipient for full I-Line notices"); break;
+		case 'h': return __tr2qs("h: Sets user in high priority mode"); break;
+		case 'j': return __tr2qs("j: Allows the use of auto D-Lines"); break;
+		case 'l': return __tr2qs("l: Recipient for new channel creation notices"); break;
+		case 'm': return __tr2qs("m: Cannot be KICKed, DEOPed, or D-Lined"); break;
```
un1versal: `Cannot be KICKed, DEOPed, or D-Lined`

This is not technically wrong but... I would use Can't instead of Cannot, again its a consistency thing with other areas or program not here specific.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'y': return __tr2qs("y: Can see certain information requests (e.g. /stats)"); break;
+	}
+	return KviBasicIrcServerInfo::getUserModeDescription(mode);
+}
+
+const QString & KviUnreal32IrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'A': return __tr2qs("A: Server administrator"); break;
```
un1versal, line 448: administrator - or admin pick one same for all other entries.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'y': return __tr2qs("y: Can see stats/links/admin requests"); break;
+	}
+	return KviBasicIrcServerInfo::getUserModeDescription(mode);
+}
+
+const QString & KviIrcdSevenIrcServerInfo::getUserModeDescription(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'O': return __tr2qs("O: IRC help operator"); break;
```
un1versal: is it op or oper or operator.... cant this stuff be any more higgledy-piggledy

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
DarthGandalf: or "from users who share channels with you"

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case '@': return __tr2qs("@: allows the user to change their host to anything with the SETHOST command"); break;
+	}
+	return KviBasicIrcServerInfo::getUserModeDescription(mode);
+}
+
+////////////////
+// UMODE Requirements
+////////////////
+
+QChar KviBasicIrcServerInfo::getUserModeRequiement(QChar mode)
```
DarthGandalf: typo in function name

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+}
+
+QChar KviBahamutIrcServerInfo::getUserModeRequiement(QChar mode)
+{
+	switch(mode.unicode())
+	{
+		case 'A': return 'A'; case 'O': return 'O';
+
+		case 'S': case 'W': case 'X': case 'a':
+		case 'r': case 'w': case 'x': return 1;
```
DarthGandalf, line 712: Why do you return a number as QChar?
In the same function you return 'o' instead

staticfox, line 712: As I said in my description, I return 0 for settable, a QChar for the mode required to set it, and a 1 for never settable. I couldn't pick a random non existent mode since projects like hyperion decide to use 59 modes, run out numbers, then switch to symbols. I'm open to better ideas, but but was the easiest implementation I could think of.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox: Oops.. At least I consistently spelled it wrong! heh.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 'O': return __tr2qs("O: Local IRC operator (LOCOP)"); break;
+		case 'a': return __tr2qs("a: Server administrator"); break;
+		case 'c': return __tr2qs("c: Recipient for cconn messages"); break;
+		case 'd': return __tr2qs("d: Recipient for server debug notices"); break;
+		case 'f': return __tr2qs("f: Recipient for full server notices"); break;
+		case 'i': return __tr2qs("i: Invisible"); break;
+		case 'k': return __tr2qs("k: Recipient for server KILL messages"); break;
+		case 'n': return __tr2qs("n: Recipient for nick changes"); break;
+		case 'o': return __tr2qs("o: IRC operator (OPER)"); break;
+		case 'r': return __tr2qs("r: User with restricted connection (or recipient for messages about rejected bots)"); break;
```
un1versal, line 293: `User with restricted connection (or recipient for messages about rejected bots)`

Can we say

`Restricted connection (or recipient for messages about rejected bots)`?

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
```diff
@@ -275,6 +273,566 @@ KviBasicIrcServerInfo::~KviBasicIrcServerInfo()
...
+		case 's': return __tr2qs("s: Recipient for server notices"); break;
+		case 'w': return __tr2qs("w: Recipient for WALLOPS messages"); break;
+		case 'y': return __tr2qs("y: Recipient for spy notices"); break;
+		case 'z': return __tr2qs("z: Recipient for oper WALLOP messages"); break;
```
un1versal, line 297: op operator pick one.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox: Reason I prefer parenthesis is because it identifies it as a separate synonym whereas a hyphen denotes it as a type. For example, on Rizon we have SWHOIS strings that return "is an IRC Operator - Network Administrator" or "is an IRC Operator - Server Administrator". IMO a hyphen would identify a "type" of operator, whereas here we're just showing the user the short hand version that people generally know as well.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox: I'm going to just use the lowercase `helpop` string here, since we have `locop` and `oper` and they all carry the same definition structure, and that is also how IRCds refer to them as well (although unreal uses `HelpOp` but for the sake of consistency, I'm lowercasing it since it's not a command.)

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox: That was from it's help file, however, I'm going to change it to `Recipient for infected DCC SEND rejection notices` as `DCC SEND` is a command and the rest is the rest.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox, line 510: This is a services class type that has been defined since IRC Services (I'm pretty sure). So Services operator, Services root administrator, Services administrator are correct.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox, line 712: In fairness, commenting this probably would have been a good idea. In my mind I visualized a table with 3 rows. The top being modes that can be set if the user has it (i.e. you can't set +o, you have to /oper for it but you can unset it after you get it). The middle row being modes that you can never touch (server set such as connected over SSL, WEBIRC, etc.) and the third row for requiring oper modes like +o, or in same cases, services modes.

If you're speaking purely on the fact that I'm returning an int within a QChar function, then I can't think of a better work around as I said in my first comment.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
un1versal: thats fine, i prefer a more readable format

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox, line 531: I've changed everything to administrator where applicable, the only two places I'm not going to change it are `Server/Network/Tech Admin only channel` because `Tech Admin` is an Unreal/Criten oper type, and `Co-Admin` because that's what Unreal officially refers to it as.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
staticfox, line 459: It had something to do with MSNTV, it's basically an Unreal thing. I'm not gonna touch it, support for it is said to be dropped within the next few years. I think it only applied this mode to users using WebTV and changed notices to privmsgs or something.

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
un1versal: \o/ yes...

src/kvirc/kernel/KviIrcConnectionServerInfo.cpp
un1versal, line 510: I mean op vs operator arent they the same thing?

we are using irc operator is oper OK thats fine, but the description operator is op in the same sense as op no? so irc op (oper) 

Im just confused by the seemingly sensless mixing of abreviations and full words that all seem to be the same

staticfox, line 510: In context, they aren't. Like I said, especially for things like this, `Services operator` is the specific definitive type of class.

https://github.com/anope/anope/blob/2.0/data/example.conf#L836
https://github.com/anope/anope/blob/2.0/data/example.conf#L860
https://github.com/anope/anope/blob/2.0/data/example.conf#L871

un1versal, line 510: yes, I see, thanks for these tidbits, I guess then unless they agree to unify their terminology upstream, the ones downstream this type stuff are doomed to spaghetti soup.

None the less we can hopefully have some wriggle room to improve and make wherever possible things more consistent.

staticfox:
Anything else anyone can think of?

un1versal:
if you squash and rebase Ill have another gander, but from what I see looks great. :+1:  :+1: 

Would be nice to get @pragmaware's feedback though.

staticfox:
Rebased and squashed

PR #1670, opened 2015-10-14
numeric: Parse 396 and output it as a snotice

staticfox:
### after

I originally got the idea from OFTC, but this really does make sense to me. Not only does it look nicer but if my host was changed (cough cough oper abuse :smile:) I'd also want to know right then and there. (assuming boolServerNoticesToActiveWindow was set). But really this is just more of an excuse to add some formatting to some highly active numerics that are function-less. Something I also want to do is add a KVS event for OnMeHostChange. (For things like idlerpg or other bots logging you out on host change)


un1versal:
Looks OK to me :)

PR #1673, opened 2015-10-17
Add options for kick message output

staticfox:
So it works fine, but I feel like it could be optimized. Not thinking straight today in light of recent events :disappointed:. Feedback?


wodim:
I don't like this option.

staticfox:
Hmm alright. Only added this due to the comments I saw when skimming through this. I personally would find it useful since I'm in > 150 channels across several networks, so when I hear my kick sound it takes me a few minutes to find out where I was kicked from, heh.

un1versal:
+1 the current kick message is _inefficient_ to say the very least, by default the kick with reason field is populated by the username and that's it imo, not even right since you already get that info on the console. So maybe populate that message with something meaningful by default or have some sort of list. (but that client internal and not all clients have the same  system).

The situation you describe which is a perfectly acceptable scenario, I dont see what the problem here is maybe it can be achieved by catching the event and having some sort of event based reaction?

I also find wodims feedback unhelpful and vague.

wodim:
My feedback is that we already have enough options and this is easily scriptable.

un1versal:
Hence I suggested some event based solution.

staticfox:
Closing in favor of better implementation.

un1versal:
Coming to a theater near you, SOON!

PR #1676, opened 2015-10-18
KviDefaultScript: fix  #1675 and improve dialog messages.

### LINKED ISSUE #1675: Confirmation dialog of restoring default script has too many buttons
It says something like "Your copy of default scripts is fresh. Restore anyway? Yes/No/Cancel" (I see the message on Russian, so not sure how exactly it looks like on English)

The buttons No and Cancel do the same thing.

664c14a on Gentoo

un1versal:
@DarthGandalf your MERGE button @blindsighttf2 I think messages work like this. thoughts?

### DIFF
--- a/src/kvirc/kernel/KviDefaultScript.cpp
+++ b/src/kvirc/kernel/KviDefaultScript.cpp
@@ -408,11 +408,11 @@ KviDefaultScriptDialog::KviDefaultScriptDialog()
 	QGridLayout * pLayout = new QGridLayout(this);
 
 	QLabel * pLabel = new QLabel(this);
-	pLabel->setText(__tr2qs("You are about to restore the default script.\nAny script changes you have made will be erased if you\nrestore all stuff or \"Clear personal data\" is selected.\n\nDo you wish to proceed?\n"));
+	pLabel->setText(__tr2qs("You are about to restore the default script.\nAny script changes you have made will be erased if\nrestore \"All\" or \"Clear Custom Scripts\" are selected.\n\nDo you wish to proceed?\n"));
 	pLayout->addWidget(pLabel,0,0,1,4);
 
 	// Construct the advanced widget
-	m_pAdvanced = new QGroupBox(__tr2qs("Select default script's elements to restore"),this);
+	m_pAdvanced = new QGroupBox(__tr2qs("Select the Items to Restore"),this);
 	m_pAdvanced->hide();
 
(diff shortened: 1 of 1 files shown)


staticfox:
Still think you should consider looking at L240 IMO

un1versal:
I did, I dont really see any added benefit. up-to-date doesn't need a reinforcement. The message seems clear to me.

src/kvirc/kernel/KviDefaultScript.cpp
```diff
@@ -408,11 +408,11 @@ KviDefaultScriptDialog::KviDefaultScriptDialog()
 	QGridLayout * pLayout = new QGridLayout(this);
 
 	QLabel * pLabel = new QLabel(this);
-	pLabel->setText(__tr2qs("You are about to restore the default script.\nAny script changes you have made will be erased if you\nrestore all stuff or \"Clear personal data\" is selected.\n\nDo you wish to proceed?\n"));
+	pLabel->setText(__tr2qs("You are about to restore the default script.\nAny script changes you have made will be erased if you\nrestore \"All\" or \"Clear custom scripts\" is selected.\n\nDo you wish to proceed?\n"));
 	pLayout->addWidget(pLabel,0,0,1,4);
 
 	// Construct the advanced widget
-	m_pAdvanced = new QGroupBox(__tr2qs("Select default script's elements to restore"),this);
+	m_pAdvanced = new QGroupBox(__tr2qs("Select the Items to Restore"),this);
```
DarthGandalf, line 415: Why Items Are So Capital?

un1versal, line 415: According to https://techbase.kde.org/Projects/Usability/HIG/Capitalization

> Use title capitalization in the following cases:
> 
>  Group box / group line labels

IF thats not the case I can remove the Camel Case.

staticfox, line 415: Shouldn't it be `Items to restore` then?

un1versal, line 415: IMO thats a group box and also the items in there must be capitalized _APPARENTLY_ according to https://techbase.kde.org/Projects/Usability/HIG/Capitalization which was indicated to me in #1631

"the" article and "to" preposition --are exceptions since they are of 5 letters or less as the guide suggests.

So is this right or not?

I also fixed the **Clear Custom Scripts** as that is part of this lot

PR #1678, opened 2015-10-18
Clean up trailing whitespaces and excess newlines

staticfox:
Making sure appveyor and travis work first


un1versal:
If it builds wack it in

PR #1679, opened 2015-10-18
light themed splash and setup label

un1versal:
@pragmaware is this better?

EDIT
It turns out, looking at older 3.x KVIrc that `kvi_splash_overlay.png` was a [filled texture](https://github.com/kvirc/KVIrc/blob/3.4/data/pics/kvi_splash_overlay.png) BUT [in 4.2.0](https://github.com/kvirc/KVIrc/blob/4.2/data/pics/kvi_splash_overlay.png) which where I based current in was no longer filled, so something weird happened there. No idea.

I have restored that filled overlay, which I presume will do some progress type animation, its hard t test in Linux because it happens almost instantly.

EDIT2
Yep I had a quick test on Windows and the progress on splash, now works as intended :) Have to love slow OS's...


staticfox:
Aww, I always liked the dark theme splash. Looks nice though :smiley:

DarthGandalf:
I like your new dark one...

un1versal:
> Aww, I always liked the dark theme splash. Looks nice though :smiley:
> 
> I like your new dark one...

Yep I do also, the wolf howling to the moon (singing an Aria) and there'ss a cartoon She Wolf character named Aria somewhere.

But.... When asked pragma said it was too heavy and depressing and the splash should be light and happy... And that Aria was meant to be Air in Italian or something like that --so meh.

pragmaware:
Well... it was just my opinion. If you people like the wolf, then let's have the wolf :D

staticfox:
Yay! :heart:

un1versal:
Closed :)

PR #1680, opened 2015-10-18
Fix SASL plain auth; Fix #1677

### LINKED ISSUE #1677: SASL protocol error
KVIrc is sending an extra null byte after the password when authenticating via SASL PLAIN:

```
[18:56:33] AUTHENTICATE ***redacted***
[18:56:34] :local.test.redatcted 904 061AAAAAW :SASL authentication failed
```

This is not in accordance with [RFC 4616](https://tools.ietf.org/html/rfc4616), and causes authentication to fail with some services (Anope).

ctrlaltca:
This PR removes the spurious 0x00 appended to the end of the base64-encoded string used in SASL plain authentication.

### DIFF
--- a/src/kvilib/net/KviSASL.cpp
+++ b/src/kvilib/net/KviSASL.cpp
@@ -35,7 +35,7 @@ namespace KviSASL
 	{
 		if(szIn=="+")
 		{
-			int answerLen = 3 + (2 * baNick.size()) + baPass.size();
+			int answerLen = 2 + (2 * baNick.size()) + baPass.size();
 			char * answer = (char *) KviMemory::allocate(answerLen);
 			char * answer2 = answer;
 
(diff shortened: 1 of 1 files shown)


un1versal:
:+1: I use SASL but never came across this issue, but none of the servers are the same as the issue report.

un1versal:
Ive tested this PR and seems to continue to work  on the networks I use SASL with so :+1: 

Just one thing, ive no idea why theres a semi colon at https://github.com/ctrlaltca/KVIrc/blob/fix_1677/src/kvilib/net/KviSASL.cpp#L62 I removed that on my test.

ctrlaltca:
Yup, that semicolon is not needed, but i don't think it hurts anything.

un1versal:
> Yup, that semicolon is not needed, but i don't think it hurts anything.

Hurts my eyes :)

Anyway, cant comment on code changes, but here is my +1 for merging

PR #1681, opened 2015-10-18
Fix fixup_bundle() on osx + qt >= 5.5

ctrlaltca:


DarthGandalf:
I think, need to add one more build to travis, currently it uses only qt4 on osx.

ctrlaltca:
> I think, need to add one more build to travis, currently it uses only qt4 on osx.

You are right; also, as an user, i see no reason to choose qt4 over qt5 on osx.

staticfox:
> You are right; also, as an user, i see no reason to choose qt4 over qt5 on osx.

Perhaps we can use this as an excuse to start slowly migrating away from qt4?

DarthGandalf:
@ctrlaltca `brew install qt` vs `brew install qt5`?

ctrlaltca:
@DarthGandalf exactly; also:
- cmake => 2.8.12 is needed, but homebrew already provides 3.x, so no problem;
- we need to pass -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.whatever/ to cmake

Example: https://github.com/Cockatrice/Cockatrice/blob/master/travis-compile.sh

DarthGandalf:
I prefer `brew --prefix qt5`, it looks a bit nicer than `echo`

Also I see, you do the same for linux, but `source /opt/qt55/bin/qt55-env.sh` should be enough with beineri's ppa

ctrlaltca:
Yep, `source /opt/qt55/bin/qt55-env.sh` should suffice
I had to add the LD_LIBRARY_PATH thing to have cpack determine the binary dependencies for the debian package.

un1versal:
is this good to go in?

PR #1661, opened 2015-10-08
Split from #1631: cpp more strings and doc typos and fixes - need helpfull feedback

un1versal:
This is also split from #1631 but this contains a more controversial subset of #1660 
What I Tried to do here if it isnt obvious, is in addition to some typos and similar fixes to #1660, to capitalize where possible commands/actions or service but some maybe mixed up at this time.
### Notes:

Please on the cases that fit into the categories below **just add with your review** this should be **lower case** and this **can be all capitals** where applicable. I will fix and push the changes as and when reviews come in.

AUTH command vs auth "authorize(d)"
Some of these may be incorrect where **AUTH means just authorized** and not AUTH the command.
#### Excessive... use of... ellipses...

Oops... lines IMO its Oops!

```
oops
o͝ops,o͞ops/
exclamation informal
exclamation: oops

    used to show recognition of a mistake or minor accident, often as part of an apology.
    "“Oops! I'm sorry. I just made you miss your bus.”"
```

Also most sentences dont seem to be very reticent in nature.
I purposefully capitalized the first letter after ellipses, to bring attention to the fact these have to be changed after review to something appropriate and correct.

Rest review as normal with helpful feedback to aid proper correction.
### Please review and provide helpful feedback

Thank you.

PR #1684, opened 2015-10-21
OptionsWidget_(various): capitalization fixes

un1versal:
Using https://techbase.kde.org/Projects/Usability/HIG/Capitalization as a rule more or less.


staticfox:
Perhaps this and #1661 should both have their commits squashed in to one? Having a separate commit per file when the update (capitalization/spelling/grammar) is the exact same seems a bit silly IMO.

un1versal:
I separated this (2 of these and added the rest) from #1661 to commit them separately. This one is going in tomorrow. The other IDK theres a lot more in there. meh.

As for how many commits there are in each PR, well, does it matter? When its ready to merge I can squash them anyway just before merge so its not a big deal.

PR #1693, opened 2015-10-25
encoding: Set default encoding to UTF-8

### LINKED ISSUE #1550: Change default encoding to Smart UTF-8 [System encoding]
Since mIRC changed to UTF-8 by default (years ago), most of KVIrc users see gibberish instead of properly accented letters, since by default KVIrc uses System encoding (at least on Windows), similar to this:

image

mIRC is detecting & resolving this as far as i know, so only KVIrc users are affected by this.

Using ISO-\* is unnecessary anymore, UTF-8 covers more than enough languages for our use-case.
http://www.unicode.org/standard/supported.html

It may need to handle exceptions in some cases where UTF-8 doesn't support a particular language, but I believe these could be hard-coded in KVIrc easily.

On Freenode we talked about encoding problems and wodim propsed to change the default encoding to Smart UTF-8 [System encoding], that would solve this kind of problems without breaking everything else. 

I think this would be a great idea.

Additional suggestions and thoughts are welcome!

staticfox:
Closes #1550 

Thoughts?

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -400,7 +400,7 @@ KviStringOption g_stringOptionsTable[KVI_NUM_STRING_OPTIONS]=
 	STRING_OPTION("IconThemeSubdir","",KviOption_sectFlagUser | KviOption_groupTheme | KviOption_resetReloadImages | KviOption_encodePath),
 	STRING_OPTION("IncomingPath","",KviOption_sectFlagUser | KviOption_encodePath),
 	STRING_OPTION("MyAvatar","",KviOption_sectFlagUser),
-	STRING_OPTION("DefaultTextEncoding","",KviOption_sectFlagFrame),
+	STRING_OPTION("DefaultTextEncoding","UTF-8",KviOption_sectFlagFrame),
 	STRING_OPTION("WinampTextEncoding","",KviOption_sectFlagFrame),
 	STRING_OPTION("Mp3TagsEncoding","",KviOption_sectFlagFrame),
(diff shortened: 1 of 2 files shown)


Stanzilla:
:+1:

wodim:
:-1:

A better option would be smart UTF-8 with fallback to local locale.

un1versal:
Smart UTF-8 was what was discussed on IRC and ticket. I see though MdiWindowSystemTextEncodingPopupSmartUtf8 is is still valid due to MDi windows being gone?

~~Slightly related also, has anyone noticed that the PM encoding differs form the encoding you set for the server and what is then defined for channels? I see that here, I wonder if its just me.~~

~~For instance set UTF-8 as the encoding for a server specifically, connect, join some channel check encoding (should be UTF-8 as expected) open pm to someone, check encoding (should be UTF-8 but is instead is something else)~~

I found out what that issue was, apparently config files seem like they arent being rewritten after settings change or applied properly. another bug I guess.

un1versal:
@blindsight, is it pheasible to apply smart UTF-8 and push changes to this PR, or would you like to do something else.

3rd alternative is wodim takes care of it. (wink wink, nudge nudge)

staticfox:
Closing since if I end up adding smart utf8 then ill just make a new PR

un1versal:
Im still wondering how this smart UFT-\* works, never really heard of it, and reading the code its makes little sense to me.

ctrlaltca:
When a message is received, the code checks if it's valid UTF8.
If it does, it's then decoded as UTF8; if not, it's decoded using the fallback encoding.

PR #1694, opened 2015-10-27
input: Draw blank buffers on nickname restorations

### LINKED ISSUE #1672: Tabcompletion breaks rendering of input
Let's assume that only me is on channel.
1. type `dar` (with small letter `d`)
2. press Tab, it will expand to DarthGandalf
3. press Tab, it will just move cursor to end of Dar, without removing `thGandalf`
4. move cursor with arrows, cursor doesn't go out of Dar
5. type something (or press backspace), the input box will be fixed.

Using 664c14a on Gentoo with Qt4

staticfox:
This bug was introduced in c385848. Determine if the buffer
being drawn is dirty via spellcheck or nickname restorations.
Fixes #1672

---

Thoughts? Tested and works on my end, it also should not interfere with the spellcheck system.


un1versal:
:+1: Ill test this in a bit, its a gold nugget. :smile:

un1versal:
Tested and confirmed fix, @DarthGandalf this fixes your reported issue.

PR #1695, opened 2015-10-27
widget: Improve the Connect Now button

staticfox:
Make the inactive state color match the active state
Remove the blue glow surrounding the border and text

---
# Before:
## Active, disabled

# After:

### DIFF
--- a/src/modules/options/OptionsWidget_servers.cpp
+++ b/src/modules/options/OptionsWidget_servers.cpp
@@ -1402,10 +1402,13 @@ OptionsWidget_servers::OptionsWidget_servers(QWidget * parent)
 		addWidgetToLayout(m_pConnectCurrent,0,3,3,3);
 		connect(m_pConnectCurrent,SIGNAL(clicked()),this,SLOT(connectCurrentClicked()));
 
-		QPalette pal(QColor(0,0,128));
+		QPalette pal(QColor(0,0,0));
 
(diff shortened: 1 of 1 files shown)


un1versal:
I dont like the inactive enabled states, they are too dissimilar and dont make much sense to me, why we need this 3rd state?

staticfox:
Because it existed before but was only made evident for when you changed the button to green and the inactive state remained blue (unless you intended it to be this way?)

edit: The 3rd state has always exited whether you realized it or not, I updated the button color to match the active state and changed the text to black because if the user is clicked away, you generally want to keep the button more subtle so they can focus on their active window. (like in photography, your eye goes the whitest point so if a user is scoped in to another window that isn't kvirc, it would make sense to darken the text)

un1versal:
what conditions is this 3rd one you mention visible?

Select the network
capture

Select server 
capture1

I cant see any 3rd thing anywhere, what am I missing?

staticfox:
select a server then click on another program like your web browser with the server browser still in view

wodim:
Commit these things yourself, it is not like anybody will complain

un1versal:
I was in the middle of testing this.

un1versal:
OK now I had a chance to review and test some adjustments made in https://github.com/kvirc/KVIrc/commit/decd95c446e2df9e269781c7ea0ee738c2f9007b as follows

active - focus
active - focus

active - unfocus
active - unfocus

inactive
inactive

@blindsighttf2 Do you agree?

PR #1698, opened 2015-10-29
Remove unused functions

staticfox:
Guessing this is left over from mdi, do we really need it? It's never called anywhere.


un1versal:
+1 for cleanup, would be good to get a few blessings before kicking it in.

staticfox:
Seeing as this seems safe and we can always go through git history to re-add this...

un1versal:
Window stacking is definitely gone.

PR #1699, opened 2015-10-29
build: Use git show to set the current source date

staticfox:
I'm pretty sure `KVI_SOURCES_DATE_NUMERIC` is not going to be correct, but it'll be consistent... that and I'm not really sure what it's used for other than internal tracking. Feedback? Yes don't laugh at `CMAKE_KVIRC_SOURCES_DATE_HEX`, I can't think of a better way to jam it as a hex value. :disappointed:


un1versal:
Seems to work well

capture

wodim:
I’m all moist.

un1versal:
+1 for merge

ctrlaltca:
From my kvi_sysbuildinfo.h:

```
#define KVIRC_SOURCES_DATE_GIT "20151029"
#define KVIRC_SOURCES_DATE_GIT_HEX "0x20151029"
```

The hex date ends up being a string; i suppose this can be problematic

staticfox:
Are you sure? I was playing around with it last night and kept getting type errors when trying to print it as a string, other than an int.. e.x.

`QDebug ("blah %s",KVIRC_SOURCES_DATE_GIT_HEX);` would throw a type conversion to int error while `QDebug ("blah %d",KVIRC_SOURCES_DATE_GIT_HEX);` would work just fine.

ctrlaltca:
That variable is used at startup and forces the initial setup if the value is < KVI_SOURCES_DATE_NUMERIC_FORCE_SETUP.
When the setup ends, it's written in the "rc" file (typically ~/.kvirc4.rc).
This is a valid rc file:

```
# KVIrc configuration file
[Main]
LocalKvircDirectory=/Users/fab/.config/KVIrc
SourcesDate=538051606
```

538051606 is the base-10 of 0x20120416.
This is the file that gets saved after your patch:

```
# KVIrc configuration file
[Main]
LocalKvircDirectory=/Users/fab/.config/KVIrc-test3
SourcesDate=0x20151029
```

The date is saved as a string, and it won't be loaded correctly at the next startup:

```
Process 27765 stopped
* thread #1: tid = 0x17a601, 0x000000010023dd74 kvirc`KviApplication::findLocalKvircDirectory(this=0x0000000102201c30) + 900 at KviApplication_setup.cpp:679, queue = 'com.apple.main-thread', stop reason = step over
    frame #0: 0x000000010023dd74 kvirc`KviApplication::findLocalKvircDirectory(this=0x0000000102201c30) + 900 at KviApplication_setup.cpp:679
   676      m_szLocalKvircDir = cfgx.readEntry("LocalKvircDirectory","");
   677  
   678      unsigned int uSourcesDate = cfgx.readUIntEntry("SourcesDate",0);
-> 679      if(uSourcesDate < KVI_SOURCES_DATE_NUMERIC_FORCE_SETUP)
   680          return false; // we force a setup anyway
   681  
   682      // If we have it, ok...done
(lldb) p uSourcesDate
(unsigned int) $5 = 0
```

As a result, on a clean profile, the setup is run again at every startup, because 0 < KVI_SOURCES_DATE_NUMERIC_FORCE_SETUP.

cmake/kvi_sysbuildinfo.h.cmake
```diff
@@ -17,5 +17,7 @@
 #define KVIRC_BUILD_COMPILER "${CMAKE_KVIRC_BUILD_COMPILER}"
 #define KVIRC_BUILD_COMPILER_FLAGS "${CMAKE_KVIRC_BUILD_COMPILER_FLAGS}"
 #define KVIRC_BUILD_REVISION "${CMAKE_KVIRC_BUILD_REVISION}"
+#define KVIRC_SOURCES_DATE_GIT "${CMAKE_KVIRC_SOURCES_DATE}"
+#define KVIRC_SOURCES_DATE_GIT_HEX "${CMAKE_KVIRC_SOURCES_DATE_HEX}"
```
ctrlaltca: To fix the issue, just remove the quotes from this line, so that the #define will get an actual hex value and not a string.

un1versal:
@ctrlaltca is this OK as is now? if so push it in :wink:  :wink:

ctrlaltca:
Fine, merging

PR #1711, opened 2015-11-02
Minor fixes

stkw0:


stkw0:
I can't see this 400+ hits. 
Use ++i (a trivial change) have more impact in non primitive types, although in this case there is not a giant improvement.

un1versal:
You cant? how are you searching, with a stick? 

For e.g. KviQString.cpp has 3 hits
https://github.com/kvirc/KVIrc/blob/master/src/kvilib/core/KviQString.cpp#L506
https://github.com/kvirc/KVIrc/blob/master/src/kvilib/core/KviQString.cpp#L521
https://github.com/kvirc/KVIrc/blob/master/src/kvilib/core/KviQString.cpp#L1264

 and theres other 449 hits in other files which make up for the rest of the entries since you got 1 down here

Stanzilla:
That is a useless mini optimization change that should not even come up in a discussion :p

stkw0:
I only find one use of i++ for one iterator, and I change that. And one use of a invalid pointer (the important thing). 
I agree with @Stanzilla :D

un1versal:
> That is a useless mini optimization change that should not even come up in a discussion :p

the only useless thing I see here is your comment.

so if its so useless either do all or do none.

remove that and leave fix.

DarthGandalf:
@un1versal you're ridiculous.

un1versal:
> @un1versal you're ridiculous.

Im the ones whos ridiculous for not letting shit fly? lol you are welcome

DarthGandalf:
@un1versal do you think your "capitalization" changes in comments in internal scripts (e.g. `gendoc.pl`) are more useful than this?

un1versal:
that commnet merits the same reply as previous comment.

stakewinner00:
I'm agree with "should not even come up in a discussion :p". The utility of that change will depend on the iterations.

un1versal:
\o/

techhelper1:
@un1versal You sir are an idiot.

PR #1714, opened 2015-11-03
Add Favorite Servers!

### LINKED ISSUE #1477: 'Favorites' list in KVIrc
**Reported by mhoes on 07 Feb 2015 09:22:43 UTC**
An editable list of your favorite servers and server/channels. Somewhat like the 'favorites' or 'bookmarks' lists in some browsers. So you can easily navigate to one of your top server/channels. 

Currently there is 'only' the 'most recently used list', which is not necessarily the same as your favorites, and 'Settings -> Configure Servers' which is a list of _all_ 'known' IRC servers out there.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1477

staticfox:
Added favorites, thoughts?

Checklist of stuff I still have to do
- [x] Implement favorite base
- [x] Add  icon
- [x] Add right click button for favoriting

---

Fixes #1477


src/modules/options/OptionsWidget_servers.cpp
```diff
@@ -1817,8 +1842,8 @@ void OptionsWidget_servers::commit()
 
 void OptionsWidget_servers::customContextMenuRequested(const QPoint &pnt)
```
un1versal, line 1845: This is a cosmetic thing I noticed when sending PR,
Do you think its worth adding a indenting fix for this function below so all entries are not all over? 

like line 1849, 1860 1862, 1863 etc.

un1versal:
+1 and I have a Q, will the favorited server be marked on the recent server list also?

You can also drop the merge commit message by doing a interactive rebase.

staticfox:
fixed

un1versal:
OK some suggestions/feedback as discussed on IRC which seem to make sense. I dont mean to be pedantic just somewhat thorough. :smile: 

1) When a Server is already a favorite the label should say **Unfavorite Server** both on **right click menu label** and **toolbox tooltip**
capture

2) When **Only display favorite server** is enabled, and you **Unfavorite a Server** the list should be refreshed automatically and the Unfavorited server removed form list.
capture1

3) **Recent Server Lists** If the recent server was a favorite the entry on these lists should also reflect this IMO. Theres 3 of these instances

**Configure servers dialogs** Recent Servers List
capture

**Configure KVIrc servers dialogs** Recent Servers List
capture

Main KVIrc Toolbar, Recent Server Lists
capture

This next one is somewhat related, I think it would be nice to address it while messing around with these, but could be done later.

4)The toolbox entries would look better and help usability if they would be grouped by similar akin functions and reordered

**Current**
capture

**Mockup of suggestion** 
capture1

capture

staticfox:
1) Done, however it's not really possible to do on the right side since it's a static button.
2) Done
3) Still looking in to.. it's weird. :|
4) I'd like to get others feedback on that first

staticfox:
Alright so after looking this over, I think I'm at a good point for this to go in. re: @un1versal 

> 4)The toolbox entries would look better and help usability if they would be grouped by similar akin functions and reordered

IMO It seems pretty in order as is. You have adding and deleting servers at the top group, copy/paste in the middle, and favoriting at the bottom.

Aside from that, anyone have anything they would like to add?

un1versal:
Not really just that the favorite server should be 3rd entry and only have one separator so like

```
  add network 
  add server 
  favorite server
  remove
  copy
  paste
  -------
  Import servers
```

That feels more natural, then what is atm; however we can do that in separate PR
Lets give this one more spin and merge it.

staticfox:
The above can be debated about in a ticket, but for now, I'm just going to pull this in.

PR #1723, opened 2015-11-06
kvs: Add event for OnMeHostChange

staticfox:
So basically, the idea here is to make it so you can manually script your client to join channels _after_ you have had your host changed which can aid in both privacy and channel flood (i.e. User has joined, services set mode +whatever on User, User has quit irc changing host, user has rejoined, server reset chan modes the user had). I'm honestly quiet terrible with scripting but I can imagine that others may find more reasons to use this. The only network I've found that doesn't support this event is OFTC due to their re-implementation of host changes within their IRCd's fork (it gets sent as a literal server notice).

Thoughts?

### DIFF
--- a/src/kvirc/kvs/event/KviKvsEventTable.cpp
+++ b/src/kvirc/kvs/event/KviKvsEventTable.cpp
@@ -3986,6 +3986,26 @@ KviKvsEvent KviKvsEventManager::m_appEventTable[KVI_KVS_NUM_APP_EVENTS]=
 		"$0 = source nick\n"
 		"$1 = source username\n" \
 		"$2 = source hostname\n" \
-		"$3 = ban mask")
+		"$3 = ban mask"),
 
+	/*
+		@doc: onmehostchange
+		@type:
+			event
+		@title:
+			OnMeHostChange
+		@short:
+			Local user's visible host is changed
+		@parameters:
+			$0 = source server
+			$1 = new host
+		@window:
+			Console
+		@description:
+			Triggered when your visible hostname has been changed. The source could either be services or the IRCd.[br]
+			Note that not all IRC Network support this event, however, major IRC networks including Freenode do.
+	*/
+	EVENT("OnMeHostChange", \
+		"$0 = source server\n" \
+		"$1 = new host\n")
 };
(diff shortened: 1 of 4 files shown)


src/kvirc/kvs/event/KviKvsEventTable.cpp
```diff
@@ -3986,6 +3986,26 @@ KviKvsEvent KviKvsEventManager::m_appEventTable[KVI_KVS_NUM_APP_EVENTS]=
...
+		@title:
+			OnMeHostChange
+		@short:
+			Local user's visible host is changed
+		@parameters:
+			$0 = source server
+			$1 = new host
+		@window:
+			Console
+		@description:
```
un1versal, line 4004: instead... 

```
Triggered when your visible hostname has been changed. The source could either be services or the IRCd.[br]
            Note that not all IRC networks support this event, however, major IRC networks including freenode do.
```

Wouldn't it be nice to have a way to find out which IRCd support this exactly and remove the guessing out of it? ..

.`that doesn't support this event is OFTC`

 Ide work that in there.

staticfox, line 4004: It's impossible, technically all IRCds support this, however, the only one that doesn't is OFTC.

un1versal:
Aside from minor string corrections. some observation for the future.

I love that KVIrc can be extended in all manners, but I dont like that by default many events handlers  dont exist, not even a basic/simple handler —disabled by default even!

+1 and ++1 if someone actually wrote defaults handlers for all of these.

src/kvirc/kvs/event/KviKvsEventTable.cpp
un1versal, line 4004: then saying that makes sense and aids in removing doubt.so something like.

```
Triggered when your visible hostname has been changed. The source could either be services or the IRCd.[br]
            Note that OFTC IRCd doesn't support this event (it gets sent as a literal server notice); however, major IRC networks including freenode do.
```

un1versal, line 4004: also Freenode -> freenode (always) https://github.com/kvirc/KVIrc/commit/405b2f3853cd5023f9d755d8f929fd4411d45278

staticfox:
This should work with all IRCds now.

un1versal:
kick it

PR #1727, opened 2015-11-07
kvsvariant: Fix -Wformat-extra-args in GCC

staticfox:
Output uintptr_t with an appended % before PRIxPTR

---

clang and msvc shouldn't have an issue with this right?


un1versal:
:+1: DarthGandalf and wodim should have a looksee

PR #1728, opened 2015-11-07
Fix no output generated notes

staticfox:
e.g. `KVIrc/src/modules/term/TermWindow.h:0: Note: No relevant classes found. No output generated.`

### DIFF
--- a/src/kvilib/ext/KviDbusAdaptor.h
+++ b/src/kvilib/ext/KviDbusAdaptor.h
@@ -33,7 +33,6 @@
 
 	class KVILIB_API KviDbusAdaptor: public QDBusAbstractAdaptor
 	{
-		Q_OBJECT
 		Q_CLASSINFO("KVIrc D-Bus Interface", "org.kvirc.KVIrc")
 
 	public:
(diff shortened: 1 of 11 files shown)


un1versal:
:+1: DarthGandalf and wodim should have a looksee

un1versal:
@wodim @darthgandalf @Dessa thoughts? Seems to do what it says on tin.

un1versal:
@ctrlaltca thoughts, is this good?

ctrlaltca:
I think this is not good: the `TermWidget` class in `TermWindow.h` needs `Q_OBJECT` since it defines some slots.

The `KVIrc/src/modules/term/TermWindow.h:0: Note: No relevant classes found. No output generated.` message is due to `moc` not finding anything that needs to be parsed in that file.

As you can see that file begins with a `#ifdef COMPILE_KDE4_SUPPORT`; if you get that warning, you're probably compiling without kde4 support, so all that code is not parsed by moc, thus the warning.

This is muting the warning, but it probably break compilation for people compiling with kde4 support. I suggest to actually test it.

DarthGandalf:
It doesn't break _compilation_ (travis builds with and without kde), but it
might (or not) break something in runtime.

2015-11-10 8:06 GMT+00:00 ctrlaltca notifications@github.com:

un1versal:
is there anyone with KDE that can test?

Issue #1614, opened 2015-09-20
texticon typing doesnt trigger some icons unless you use texticon control character.


un1versal:
Reference:

```
static KviTextIconAssocEntry default_associations[] =
{
    { ":)"  , KviIconManager::Smile       },
    { ":*"  , KviIconManager::Kiss        },
    { ":D"  , KviIconManager::BigGrin     },
    { ":("  , KviIconManager::Ugly        },
    { ":/"  , KviIconManager::Angry       },
    { ":O"  , KviIconManager::Surprised2  },
    { ":P"  , KviIconManager::Tongue      },
    { "BP"  , KviIconManager::Tongue2     },
    { ";)"  , KviIconManager::Eye         },
    { ":')" , KviIconManager::TearSmile   },
    { ":'"  , KviIconManager::Cry         },
    { ":S"  , KviIconManager::Afraid      },
    { ":|"  , KviIconManager::Demoralized },
    { ":P'" , KviIconManager::Slurp       },
    { 0     , 0                           }
};
```

Some icons like `":')" , KviIconManager::TearSmile``  typing`:')``` doesnt trigger the associated icon in fact quite a few of these dont do anything.

You need to press ctrl+I to have the texticon control character prefixed in before things like `:')` work.

![capturea](https://cloud.githubusercontent.com/assets/3521959/9980347/f7ad6996-5f8c-11e5-9f15-91b75f69ab98.PNG)
![s](https://cloud.githubusercontent.com/assets/3521959/9980350/573d86ac-5f8d-11e5-8392-33d81d6fbb04.PNG)

This isn't true for all the examples here for instance `:)` works out of the box without any control character.

So why do some need it and why do some dont? Why do we need a control character at all?

---

There is also an issue with texticon control character that you end up with two of them if you press ctrl+i then select an icon with mouse in this pop up menu. Which seems like a separate issue. https://github.com/kvirc/KVIrc/issues/1615

un1versal:
This is Qt version independent it happens with Qt4.x or Qt5.x

un1versal:
related bug report https://github.com/kvirc/KVIrc/issues/1364 from 4.2.0 so this has been broken forever it seems my attempts at fixing failed and didnt make issue better or worst.

pragmaware:
This is because in many cases KVIrc can't reliably determine that a character sequence should be rendered as a text icon. Having all "BP" sequences rendered as a tongue would be wrong.
[...]

PR #1741, opened 2015-11-13
Partially address #1614 - trigger texticons without control character

un1versal:
This Fixes `:'` and `:')` from #1614 entry without control icon

However `:P'`  and others still dont work (without a control character still and Ive no idea how to fix them.

I appreciate not many people like these, but this should work before next big release is out. this is another small step towards making this happen.


DarthGandalf:
This is a completely different icon... Glasses with tongue vs eyes with teeth

un1versal:
> This is a completely different icon... Glasses with tongue vs eyes with teeth

Dont understand whats that have to do with anything,

DarthGandalf:
@un1versal `KviIconManager::Tongue2` is some concrete icon, right? what does it represent, glasses with tongue or eyes with teeth?

un1versal:
capture

According to that its open eyes with tongue (could be glasses) 

If you have another suggestion that **works** for triggering it without control character, Im all ears  all my attempts to make it `8p` or `8P` failed.

un1versal:
The issue here seems to be that any other prefix but colon has been removed intentionally so finding a key combo that still makes sense is the only workaround available and Im out of ideas atm.

DarthGandalf:
`:B` is eyes with teeth.
If I'm typing a smiley `:B` by which I meant eyes with teeth, and suddenly it transforms into image with tongue instead of teeth, that's not very good.

un1versal:
Right, we dont even have that image eyes with teeth or do we?

Ive literally no idea how to place this trigger, all prefix but colon has been removed on purpose according to code comments https://github.com/un1versal/KVIrc/blob/Mary-Poppins/src/kvirc/ui/KviIrcView_getTextLine.cpp#L1062-L1064

so this is broken as a result of that...

DarthGandalf:
> Right, we dont even have that image eyes with teeth or do we?

I don't know, you're the icon guy :-P

I'm fine with having `:B`, but the image should match, and probably it shouldn't be named "tongue2" anymore, if you're changing tongue to teeth.

un1versal:
I suppose that chnaging the icon to workaround the problem in core also would work..., yep I like that solution even if its not the right way to solve the problem, at least I can get this in and the others fixed.

DarthGandalf:
"also"? this is what this PR does from the beginning: changes one icon to another.
I'm just asking this to be more complete (and consistent) change, not a half-way.

un1versal:
well done the changes as discussed, and since https://github.com/kvirc/KVIrc/issues/1614 was closed with such reason merging this is now the full monty...

Ide love to drop support for these texticons but when removing the icons it will be and huge amount of work to reorganize the icons entries and fix the icons assignments in files that use numbers and not icon names...

That or do what I did for when MDI was dropped. leave a bunch of dead entries behind in code with notes and replace them with other icons.

un1versal:
sneaking in the last splash updates... say goodbye to the splash for good say hello to release art

PR #1743, opened 2015-11-13
reguser: Add base for ignoring highlights

### LINKED ISSUE #1605: Light ignore
Requesting a feature that lets you only ignore highlights from a certain user, instead of all messages. The message would still show in the channel/query window but not trigger the highlighting system, so task bar flash on Windows, the notifier popup etc.

staticfox:
Closes #1605 

---

Don't merge this, I'm PRing this so I can motivate myself to finish this.


un1versal:
+1

PR #1748, opened 2015-11-23
follow up #1743 add ignore highlight to internal aliases.kvs

un1versal:
This adds the (missing from #1743) recently added highlight ignore to aliases.kvc

todo: possible add a Check all box to RegisteredUserEntryDialog.cpp

@staticfox 

I think these changes types are really bad for existing users who upgrade KVIrc, unless you restore your default scripts, the popups and aliases will never be updated to reflect these changes.

### DIFF
--- a/data/defscript/aliases.kvs
+++ b/data/defscript/aliases.kvs
@@ -360,7 +360,7 @@ alias(ignore)
 	if(!$0)
 	{
 		echo $tr("Usage:","defscript")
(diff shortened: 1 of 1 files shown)


staticfox:
> I think these changes types are really bad for existing users who upgrade KVIrc, unless you restore your default scripts, the popups and aliases will never be updated to reflect these changes.

That's going to be a whole nother' game any way you slice it. I'm sure we can come to a solution on it though.

PR #1750, opened 2015-11-28
fix show Qt in about dialog

un1versal:
One commit fixes a typo in return function. which I think prevents it from working. 

The other commit removes KviOsInfo::qtVersion();

```
infoString += "<br>";
infoString += __tr2qs_ctx("Qt Version","about");
infoString += ": ";
infoString += KviOsInfo::qtVersion();
```

Which seems to blow up KVIrc instantly when you try to open the about dialog.

capture

just need to increase the default window height to remove the scrollbar and show all the info

### DIFF
--- a/src/kvilib/ext/KviOsInfo.cpp
+++ b/src/kvilib/ext/KviOsInfo.cpp
@@ -718,6 +718,6 @@ namespace KviOsInfo
 
 	QString qtVersion()
 	{
-		return QString(qVersion());
+		return QString(qtVersion());
 	}
 }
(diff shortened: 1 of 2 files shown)


src/kvilib/ext/KviOsInfo.cpp
```diff
@@ -718,6 +718,6 @@ namespace KviOsInfo
 
 	QString qtVersion()
 	{
-		return QString(qVersion());
+		return QString(qtVersion());
```
DarthGandalf, line 721: How does this NOT blow up kvirc?
This is infinite recursion which finishes with stack overflow.

DarthGandalf:
> Which seems to blow up KVIrc instantly

Yes, that's what stack overflow does after the first commit of this PR.

un1versal:
So you saying return QString(qVersion()); is correct and not a typo?

By the way with this PR works 100% no issues.

wodim:
what the fuck dude

PR #1755, opened 2015-12-07
Added 64-bit ack option to DCC receive.

ghost:
Issue #1447 

It's not specified in the original issue, but I'm going to assume that the user is attempting to download from iroffer or an mirc client.  This is a common issue that I've seen among non-mIRC users.  From my testing, it appears to be an issue where iroffer chose to follow the mIRC method and use 64-bit acks for files larger than 4GiB.  However, I can't find any client besides mIRC that actually sends 64-bit acks by default.

It culminates in the fact that iroffer assumes that it will receive a 64-bit ack (it does have 32-bit ack detection, but it's poor at best) and when iroffer gets to the end of the file, the acks don't match up and it makes some decision that the transfer is done and closes the socket before all of the data has been sent.

The best thing that can happen is for mIRC/iroffer to use 32-bit acks, but that's not going to happen.

The workaround that I've used in the past is to just not send acks for files larger than 4GiB, but this breaks the protocol and is not the best solution.

I think the best option that KVIrc can do is to create an option in the DCC file transfer options to send 64-bit acks on files larger than 4GiB.

This patch adds an option in "DCC > File Transfer > Advanced > Bug Compatibility" that will enable the usage of 64-bit acks when the file is larger than 4GiB.

I haven't tested this extensively, but it's worked so far.  Unfortunately, when dealing with files larger than 4GiB on a slow connection can cause testing to be rather slow and this issue doesn't happen 100% of the time, since in my experience it will randomly succeed sometimes with the standard 32-bit ack protocol.

Note: This partially reverts the bTolerateErrors workaround introduced in dc6c4f5 since it doesn't seem to do what was intended.

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -340,7 +340,8 @@ KviBoolOption g_boolOptionsTable[KVI_NUM_BOOL_OPTIONS]=
 	BOOL_OPTION("ShowUserFlagForChannelsInWindowList",true,KviOption_sectFlagWindowList | KviOption_resetUpdateGui),
 	BOOL_OPTION("EnableCustomCursorWidth",false,KviOption_sectFlagGui | KviOption_resetUpdateGui),
 	BOOL_OPTION("ShowFavoriteServersOnly",false,KviOption_sectFlagFrame),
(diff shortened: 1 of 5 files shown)


un1versal:
pending some tests (need to phone a friend) this should be OK to go as it fixes #1447

staticfox:
Seems fine to me, I'll merge it in a bit if no one else objects.

PR #1761, opened 2015-12-11
Alignment fixes to masks dialogs + add a more descriptive title to servers dialog entry

un1versal:
@staticfox I managed to fix one dilaog message alignment

lex7uc3d

The other instances of alignment I couldn't figure it out ~~see https://github.com/kvirc/KVIrc/compare/master...un1versal:My-Favorite-Martian~~ only the ones on this PR worked, the rest either didnt work or made things look worst.

As you asked this is how it should be for the others.

capture1

So these below ones I couldn't figure out ~~in https://github.com/kvirc/KVIrc/compare/master...un1versal:My-Favorite-Martian~~

> **Edit**: thx to Thereign for help with encryption menu text align.... looks great. (cut off text still not looked at)

capture

And another with two issues

capture

Any hints would be good I cant figure any of it out.

### DIFF
--- a/src/kvirc/ui/KviCryptController.cpp
+++ b/src/kvirc/ui/KviCryptController.cpp
@@ -129,8 +129,10 @@
 		pLayout->addWidget(m_pDecryptHexKeyCheck,8,3);
 
 		m_pOkButton = new QPushButton(__tr2qs("OK"),this);
-		pLayout->addWidget(m_pOkButton,9,0,1,4);
+		pLayout->addWidget(m_pOkButton,9,2,1,2);
+		m_pOkButton->setMinimumWidth(80);
 		connect(m_pOkButton,SIGNAL(clicked()),this,SLOT(okClicked()));
+		m_pOkButton->setIcon(*(g_pIconManager->getSmallIcon(KviIconManager::Accept)));
 
 		pLayout->setRowStretch(3,1);
 		pLayout->setColumnStretch(2,1);
--- a/src/kvirc/ui/KviCryptController.cpp
+++ b/src/kvirc/ui/KviCryptController.cpp
@@ -72,11 +72,15 @@
 
 		QGridLayout * pLayout = new QGridLayout(this);
 
+		QHBoxLayout * pTitleLayout = new QHBoxLayout(this);
+
 		QLabel * pLabel = new QLabel(this);
 		pLabel->setPixmap(*(g_pIconManager->getSmallIcon(KviIconManager::Locked)));
-		pLayout->addWidget(pLabel,0,0);
+		pTitleLayout->addWidget(pLabel,0);
 		pLabel = new QLabel(__tr2qs("Cryptography/text transformation"),this);
-		pLayout->addWidget(pLabel,0,1,1,3);
+		pTitleLayout->addWidget(pLabel,1);
+
+		pLayout->addLayout(pTitleLayout,0,0);
 
 		QFrame * pFrame = new QFrame(this);
 		pFrame->setFrameStyle(QFrame::HLine | QFrame::Sunken);
--- a/src/kvirc/ui/KviCryptController.cpp
+++ b/src/kvirc/ui/KviCryptController.cpp
@@ -94,11 +94,11 @@
 		connect(m_pListBox,SIGNAL(currentItemChanged(QListWidgetItem *, QListWidgetItem *)),this,SLOT(engineHighlighted(QListWidgetItem *, QListWidgetItem *)));
 		pLayout->addWidget(m_pListBox,3,0,6,1);
 
-		m_pDescriptionLabel = new QLabel(this);
-		m_pDescriptionLabel->setWordWrap(true);
-		m_pDescriptionLabel->setFrameStyle(QFrame::Sunken | QFrame::StyledPanel);
-		m_pDescriptionLabel->setAlignment(Qt::AlignTop | Qt::AlignLeft);
-		pLayout->addWidget(m_pDescriptionLabel,3,1,1,3);
+		m_pDescriptionText = new QTextEdit(this);
+		m_pDescriptionText->setReadOnly(true);
+		m_pDescriptionText->setFrameStyle(QFrame::Sunken | QFrame::StyledPanel);
+		m_pDescriptionText->setAlignment(Qt::AlignTop | Qt::AlignLeft);
+		pLayout->addWidget(m_pDescriptionText,3,1,1,3);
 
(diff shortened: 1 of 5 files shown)


ghost:
This worked for me on the Crypto title layout.

> ```
>            QGridLayout * pLayout = new QGridLayout(this);
> 
>            QHBoxLayout * pTitleLayout = new QHBoxLayout(this);
> 
>            QLabel * pLabel = new QLabel(this);
>            pLabel->setPixmap(*(g_pIconManager->getSmallIcon(KviIconManager::Locked)));
>            pTitleLayout->addWidget(pLabel,0);
>            pLabel = new QLabel(__tr2qs("Cryptography/text transformation"),this);
>            pTitleLayout->addWidget(pLabel,1);
> 
>            pLayout->addLayout(pTitleLayout,0,0);
> ```

This comes from the QGridLayout and that the "Cryptography/text transformation" label starts on column 2.  Just make row 1 of the QGridLayout contain a single QHBoxLayout instead where the icon takes up the minimum width.

You might be able to extend it to your other problem titles.  I didn't check.

ghost:
For the description box being cut off, this probably only happens when your screen is tiny or you've shrunk the area down a lot.  If you want to add a scroll bar, you can try changing the `QLabel` to something like `QTextEdit` or `QPlainTextEdit` and use the `setReadOnly(true)`.

un1versal:
Thank you, that was very helpful indeed. I always get flustered with Qt dialogs, :smile: , Ill test these things and see how it goes.

un1versal:
Thats another out of the way... 5 to go

just realised that the other 4 need a sunken line also... that or remove it from where it is now.

capture

ghost:
Same `QHBoxLayout` trick works for the others as well.

Look in file `src/kvirc/ui/KviMaskEditor.cpp` starting on line 166.

> ```
>    QHBoxLayout * pTitleLayout = new QHBoxLayout(this);
> 
>    QLabel * l = new QLabel("",this);
>    l->setPixmap(*(g_pIconManager->getSmallIcon(m_eIcon)));
>    pTitleLayout->addWidget(l,0);
> 
>    l = new QLabel(szDescription,this);
>    pTitleLayout->addWidget(l,0);
> 
>    g->addLayout(pTitleLayout,0,0);
> ```

The question is... is the KVIrc way to use the `KviTalHBox`?

un1versal:
> The question is... is the KVIrc way to use the KviTalHBox?

It does. thx for the next bit its just missing the sunken line but looks good.

un1versal:
@TheReign nearly... just need to move the line up a bit

capture

wodim:
In the crypto dialog you could change the OK button so it's aligned to the right and it does not take the whole width for itself.

un1versal:
lets see what this gives https://github.com/kvirc/KVIrc/compare/master...un1versal:My-Favorite-Martian and Im giving up for tonight

un1versal:
capture

nearly but the sunken line wont line to top... and button needs adjusting somemore...

Jobs for tomorrow, unless someone is feeling generous.

thx thereign, a little more and this is cracked. nitenite...

ghost:
@un1versal 

You're adding the sunken line to row 1 (0-indexed), which is correct, but you're not increasing the rows for the subsequent rows that existed before.

`g->addWidget(pFrame,rowNum,colNum,rowSpan,colSpan);`

This is a `QGridLayout` so you're explicitly specifying the row number rather than just appending them.

So the `pLayout->addWidget(m_pOkButton,9,0,1,Qt::AlignRight);` in your commit, the `Qt::AlignRight` isn't what you want there.  You probably want either `-1` (which should force it to consume all of the columns to the right) or to set the alignment with another method.

Off the top of my head, `addWidget(m_pOkButton,9,1,1,-1);` might do what you want (and you might need to increase the row-span of `m_pListBox` from `6` to `7`).

un1versal:
I simply do not understand Qt at all, so here is trial and error with mostly error.

un1versal:
Ok added a few more fixes for encryption menu

> In the crypto dialog you could change the OK button so it's aligned to the right and it does not take the whole width for itself.

Done, I tried one cell but it looked very odd this way it fits with other dialogs.
capture

The test alignment is also fixed, but the sunken line even though code is there cant get it to work properly, so I need to figure that out still

capture1

ghost:
@un1versal 

See https://github.com/TheReign/KVIrc/commit/9125e91ab70992697b78e4d09ec38f43ee8ab8c8 for how I added the HR to the mask editor.  See how I adjusted the row index for everything below the HR by adding one to it?  You need to re-index the rows below when you add a new row.  That'll fix why your HR isn't showing up, because you're adding the filter row to the same row you already added the HR to.

Also, alternative idea on the OK button placement: https://github.com/TheReign/KVIrc/commit/4adcbd5ed1e41ad3d2059333a38cbca2fc0c01b4

ghost:
Here's my suggestion for QTextEdit which allows for a scroll bar when the text overflows.

https://github.com/TheReign/KVIrc/commit/132eb888ad79f03db50d6d760b257aeceaad08ac

un1versal:
I tested something like TheReign@4adcbd5 and the button was a tad small when comparing to other dialogs, but thx anyway.

I can git cherry-pick your other two commits into this PR, but Ill do it tomorrow I think and after a quick test.

Great job btw @TheReign , I think would be nice and about time to send you an invite to the dev team @pragmaware thoughts? Hint hint nudge nudge.

un1versal:
@TheReign actually just cherry-picked them now and removed my failed attempt at adding HR line, I did think about adjusting the rest, but a combination of laziness and trying to make some sense of Qt align properties by trial and error is the only way Ill ever (spaghetti monster forbid)  learn.

un1versal:
This is how it looks now :+1: 

capture

There seems to be more padding in between the top line and the HR line in some dialogs then others but thats a minor cosmetic only noticed when you have different dialogs up together like in screenshot
tbh not any reason to sneeze at this as its done the job.

thanks will be merging in the morning if no one has any objections.

un1versal:
Here are some last minutes thoughts.
- Apply/Close button on Mode editor is also spans across whole length and that varies a lot..,
  This dialog is really weird, since when server is disconnected if you bring up the dialog and even the HR line is cut off...
- there's a lot a variance of size between the add/remove buttons width (they never just look consistent
- there's a horizontal scrollbar even thought no content is displayed (when program main window isnt full size)
- Mode editor has an extra frame around outside dialog the other dialogs dont have this.

So Im merging this as is, the rest can be solved separately as generally this is MUCH better now.

PR #1764, opened 2015-12-11
Travis: Add build with 'debuild'

AlexanderP:


DarthGandalf:
Why so many pull requests?
You can just push to the same one (even with `--force`) until it works

AlexanderP:
Хорошо. Надо определится почему скрипт не получает переменные.

DarthGandalf:
Он не может найти kvirc_4.3.2~git-5611-g3825cfe.orig.tar.gz

AlexanderP:
нет он не может получить версию. 

git describe --dirty                                            /media/sdb5/build-kvirc/1123/kvirc/KVIrc/dist/ :(
fatal: No names found, cannot describe anything.
git branch | grep "*" | sed 's/* //g'                         /media/sdb5/build-kvirc/1123/kvirc/KVIrc/dist/ :(
(detached from FETCH_HEAD)

Придется сделать костыль.

DarthGandalf:
> fatal: No names found, cannot describe anything.

А, потому что это shallow clone, история загружена не целиком, а последний таг был давным-давно.

Можно травису сказать, чтоб грузил больше

DarthGandalf:
или можно `git describe --always`, тогда он будет даже без тагов

AlexanderP:
Про гит я потом подумал, вот такой 'костыль' решит проблему: 

test -z ${dat} && dat="git-5748-gf9deb85" \
branch="travis_debian" \
commit="f9deb8576daed3bfe83afd0b78257125bde41ecb" \
datct="1449857731"

AlexanderP:
Стоит его включать? 

если так git describe --always, то надо правила переписать и думать как получить количество комитов

DarthGandalf:
Кого стоит включать?

"f9deb8576daed3bfe83afd0b78257125bde41ecb" можно из переменных окружения самого трависа взять

AlexanderP:
test -z ${dat} ... в скрипт сборки, но я по другому добавил.

DarthGandalf:
> we going to need phonon4qt5-backend-gstreamer as a runtime lib

I disagree. `phonon4qt5-backend-vlc` works fine too. KVIrc shouldn't force users to use gstreamer.

AlexanderP:
При сборке ставился "phonon4qt5-backend-null" может так и надо оставить.

DarthGandalf:
I think it should just depend on `libphonon4qt5-4`, as it links to that library, and not rely on any particular backend.

AlexanderP:
I agree.

un1versal:
> I think it should just depend on libphonon4qt5-4, as it links to that library, and not rely on any particular backend.

Thats fine if you want to drop `phonon4qt5-backend-gstreamer` but if you drop that you need to make sure a backend is installed and without that is not, so phonon audio wont work here, only Qt5Multimedia audio will work.

un1versal:
If you look at http://packages.ubuntu.com/trusty/libphonon4qt5-dev you see it, the installed ``phonon4qt5-backend-null` will not work at all and thers no proper backends installed at all.

DarthGandalf:
@un1versal end user doesn't even need `libphonon4qt5-dev` to be installed. That's the difference between binary distros and distros where user compiles everything from source.

If ubuntu chose to make `libphonon4qt5-dev` to depend on `phonon4qt5-backend-null`, so be it.
But, as I said, KVIrc shouldn't force users to use any particular backend of phonon, be it `null` or not.

If user wants to switch audio from default qt to phonon, probably they know what they are doing.

un1versal:
but then you know phonon wont work out of the box and that defeats the purpose for any end user because there;s no working phonon anything by default, assuming they dont have it setup for some other app (and we must assume they dont)

Heres the rub, as is phonon will be auto-detected by kvirc and will not work, so a user has to manually change default kvirc settings to get audio working... This at least should be fixed in kvirc if no backend for phonon is installed but qt is and does work out of the box.. 

We cant have broken design by default.

> @un1versal end user doesn't even need libphonon4qt5-dev to be installed. That's the difference between binary distros and distros where user compiles everything from source.

Well are you saying then that this is only an issue on users that compile kvirc? Or am I making a false assumption?

DarthGandalf:
> but then you know phonon wont work out of the box and that defeats the purpose for any end user because there;s no working phonon anything by default, assuming they dont have it setup for some other app (and we must assume they dont)

If user makes a choice to select Phonon instead of the default audio system used by KVIrc, then we must assume that the user really wants Phonon instead of the default audio system used by KVIrc. If that's the case, we must assume that the user wants it for some reason, and not just clicking random buttons in Settings. It would be stupid to protect users against _that_.

> Heres the rub, as is phonon will be auto-detected by kvirc and will not work

Does KVIrc ever select Phonon by default? I thought it doesn't.

> Well are you saying then that this is only an issue on users that compile kvirc?

No.
I don't know why `libphonon4qt5-dev` depends on `phonon4qt5-backend-null`, but existence of null backend doesn't change anything. Existence of a non-null backend is important, and also such backend should be selected in Phonon's settings. Examples of a non-null backends of Phonon are xine, GStreamer, DirectShow, VLC, QuickTime.

I don't know why you're forcing GStreamer to users.

> If you look at http://packages.ubuntu.com/trusty/libphonon4qt5-dev you see it, the installed `phonon4qt5-backend-null will not work at all and thers no proper backends installed at all.

Backends installed and selected on the build machine (if Phonon is even configured there at all) has nothing to do with backends installed and selected on user's machine.

---

This is issue only when user doesn't know what they are doing. It was exactly the case for you, as you assumed Phonon 5 should work even if you never used Phonon 5 before and don't even use KDE 5. Phonon only makes sense for KDE users after all.

un1versal:
> Does KVIrc ever select Phonon by default? I thought it doesn't.

As is, yes, last time I checked.

> This is issue only when user doesn't know what they are doing. It was exactly the case for you, as you assumed Phonon 5 should work even if you never used Phonon 5 before and don't even use KDE 5. Phonon only makes sense for KDE users after all.

Not true, if KVIrc prefers phonon over qt (if both are compiled in) and it wont have ANY backend as is thus wont work.

But if you think Im wrong thats fine too..

> Examples of a non-null backends of Phonon are xine, GStreamer, DirectShow, VLC, QuickTime.

Yes sure, great, but assuming you install kvirc on a system that has none setup, phonon will not work and is preferred over qt.
Even if you install VLC you still wont have any out of the box working audio with phonon. thats been my experience and tested.

> I don't know why you're forcing GStreamer to users.

Ive been trying to explain to you the reason, but clearly Im failing to do that, and since Im apparently unable to do explain, then I give up, If people start opening reports, then somone will have to deal with this issue then. It wont be me as Ive no idea how to fix it anyway.

DarthGandalf:
In that case the default selection logic should be fixed, not forcing users to use GStreamer.

un1versal:
Sure I fully agree with that, but I fix it the only way I can.

PR #1765, opened 2015-12-12
lithandler: Exec query sounds on query ACTIONs

staticfox:
Right now if someone queries you with an action (i.e. /me) there will be no sound. This fixes it, thoughts?


un1versal:
OK, just a thought, what about adding this to General Options -> Sounds?

staticfox:
As what? The way CTCP ACTION is treated is no different than a private message really.

un1versal:
No idea, it was my stumble out of bed thought for this.

un1versal:
Alright for me.

Issue #1650, opened 2015-10-02
Remove Qt 4 support


wodim:
What is stopping us from doing this? Maybe supporting people who still do not have Qt 5 on Linux? Do those still exist?

I'm asking because of changes such as df6c711 or ed9d80e.

DarthGandalf:
Yes, they exist.
E.g. ubuntu 12.04 lts.

DarthGandalf:
for ubuntu, ppa can probably be used though.

wodim:
@DarthGandalf In that case we have to bundle Qt 5 in our PPA or...? There is no Qt 5 at all for 12.04?

DarthGandalf:
Either https://launchpad.net/~ubuntu-sdk-team or https://launchpad.net/~beineri should work... (from http://askubuntu.com/questions/279421/how-can-i-install-qt-5-x-on-12-04-lts)
Well, travis has 12.04, you can try to use qt5 from one of those ppas.

wodim:
I could understand if we needed to support Qt 4 for for example one more year, but I just checked and 12.04 will continue to have support until almost 2018. So maybe a PPA is a reasonable solution.

ctrlaltca:
Things that i found to be problematic in dropping qt4, and sticking with qt5 only:
- since qt 5.3.0, they dropped support for processors without sse2 instruction set (pentium < 4 and AMD Athlon)
- osx < 10.7 is no more supported (so no more carbon, only cocoa)
- windows xp is still "supported", but not even tested anymore, so expect some breakage on new versions.

More info here: http://doc.qt.io/QtSupportedPlatforms/index.html

DarthGandalf:
Travis can use 14.04 now, http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/
14.04 has qt 5.2

un1versal:
32-bit builds wont run with Qt5 (on some machines) Some complaint in IRC

So I think for now we should provide at least random git builds with Qt4 for the die hard, cant afford newer machine, I want my windows 4.1 to work kind of guys.

Dessa:
so has it been confirmed that qt4 builds work or is that some random guess based on that older builds worked? because that has little relevance, due to code changes.

i'd rather get this done sooner then later so i'll try to get a debug build of qt5 and friends done over the weekend, maybe a kvirc debug build would be sufficient but who knows.

un1versal:
> ctrlaltca wrote:
[...]

PR #1768, opened 2015-12-13
Remove deprecated and unused QMatrix references

ghost:
The `QMatrix` class has been deprecated in Qt5.  I also couldn't find anywhere where `m_pMatrix` (which isn't even declared as a pointer) used in the current code.

### DIFF
--- a/src/modules/objects/KvsObject_painter.h
+++ b/src/modules/objects/KvsObject_painter.h
@@ -43,7 +43,6 @@ protected:
 	QPrinter * m_pPrinter;
 	bool bDonotdeleteinternalqpainter;
 	QPainter * m_pPainter;
-	QMatrix  m_pMatrix;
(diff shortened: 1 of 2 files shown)


un1versal:
Im not familiar with these things, but Qt4 is still supported, what happens with this in those instances?

Somewhat related #1650

DarthGandalf:
If this compiles, should be fine.
Windows failed build doesn't seem related to this.

ghost:
@un1versal It's just a straight deletion of 2 unused lines, so even if we still support Qt4, it won't change anything.

un1versal:
:+1: for cleaning up

un1versal:
failure is related to https://github.com/kvirc/KVIrc/issues/1769 this has been blessed so its in. thx.

PR #1777, opened 2015-12-19
Fix KviCryptController layout double add run-time warning

ghost:
You get this message when attempting to open the encryption engine settings window.

> QLayout: Attempting to add QLayout "" to KviCryptController "", which already has a layout


un1versal:
How are you getting this message?

Never seen it myself.

ghost:
Open KVIrc from terminal, then join a channel, and then open the crypt settings.

PR #1780, opened 2015-12-23
Fixes saving of missing icons in theme

ghost:
Fixes part of #1643 

Issue is that in `src/kvirc/kernel/KviIconManager.cpp` several lines have "No longer used, do not remove entry, just reuse entry for new icons." in the `g_szIconNames`.  This causes there to be null icons being attempted to be saved.

Checking for null icons before saving them fixes it.

It's been proposed that not saving icons is a better method.  I don't know what's best, so this may not fix the bigger problem, only remedies a bug.

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -1159,7 +1159,7 @@ namespace KviTheme
 				szPixPath.append(g_pIconManager->getSmallIconName(j));
 				szPixPath.append(".png");
 
-				if(!pix->save(szPixPath,"PNG",90))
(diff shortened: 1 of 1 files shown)


un1versal:
With MDI -> SDI change many icons were dropped, removing the entries will break the icon mapping by entries as some parts of KVIrc use icon numbers instead of icon names to call a specific icon.

ghost:
Obviously you can't remove the entries, I was merely explaining why there were null icons attempting to be saved.

un1versal:
It never occurred to me what you explained that this was the reason. Im assuming you tested this? If so, thx and will merge in a bit.

PR #1783, opened 2015-12-24
Fixes compressed log files being saved improperly on stopLogging()

ghost:
Issue: When `stopLogging()` was called, compressed log files weren't being flushed and saved properly. If the log had never reached it's first flush, then no log file would be created. If it had gone through at least one flush, then the last flush wouldn't be saved to the log file.

The issue is caused by using the wrong filename to write to the compressed log in the `stopLogging()` function. It copied the contents of the *.log.gz.tmp into a buffer, and then wrote the compressed file back to that same *.log.gz.tmp, then proceeded to delete that *.log.gz.tmp. Thus, leaving you with whatever had been previously flushed by the timer event.


un1versal:
Need to find out later time if that ticket is related.

PR #1800, opened 2015-12-31
Fixes the display of the key sequence for actions in menus.

ghost:
Instead of having &lt;label&gt;&lt;tab&gt;&lt;shortcut&gt; it is now properly right aligned like other applications.

Please test with KDE and OSX to see how it looks.


un1versal:
build failure not related...

+1 looks good here

PR #1801, opened 2015-12-31
Properly check for PerlLibs for ld -lperl.

### LINKED ISSUE #1473: Lack of Perl not reported by CMake
**Reported by wodim on 02 Jan 2015 14:45:08 UTC**
If your installation lacks Perl dev headers, CMake does not tell you and you get an error in the compiling phase instead

```
/usr/bin/ld: cannot find -lperl
collect2: error: ld returned 1 exit status
```

Reproduced on Ubuntu 14.04

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1473

ghost:
Fixes #1473 

`find_package(Perl)` just looks to see if perl is installed, not if the perl libs are installed for linking.

Requesting someone to verify my pull request.


un1versal:
lol we both adding label at same time

ghost:
PerlLibs fails to do what I wanted.

On an Ubuntu 15.10 live disk that doesn't have libperl-dev installed.

`package_find(PerlLibs)` finds `PERL_LIBRARY=/usr/lib/i386-linux-gnu/libperl.so.5.20` and sets `PERLLIBS_FOUND`.

Unfortunately, `ld -lperl --verbose` can't find it.  It seems to require the symlink of libperl.so -> libperl.so.5.20 to successfully find -lperl.

I might be missing something obvious here.

ghost:
This works properly for PythonLibs.

`/usr/lib/i386-linux-gnu/libpython2.7.so.1` exists, but `package_find(PythonLibs)` doesn't find it for what I think is because there is no symlink of `libpython2.7.so` pointing at it.

I'm so confused.  Is cmake broken?

ghost:
`find_library(PERLLIB_FOUND NAMES perl)` works, but it seems like a dirty hack to me.

No idea how portable that is either, but it "works" on ubuntu.

DarthGandalf:
Why do you need `ld -lperl`? CMake's PerlLibs links directly to full path of libperl.so.5.20 instead of relying on `-L` and `-l` flags.

ghost:
Try building without having the libperl.so and you'll get a failed make that complains of not being able to find -lperl In "Linking CXX shared module libkviperlcore.so".

ghost:
```
perl -MExtUtils::Embed -e ldopts
-Wl,-E  -fstack-protector -L/usr/local/lib  -L/usr/lib/x86_64-linux-gnu/perl/5.20/CORE -lperl -ldl -lm -lpthread -lc -lcrypt
```

and

```
# Get links flags
execute_process(
    COMMAND ${PERL_EXECUTABLE} -MExtUtils::Embed -e ldopts
    OUTPUT_VARIABLE PERL_LINK_FLAGS
)
...
target_link_libraries(${kvi_module_name} ${LIBS} "${PERL_LINK_FLAGS}")
```

As far as I know, -lperl requires libperl.so and will not work with just libperl.so.5.20

If you've got a way to use PERL_LIBRARY from FindPerlLibs, I'm all ears.

ghost:
@DarthGandalf thoughts on this change?  It works fine on a machine with the libperl.so symlink.  I'm running another test on one without that symlink, but I think it should work.

Edit: worked on my box without libperl.so from libperl-dev.

Tested working on linux

DarthGandalf:
This fails on OS X and Windows. Check Travis and AppVeyor output...

ghost:
I guess I'm missing it.  I don't see any fails on appveyor/travis.

DarthGandalf:
It's not "failure", it's just "Perl: no" in cmake summary

DarthGandalf:
FindPerlLibs can't work on OSX at all :( http://public.kitware.com/Bug/view.php?id=12544

ghost:
Looks like FindPerlLibs on cmake is terribly out of date and isn't an option.

PR #1805, opened 2016-01-01
Add a sunken border to the tree window list and to the user list.

wodim:


src/kvirc/ui/KviTreeWindowList.cpp
```diff
@@ -547,12 +557,15 @@ void KviTreeWindowListItemDelegate::paint(QPainter * p, const QStyleOptionViewIt
 			{
 				// The QWindowsXP style does not honor our colors. It uses the system ones instead.
 				// We can't accept it.
-				p->fillRect(option.rect,col);
+				QStyleOptionViewItemV4 opt4(option);
```
DarthGandalf: One local variable shadowing another local variable with the same name.
This is confusing to read.

DarthGandalf: Ah, sorry, that's in different scope

wodim: I don't even know what the `opt4` var name referred to originally but for reasons of consistency I decided to stick with it.

un1versal:
Screenshots before and after pls?

wodim:
Before -> look at your current KVIrc.
Now ->



The same sunken border the topic and ircview widgets have is now applied to the window list and the user list also.

un1versal:
My current kvirc you cant tell as I dont use themes light enough  to be able to discern, and other people may not as well.

But sure +1 no brainer.

Stanzilla:
Great change, now it finally looks the same across all panels

PR #1813, opened 2016-01-04
Add SSL urls (ircs/ircs6) in server options dialog.

ghost:
Adds dynamic display of ircs:// and ircs6:// when "Use SSL" checkbox is toggled.

Fixes typo in irc6s warning.


wodim:
Does this work at all?

ghost:
Why wouldn't it?  It worked previously for irc:// and irc6:// .  I just extended it for SSL and had it actually display the "s".

wodim:
Eh, nothing.

un1versal:
/me does Frankenstein voice and shouts —"It liiveess!"

un1versal:
there only one ommission I can tell here., while on terminal type `kvirc --help`

one of the outpust is as follows:

```
  [ircurl]     : URL in the following form:
                 irc[6]://<server>[:<port>][/<channel>[?<pass>]]
```

should it be:

`irc[s][6]://`

ghost:
The only problem with that change is that the command line option [ircurl] doesn't support enabling ssl with ircs:// designation. (Not that I could get to work anyways)

That would require separate commit.  This was just for a display fix.

Issue #1816, opened 2016-01-04
Search tools: proposal and request for resorting flyout options


un1versal:
Regarding the menu flyout over the output/split view called Search tools... this ticket has two parts...

Part 1:)

![capture1](https://cloud.githubusercontent.com/assets/3521959/12093764/b52cdf98-b2fc-11e5-93f7-a394e6327582.PNG)

It makes little sense to me why this is called search tools when it has more functionality that is not related to searching than to it.

I suppose the simplest, is to rename it to something that probably more suitable, Perhaps:
- Output tools
- Search and output tools
- Output view tools

So please either say which name you prefer or suggest some alternative to rename that to..

---

Part 2) 

But then thats a short term solution to me, because I dont think many people realize you have all those other options available there, so it occurs to me that we could move such options like zoom/font and background options either to:

A) to the top right, as first option before encoding tools

**or**

B) to the bottom again as first option.

Then the current button would only toggle search directly and the other options would have their own dialog that could also handle the split view scenario.

I dont have any mocks at this time but I can make something up eventually.

In the meanwhile some thoughts and feedback on both part 1 and part 2 would be nice.

ghost:
2) Neither, since it applies separately to each of the split windows and wouldn't make sense if it was only a single button.  But, definitely not B as it's not an input option list.

un1versal:
well the applies separately could be handled in dialog... but it would require some more work...

So rename option then... now to what? Im ok with any of the 3 but prefer the 3rd option.

un1versal:
see https://github.com/kvirc/KVIrc/pull/1817 that should be something towards this...

Issue #1814, opened 2016-01-04
[ueserlist] selecting away users nick, doesnt follow color selction rules.


un1versal:
This is probably intended by design, but just looks wrong, 

These are my normal `#232323` / selected `#c8c8c8` colors
![capture1](https://cloud.githubusercontent.com/assets/3521959/12084830/5c255dae-b2b0-11e5-85f0-100fb4b22630.PNG)

The color `#e6e6e6` for the nick away is:
![capture5](https://cloud.githubusercontent.com/assets/3521959/12084863/c1fefcfc-b2b0-11e5-95ba-38bd232d2066.PNG)

This is what happens when I select a normal user:
![capture](https://cloud.githubusercontent.com/assets/3521959/12084878/fc760d44-b2b0-11e5-82da-d84459cf4165.PNG)

And this happens when selecting an away user:
![capture0](https://cloud.githubusercontent.com/assets/3521959/12084882/0c0374ae-b2b1-11e5-9433-85175b667b10.PNG)

That to me looks odd and feels wrong.

TheReign:
Try this: 3bcc660

ghost:
Right now it blends foreground of "away" and whatever the foreground is supposed to be normally.

Would you prefer it to not blend at all and just use foreground of "selected" when you select a user?

un1versal:
I would prefer that there's always some significant contrast, enough to moot this issue, else if thats not possible, use same foreground...

One consideration would be, what if you have a user selected that was not away and in meanwhile went away, and how could you visually tell there's a status change, and is that important?

atm I have set prefs>irc>channel/advanced `[x] Enable keep away list updated` but thats not default, so theres also that.

Make sense?

un1versal:
It just occurred to me... what about adding a  settable color to 
![capture1](https://cloud.githubusercontent.com/assets/3521959/12084830/5c255dae-b2b0-11e5-85f0-100fb4b22630.PNG)

so we can configure the away when selected color?

Not even an idea how to word that, so its probably a bad idea :D

ghost:
I guess that's an option, since I can't figure out how to de-select the user in the list so you'd lose all ability to determine by color whether the selected user was away or not.

But, with the added setting, you'd lose all other color indicators as well, since it currently is a "dimming" of sorts for away status.

Also, "Enable keep away list updated." is enabled by default.  I find that dangerous when you sit in multiple channels with 2k+ users in them, since it lags out the connection periodically, but that's another issue.

[...]

PR #1817, opened 2016-01-04
strings: address #1814 and #1816

un1versal:
Misleading strings be gone....

~~I should really add a tootip to 8413d2e tbh it needs one as this isnt something that behaves as others do.~~


un1versal:
objections? feedback? thoughts?

un1versal:
will merge later on if theres nothing in the way of acitvity here, unless someone beats me to it first.

Issue #1706, opened 2015-11-01
Default server list should be updated.


ghost:
I've just installed KVirc on ArchLinux and I was surprised to see that a lot of dead URLs are in the default server list.
There's even a reference to irc0.dreamcast.com.

It should probably be easy to run a nmap script to see which servers are still up and therefore worth keeping.

staticfox:
Yeah I've been meaning to do a full network list run through and kill off
dead ones and add new ones.

un1versal:
marespiaut

+1, and if you found it in your heart to make a Pull Request fixing these things it would even be better

ghost:
@un1versal I thought about it. I'll do it when I have time.
----- EDIT -----
Nevermind, here's a quick way to start seeing if the servers are still up

```
wget 'https://raw.githubusercontent.com/kvirc/KVIrc/master/data/config/serverdb.kvc'
grep Hostname serverdb.kvc | cut -d = -f 2 > url_list.txt
nmap -sn $(cat url_list.txt)
```

Here is a list of 141 dead URLs

```
Failed to resolve "aphrodite.portalx.org".
Failed to resolve "apollo.portalx.org".
Failed to resolve "ares.portalx.org".
Failed to resolve "arlington.va.us.skyyenet.org".
Failed to resolve "as.dal.net".
Failed to resolve "athena.portalx.org".
Failed to resolve "au.alphachat.net".
Failed to resolve "auckland.nz.galaxynet.org".
Failed to resolve "au.ipv6.alphachat.net".
Failed to resolve "aussie.sydney.oz.org".
Failed to resolve "b-network.azzurra.org".
Failed to resolve "brown.freenode.net".
Failed to resolve "calkins.freenode.net".
Failed to resolve "calvino.freenode.net".
Failed to resolve "chariot.adelaide.oz.org".
Failed to resolve "chat14.x-echo.com".
Failed to resolve "chat.ap.freenode.net".
Failed to resolve "cheapnet.azzurra.org".
Failed to resolve "crichton.freenode.net".
Failed to resolve "de.lightspike.net".
Failed to resolve "demeter.portalx.org".
Failed to resolve "efnet.demon.co.uk".
Failed to resolve "efnet.ipv6.xs4all.nl".
Failed to resolve "efnet.teleglobe.net".
Failed to resolve "einstein.freenode.net".
Failed to resolve "eos.portalx.org".
Failed to resolve "eu.alphachat.net".
[...]

PR #1827, opened 2016-01-07
serverdb.kvc cleanup

### LINKED ISSUE #1819: Can't connect to Quakenet using default server (it doesn't exist)
kvirc always tries to connect to irc.quakenet.eu.org, which doesn't exist; is this a typo for irc.eu.quakenet.org? (That would be consistent with the other global servers.) Have we been DDOSing the European Union this whole time? ;)

Steps:
1. Attempt to connect to the Quakenet network without specifying a server
2. kvirc attempts to connect to the false EU server and fails with a DNS error

(Not sure why it always attempts to connect to the EU server - it's not the first server in the list, and I'm not in the EU...)

I can work around this bug by deleting the EU server from the server list, or by manually connecting to a different server (not to the network as a whole). Oddly setting a different server as a favorite has no effect; do favorites do something else?

```
Executable Information

KVIrc 4.9.1 'Aria'

Runtime Info:
System name: Windows Release : 6.2.9200
System version: Windows 8 (x64) (Build 9200)
Architecture: Intel64 Family 6 Model 58 Stepping 9
Qt version: 5.5.1
Qt theme: windowsvista

Build Info:
Build date: 2016-01-05 10:20:26 UTC
Sources date: 20160102
Revision number: git-6040-gb7fa661
System name: Windows-6.3.9600
CPU name: AMD64
Build command: C:/Program Files (x86)/CMake/bin/cmake.exe
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=C:/projects/kvirc/build/release
   Threads=Win32
Compiler name: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/amd64/cl.exe
Compiler flags: /DWIN32 /D_WINDOWS /W3 /GR /EHsc
Qt version: 5.5.1
```

ghost:
Partial work on #1706 
Fixes #1819 

I only removed NXDOMAIN and 127.0.0.1 pools.

This should be checked/reviewed before merged. So many line changes :(


Stanzilla:
Sync with https://github.com/hexchat/hexchat/blob/master/src/common/servlist.c maybe?

ghost:
Not the goal of this PR, but would probably be beneficial if added to #1706

un1versal:
Good stuff ;) thx

PR #1832, opened 2016-01-09
ThemeEngineVersion: bump minimum compatibility

un1versal:
The point here is to warn users that the older themes 1.0.0 they are installing will likely be broken...

1) texticons.kvc inside themes has been dropped from inside themes and is now read/written to from user config directory. (which is where its always been created at first run just never used properly).

2) Also themes that have internal icons, because of the many icons that have been dropped, renamed its this will make such themes look also not so great. **things like messages and toolbar icons and throughout menus etc... will mostly be a mix between current core and whatever it was the old themes have which will look bad.

If neither of these are an issue to the theme being installed this serves as a reminder.

The message that pops up should warn user about this....

@pragmaware thoughts?

@ everyone else feedback?


un1versal:
Ive tested this by installing themes that have a 1.0.0 engine, which pretty much everything else..

No matter if installed manually or via theme manager, theres no warning or message anywhere :disappointed: 

Anyone know why? Its not immediately obvious atm

un1versal:
As is no warning whatsoever installing via theme manager from web.

capture

In theme manager the potential incopatible theme wont show up...

capture

un1versal:
tried adding https://github.com/kvirc/KVIrc/commit/c774150b23bff24e8ba3db90f77057b124a2b0d2 but the message still doesnt pop up...

sigh.

un1versal:
well.... some progress....

it turns out that neither of these if's are processed
https://github.com/un1versal/KVIrc/blob/331dd8da4fcd818b3748bbb14bce9e60a5b2d6bc/src/kvirc/kernel/KviTheme.cpp#L128-L140 neither installing old themes on new kvirc nor the original message when installing newer themes on older kvirc....

So thats either broken or broken or maybe even broken.

What I managed to get working is https://github.com/kvirc/KVIrc/commit/e16e35499db17ac9809c59c533e9b44f3873ace3

which displays https://github.com/kvirc/KVIrc/blob/e16e35499db17ac9809c59c533e9b44f3873ace3/src/modules/theme/ThemeFunctions.cpp#L218

or 

https://github.com/kvirc/KVIrc/blob/e16e35499db17ac9809c59c533e9b44f3873ace3/src/modules/theme/ThemeFunctions.cpp#L234

capture

those messages work on both counts when installing old themes on new kvirc and newer themes on older kvirc.

So still need help here figuring out first part

un1versal:
lets try to ask @ctrlaltca if he has a clew.

un1versal:
Theme handling on installing / loading any greater or lower engine version themes is 100% broken. Closing this until  master fixes this..

https://github.com/kvirc/KVIrc/issues/1848

Issue #1835, opened 2016-01-11
Set avatars via UI +  https URL wont work


un1versal:
In https://github.com/kvirc/KVIrc/issues/1604 it was closed with https://github.com/kvirc/KVIrc/commit/bd9a509651b34fd52c6ef3968715b58c70f29872 which apparently implemented HTTPS support for avatars...

This works fine if you set avatar manually like

`/avatar.unset https://cdn.pbrd.co/images/HfJaCVE.png`

however when actually doing it via UI...
![capture](https://cloud.githubusercontent.com/assets/3521959/12233568/b76edda2-b85f-11e5-8229-331c757af538.PNG)

![capture](https://cloud.githubusercontent.com/assets/3521959/12233573/c4ec47bc-b85f-11e5-84be-2e22aae78bc2.PNG)

un1versal:
Fixed by https://github.com/kvirc/KVIrc/commit/9cdc263fba19671d311e02c8374c2df106623980

thx Andrio/staticfix

![capture](https://cloud.githubusercontent.com/assets/3521959/12269173/00f838ae-b948-11e5-960f-107c24393cca.PNG)

PR #1836, opened 2016-01-12
fix recognizing https:// URL in CTCP AVATAR replies

un1versal:
thx @Andrio for pointing it out.

Someone that understands this should review and merge.

Personally I cant see a difference and https://github.com/kvirc/KVIrc/issues/1835 is still buggy even with this.


wodim:
Change `KviQString::equalCIN("https://",szRemoteFile,7)` for `KviQString::equalCIN("https://",szRemoteFile,8)`

un1versal:
> Change KviQString::equalCIN("https://",szRemoteFile,7) for KviQString::equalCIN("https://",szRemoteFile,8)

if thats true then https://github.com/kvirc/KVIrc/blob/master/src/modules/avatar/libkviavatar.cpp#L243 also needs to be 8
added in https://github.com/kvirc/KVIrc/commit/1245875326a5be9600696e417095da21f49d40ce please confirm.

wodim:
ya

un1versal:
well this is it... if you want it merge it...

Personally I cant see a difference and #1835 is still same bug even with this.

PR #1837, opened 2016-01-12
rename applet labels to reflect what they affect

un1versal:
This needs more feedback as I cant make heads or tails of what other options affect.

Its silly to be honest to name things in a manner where the user would have to guess by trial and error.

tbh no idea what other options do and code is too cryptic to me.

capture


un1versal:
anyone?

ctrlaltca:
The other color options are not used. It seems that also the "font" option is not used.

un1versal:
> The other color options are unused. It seems that also the "font" option is unused.

is that a bug or intentional? I can try to remove those in a separate pr then

ctrlaltca:
The options date back to c75b9f00eaeee4b8503ba662b3d7ccc9e1914096 (original import from csv to svn), i suppose they were used by some other toolbar applet that's not included anymore.

un1versal:
superceeded by https://github.com/kvirc/KVIrc/pull/1844

PR #1844, opened 2016-01-15
Superseeds #1837 rename toolbar applet option and remove unused

un1versal:
@ctrlaltca please have a looksee.

I didnt remove the defines because it seems odd comment them out or remove them due to total count idk what effects that would have. Some other defines have been commented out in past, idk if thats right.

am compiling for a quick runtime test.


un1versal:
capture

Seems to work just fine...

Ill do internal themes off of that and add changes here and merge if no one has any thing to say about this.

un1versal:
@ctrlaltca should I remove the defines adjust total count and renumber so they match, or will commenting them out be enough?

previously some are just commented out, idk but that seem not right.

un1versal:
this is as ready and as finished as I can get it. please review and kick in

un1versal:
anyone has anything at all to say? if not in by the morning.

wodim:
I have to say that we shouldn't change option names unless it's completely necessary.

un1versal:
They made no sense as they were before according to the current logic and now they at least in the neighborhood of sane, im still having thoughts about the labels names though.

Issue #1850, opened 2016-01-17
We don't support custom input methods on Qt 4


wodim:
After 39a9e33 we do not support custom input methods, so it needs to be reimplemented or...

un1versal:
...drop Qt4

wodim:
I need a "Make America Great Again" cap that says "Drop Qt4" instead.

un1versal:
![capture](https://cloud.githubusercontent.com/assets/3521959/12379755/1b3faaa8-bd58-11e5-8f3f-e703ceb306ad.PNG)

un1versal:
![capture](https://cloud.githubusercontent.com/assets/3521959/12379767/942a0ac6-bd58-11e5-9994-ed61cd8229d9.PNG)

wodim:
That could be our temporary splash screen if only we still had a splash screen :sob:

un1versal:
Better yet

![capture](https://cloud.githubusercontent.com/assets/3521959/12379796/589f394e-bd59-11e5-90c1-986378b27532.PNG)

alternative design
![capture](https://cloud.githubusercontent.com/assets/3521959/12379833/5c23851a-bd5a-11e5-86f4-b854b145c6cd.PNG)

un1versal:
Qt4 gone this is now invalid/fixed

PR #1859, opened 2016-01-31
Drop support of Qt4 from CMakeLists

DarthGandalf:
All C++ #ifdefs are still there... See #1650

Also switch Travis to Qt5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,15 +27,13 @@ install:
     - CMAKE_ARGS=""
     - |
         if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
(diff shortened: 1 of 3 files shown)


wodim:
Travis on OS X fails.

DarthGandalf:
@wodim Not anymore.
Though before merging this I'd prefer everyone (e.g. @ctrlaltca) to object if they want.

un1versal:
from #1850 
capture

I guess the INSTALL docs will need to reflect this also. 

my +1 for this here for what is worth.

ctrlaltca:
I have no objections :+1:

Issue #1860, opened 2016-02-01
Tooltip newline breaks make no sense


un1versal:
This was a problem before https://github.com/kvirc/KVIrc/pull/1861

some examples of whats wrong (but there are many others with or without ) screenshots were taken with https://github.com/kvirc/KVIrc/pull/1861 compiled in but even without these issues are present.
- Some tooltips main text content breaks / adds a newline where none was added.
- Some tooltips width is insufficient and breaks sentence midflow or add one word to a newline and then breaks
- Too many different types/formatting or styling for tootips would be nice if only two variations existed instead several more.
- Breaks and newlines arent respected fully.
- Some tooltips headers are centered, yet they dont center (i.e. status bar applets)

https://github.com/kvirc/KVIrc/search?l=cpp&q=%3Cb%3E%251%3C%2Fb%3E+%28%3Cu%3E%3Cfont+color%3D\%22blue\%22%3E%3Cnowrap%3E&utf8=%E2%9C%93

On that code they all wrap even though it says not to.

Would be nice if wrapping/nowrapping and breaks/newlines were respected and the tooltip set some autowidth within reason but this doesnt happen or is even respected. for e.g.

![capture](https://cloud.githubusercontent.com/assets/3521959/12709252/d2f81c08-c8a1-11e5-8126-5bb397cc471f.PNG)
![capture1](https://cloud.githubusercontent.com/assets/3521959/12709251/d2f6d5fa-c8a1-11e5-9bc8-26cfd05d6f1f.PNG)

Its also not font size or type normally I use monospace size 10 (screenshots above) the screenshot below are with size 8 and still wrapping on those examples
![capture](https://cloud.githubusercontent.com/assets/3521959/12750326/6602d756-c9ae-11e5-941e-ee251c070f6f.PNG)

![capture](https://cloud.githubusercontent.com/assets/3521959/12750366/a4a59890-c9ae-11e5-85a4-5d5b2ef31bbb.PNG)

User tooltips also suffer from this issue
![capture](https://cloud.githubusercontent.com/assets/3521959/13595673/385a9074-e505-11e5-8d85-842efbb77d68.PNG)

In some rare cases and randomly it seems, the ```Using server: some.server@some.place.somewhere also wraps for no apparent and valid reason.

screenshot from sdd (also a kvirc user) at some network @thereign is familiar with
Edit: [image removed due to content]

un1versal:
lets poke @TheReign for thoughts.

[...]

PR #1861, opened 2016-02-01
Remove centered text formatting from all tooltips and some other dialog entries

### LINKED ISSUE #1847: Declaring war on most large blocks of centered text
But not all centered text, for instance, tooltip header text and other centered header text is perfectly acceptable, Also acceptable in the Honor and Glory dialog type.

But large blocks of text like tooltips and similar, should be left aligned. **This enhances readability** since we are trained to read RTL since an early age (unless youre LTR but these things dont apply there and if they did they should be handled automatically not forced to center, also cant see many programs who use centered text for these things.

Aside from the fact not every tooltip is actually centered some are not.

In next few days I will be testing local chnages with view of proposing final changes via PR.

thoughts?

un1versal:
closes #1847 

**Known issues (not caused by this PR)** Ticket https://github.com/kvirc/KVIrc/issues/1860 

I cant solve #1860, (wasted nearly a week looking now) no idea (and dont care anymore) none of the logic makes sense theres **N** ways these things are being done all of them more confusing than the next. :D 

As for this here it is: review it and see if I can get any feedback on it as some parts I was not 100% sure.


un1versal:
Can you have a looksee and check https://github.com/un1versal/KVIrc/commit/deaced93caf9e75c56296d5980cc01d33abd35f2 is OK

wodim:
looks gud

un1versal:
> looks gud

great ;) thx... merge it then?

PR #1880, opened 2016-02-19
KviTopicWidget/KviChannelWindowcontext/KviConsoleWindow: fix & improve some tootltips

un1versal:
Improved consistency of START_TABLE_BOLD_ROW colors, our tooltips of this kind have many many many styles, this goes in one way to reduce the variety and start making these more consistent.

Also The topic tooltip header title text needed to be centered to match the other similar type styles + appended colons to Set by: and Set on: labels 
@wodim if you have any tips on how to strip the topic colors from `txt += KviHtmlGenerator::convertToHtml(KviQString::toHtmlEscaped(m_szTopic));` Line 365 that would be nice, because when topics are colored they mostly become unreadable on this tooltip in some cases.

~~Also libkvitrayicon added a marginally better title for that context menu @wodim, would be nice if you also had a tip on how to add some indenting to this header..~~ merged in master still need solution


un1versal:
~~I just added c24397fd to address the Channel Mode Topic style treatment, however it is a bit broken atm, I need to figure out what went wrong with c24397fd~~
**edit:** 
Fixed, turns out this needed html + also noticed an already existing <html> was not closed in 33356ee

un1versal:
This works lovely now with exception of the silly topic tooltip whne the topic has colors some colors are unreadable against the tooltip yellow BG(when the OS style applies a yellow tooltip BG)

@wodim merging this then its muc better now though much work around these things is required to make it more consistent.

Issue #1885, opened 2016-02-24
Windows: Opening KVIrc via IRC URL or shortcut opens a new instance every time


kidmar:
I use Firefox browser to open irc:// links.
When I click on those links, kvirc open itself and connects to the server correctly.
The second time I click on the same link, a new copy of kvirc opens and connects to the same server again (or to another server if it was a different link).

In previous versions (<= 4.3.1) I only had 1 copy of kvirc open. Clicking the second time on the same link, didn't do anything (as I was already connected).
I think this is a regression (and can't find a way to workaroud this)

```
KVIrc 4.9.1 'Aria'

Infomazioni Runtime:
System name: Windows Release : 6.2.9200
System version: Windows 8 Professional (x64) (Build 9200)
Architettura: Intel64 Family 6 Model 60 Stepping 3
Qt version: 5.5.1
Qt theme: windowsvista

Informazioni Compilazione:
Build date: 2016-02-23 10:07:03 UTC
Sources date: 20160102
Revision number: git-6246-g8ff3bff
System name: Windows-6.3.9600
CPU name: AMD64
Build command: C:/Program Files (x86)/CMake/bin/cmake.exe
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=C:/projects/kvirc/build/release
   Threads=Win32
Compiler name: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/amd64/cl.exe
Compiler flags: /DWIN32 /D_WINDOWS /W3 /GR /EHsc
Qt version: 5.5.1
```

un1versal:
I had a glace at appveyor config and the cmake configuration at https://github.com/kvirc/KVIrc/blob/master/.appveyor.yml#L118

I can see anything in there that would imply allow multiple sessions or instances, are you sure your links aren't passing `-f` flag to the KVIrc executable?

un1versal:
mmm odd, Just fired up a Windows VM and If you clcik the KVIrc shortcut each time you click a new instance/session opens, and Im 100% sure that thers no `-f` flags anywhere there.

My Linux instance is self compiled KVIRc version and without `-f` it will not open a new instance and throws a warning message. Marking this as a Windows bug for now.

kidmar:
No -f flag in links.
Also on my machine running kvirc.exe (or a shortcut) multiple times opens multiple instances.

un1versal:
Yea Ive confirmed it also, now we wait...

[...]

PR #1891, opened 2016-03-02
Get rid on undesired message: if its not the same on all OS's then  d…

un1versal:
…oesnt belong

Fix compiling on Windows also (hopefully)

### DIFF
--- a/src/kvirc/kernel/KviIpcSentinel.cpp
+++ b/src/kvirc/kernel/KviIpcSentinel.cpp
@@ -217,7 +217,6 @@
 	{
 		setObjectName("kvirc4_ipc_sentinel");
 #if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
-		setWindowTitle(__tr2qs("IPC Sentinel - KVIrc"));
 		setWindowFlags(Qt::FramelessWindowHint);
 #else
 	#if defined(COMPILE_X11_SUPPORT) && defined(COMPILE_QX11INFO_SUPPORT)


un1versal:
I wonder if https://github.com/kvirc/KVIrc/blob/0d4df593543eb3ae1997ceec47028da851f7c9b9/src/kvirc/kernel/KviIpcSentinel.cpp#L39-L47 is looking for the window title to be `kvirc4_ipc_sentinel`

ctrlaltca:
Exactly. This commit https://github.com/un1versal/KVIrc/commit/f41c1f2868ee7d318b29e4c076db45a8b0346b3d#diff-bb6b2b0631283944011aa0ef4ca449d8L220 actually broke the IPC sentinel.
IPC stands for inter-process communication. When you run KVIrc, it will first check if a KVIrc instance is already running and "pass the command" to that instance instead of opening a new one.
I suppose this is related to #1885

ctrlaltca:
Also, that string would be better placed in a constant so that it's easy to detect that it should not be changed.

Dessa:
so i suppose that specific string isn't user facing at all and should be put back the way it was before?

ctrlaltca:
I've opened #1892, let's wait for appveyor to finish the build and test its output.

un1versal:
#1885 was already broken before https://github.com/un1versal/KVIrc/commit/f41c1f2868ee7d318b29e4c076db45a8b0346b3d#diff-bb6b2b0631283944011aa0ef4ca449d8L220

thx in any case https://github.com/kvirc/KVIrc/pull/1891 at least fixed compiling

PR #1894, opened 2016-03-03
Hopefully fix WinXp support on AppVeyor builds

ctrlaltca:
They install fine but then when running KVIrc i receive the infamous "... is not a valid Win32 application" (in the following example, localized in italian for your fetish pleasure)
schermata 2016-03-03 alle 08 46 28

Looking at the appveyor script the problem is that Visual Studio 2013 requires us to specify the use of the "vs120_xp" toolkit when building in order to build winxp compatible executables.

CMake has a specific command-line option for that ([-T <toolset-name>](https://cmake.org/cmake/help/v3.2/manual/cmake.1.html)), but it's obviously broken/unusable as everything else a bit advanced in cmake.

This PR takes a more direct approach, forcing a linker flag when MSVC is used.

Yeah i know that Xp is old, it's not secure, every time i boot it up a kitten dies, etc.. but if in order to support it we just need such a simple change, why not?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1017,6 +1017,8 @@ endif()
 if(MSVC)
 	add_definitions(-D_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES -DNOMINMAX)
 	add_definitions(-DMSVC)
+	# Support from Windows XP
+	SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /SUBSYSTEM:WINDOWS,5.01")
 endif()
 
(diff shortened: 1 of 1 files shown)


Dessa:
for 5.0 this gets a +1 for me, when this is done, it should probably be dropped.

ctrlaltca:
I can confirm that the test build available at https://ci.appveyor.com/api/buildjobs/1m5bfecen03sk59m/artifacts/KVIrc-4.9.1-dev-git-6266-g309a312-x86.exe runs fine on windows xp.

schermata 2016-03-03 alle 11 37 36

un1versal:
look at that, supporting the unsupported :grinning: congrats

Now for the other umpteen bugs, joking , seems like their growing on trees these days. :cry:

Issue #1900, opened 2016-03-09
Unnecessary check for Qt Svg


Arfrever:
Qt Svg is not used anywhere in KVIrc.
Nothing includes Qt Svg headers.
Nothing is linked against `libQt5Svg.so`.

I suggest to comment out this check in `CMakeLists.txt`:

```
option(WANT_QTSVG "Whether to compile Qt SVG support" ON)
...
# Qt SVG support
if(WANT_QTSVG)
        find_package(Qt5Svg)
        if(Qt5Svg_FOUND)
                list(APPEND qt5_kvirc_modules Svg)
                include_directories(${Qt5Svg_INCLUDE_DIRS})
                set(COMPILE_SVG_SUPPORT 1)
        endif()
endif()
```

And to delete references to `libqt5svg5-dev` in `dist/debian/control` and `doc/INSTALL.txt`.

un1versal:
not sure if this is valid, you remove this and KVIRc wont be able to work with svg, something we may want in future or currently, pretty sure I saw a fork using svg instead of pngs with no mods except referencing svg instead of png (or Im imagining things) besides with this we can have any svg graphics on scripts and other things.

wodim:
@un1versal are you drunk?

Arfrever:
If usage of Qt Svg is introduced in the future, then check for Qt Svg would be simply uncommented.

Checks for e.g. Qt D-Bus and Qt WebKit define COMPILE_DBUS_SUPPORT and COMPILE_WEBKIT_SUPPORT.
COMPILE_DBUS_SUPPORT and COMPILE_WEBKIT_SUPPORT are used in some files (e.g. `#ifdef COMPILE_DBUS_SUPPORT`).
But COMPILE_SVG_SUPPORT is not used anywhere.

Presence of check for unused library in `CMakeLists.txt` might deceive these packagers of KVirc who forget to grep sources of KVIrc (not me).

un1versal:
> @un1versal are you drunk?

Nope...  https://github.com/logangorence/kvircx

wodim:
@un1versal what is that?

un1versal:
> @un1versal what is that?

Are **you** drunk?

ctrlaltca:
The cmake rule has been added by @Noldor73 in https://github.com/kvirc/KVIrc/commit/c0c7dfa7c9ef7952d264c4e6eac171c99fcfd226 as an optional component.
[...]

PR #1904, opened 2016-03-12
cmakelists and files: cleanup references to Qt SVG

un1versal:
## This is not for merging yet

Reference: https://github.com/kvirc/KVIrc/issues/1900

Need to investigate this further and gain feedback from @HelLViS69 and @pragmaware who worked on this according to changelog entries.

please **Anyone** in meanwhile let me know if anything else needs removing or cleaning up regarding this rather do a complete job not leave half of it undone.


un1versal:
@DarthGandalf clang failure, is that something to worry about?

DarthGandalf:
Well, it segfaulted for some reason... After retry it didn't.

staticfox:
See ticket

PR #1907, opened 2016-03-19
Added $dcc.currentSpeed KVS function

ghost:


wodim:
Does this work?

wodim:
My only problem with this is the name, I think "current" would be more appropriate than "instant", but other than that it looks, well, the way it's meant to look. So if it's good merge it

Stanzilla:
Yeah `instant` is the wrong word here, imo. +1 for `current`

ghost:
Instant is what the internal function is, which is what I mimicked. I don't see an issue with it, but I guess if people want it changed, then ok. I like keeping things consistent. But, 2 votes isn't enough for me.

staticfox:
> Instant is what the internal function is

Users will only ever see the external/api level name of things so if it means breaking consistency against internal naming conventions to be more logical then it is best to do it that way.

un1versal:
+1

Ive no strong opinion on naming if it was a UI option maybe there be room for thought,, If it works, go ;) 

WB btw

PR #1909, opened 2016-03-21
Display host changes with CHGHOST

staticfox:
@DarthGandalf are you OK with this? IMO It's better to just leave it as `KVI_OUT_NICK` since it implies a user-side identity change and adding _another_ kvi option would be eh for something so similar. As an aside, perhaps we could make a KVS event now for `OnUserHostChange` or something. _shrugs_


DarthGandalf:
If this works with "split view", fine with me.
I don't like repeating the same logic too much for query and channel
cases...
Anyway, thanks.

2016-03-21 4:27 GMT+00:00 BlindSight notifications@github.com:

staticfox:
I'm not a fan of the duplicate logic either, but it seems most of the file does it.. may make a better solution at some point to eliminate that.

PR #1910, opened 2016-03-24
help: Move all IRC support channels to freenode

staticfox:
So realistically, I went through all the support channels we offer through locales and they weren't even registered nor did they have any users (aside #kvirc-fr on freenode due to channels not expiring). I feel like we should just point all "help" resources via IRC to #KVIrc on freenode. No I'm not picking network favorites or anything, just, sending European users to an empty void seems totally unfair. Plus, #KVIrc has plenty of users and devs that speak languages other than English.

Amendment: Just pulled kvirc-fr.info since it's also down. Perhaps we could make the new website able to be automagically translate in to a desired locale? Then we can get rid of www.kvirc.ru as well and maintain only a single website. @RCMainak


src/kvirc/kernel/KviInternalCommand.cpp
```diff
@@ -89,10 +89,7 @@ static const char * internalCommandTable[KVI_NUM_INTERNAL_COMMANDS] =
 	"quit",
 	"openurl http://www.kvirc.ru",
 	JOIN_CHANNEL_ON_NETWORK("#KVIrc","freenode"),
-	JOIN_CHANNEL_ON_NETWORK("#KVIrc.net","azzurra"),
-	"openurl http://kvirc-fr.info",
-	JOIN_CHANNEL_ON_NETWORK("#KVIrc-fr","freenode"),
-	JOIN_CHANNEL_ON_NETWORK("#KVIrc-fr","europnet")
+	"openurl http://kvirc-fr.info"
```
Stanzilla: that site is dead, can just remove it as well

staticfox: oh I never bothered to check lol

src/kvirc/ui/KviMenuBar.cpp
```diff
@@ -147,21 +147,6 @@ void KviMenuBar::setupHelpPopup(QMenu *pop)
 	help->addSeparator();
 	pAction = help->addAction(*(g_pIconManager->getSmallIcon(KviIconManager::Channel)),__tr2qs("Join KVIrc International Channel on freenode"));
```
Stanzilla: I guess we can remove the `International` then, might replace it with `Support`? and also drop `on Freenode` since there is only one now?

staticfox: I'm not particularly wild in specifically labeling the channel as `support` since then it seems single use, such as a help channel where the only purpose is to ask questions and leave. As for the `International` part, I'll leave that up to @un1versal since he's better with words than me ^^. And I'd still like to keep `on freenode` to specify to the user that it's a trusted network and not a random 1 user hacknet in zimbabwe.

un1versal:
We only do help in English afaik and if we dont we should.

I also agree that we have too many channels probably ghost channels going back to some past age where whoever thought the current spaghetti was a good idea. Its not.

+1

As for naming neither support nor International belong there.

Issue #1915, opened 2016-04-01
[Feature Request] [Windows] Option to hide the menu bar


Stanzilla:
New option under settings => toolbar to hide the menu bar and only show it when `ALT` is pressed.

wodim:
New option under settings => toolbar to get loads of people in our channel asking where their menu bars went.

Stanzilla:
Make it (the option) off by default, ofc!

wodim:
I will implement it, probably. But people will disable the menu bar and then won't know how to make it return, that is a fact.

un1versal:
In Linux this is easy, install appmenu-qt5 and watch KVIrcs menu disappear.

un1versal:
Just so comments arent lost, the menubar should always be visible by default with possibility to hide/unhide by pressing Alt.

I also think just in case we should have some sort of visual indicator on statusbar whem menubar is hidden via this method..
This suggestion is because of what I written above, by installing appmenu-qt5 it breaks KVIrcs menubar (hides it permanently) and then we wouldnt know if its because of this or because of that when ppl complain. @1977David already complained to me about the menu being gone on his build.
Coincidentally I only found out about appmenu-qt5 tonight idk yet if thats the reason for his issue but it does cause it here when tested.

un1versal:
> I also think just in case we should have some sort of visual indicator on statusbar whem menubar is hidden via this method..
> This suggestion is because of what I written above, by installing appmenu-qt5 it breaks KVIrcs menubar (hides it permanently) and then we wouldnt know if its because of this or because of that when ppl complain. @1977David already complained to me about the menu being gone on his build.
> Coincidentally I only found out about appmenu-qt5 tonight idk yet if thats the reason for his issue but it does cause it here when tested.

The suggestion above is due to https://github.com/kvirc/KVIrc/issues/1917 (4 user reports in total for that issue over IRC)

1977David:
Hello...

After removing appmenu-qt5 the menu did appear again. So all fine now. 

See here:

https://github.com/kvirc/KVIrc/issues/1917

un1versal:
fixed by #1919

PR #1916, opened 2016-04-01
DO NOT MERGE Show menubar only when the Alt key is hit

wodim:
This is a tentative implementation for #1915. There's no option at all (it's always on), I just want to know if it works well for others and on OSes other than Linux.

### DIFF
--- a/src/kvirc/ui/KviMainWindow.cpp
+++ b/src/kvirc/ui/KviMainWindow.cpp
@@ -918,12 +918,11 @@ void KviMainWindow::contextMenuEvent(QContextMenuEvent *)
 	// do nothing! avoids builtin popup from qmainwindow
 }
 
-void KviMainWindow::keyPressEvent(QKeyEvent *event)
+void KviMainWindow::keyReleaseEvent(QKeyEvent *event)
 {
-	switch(event->key())
+	if(event->key() == Qt::Key_Alt)
 	{
-		case Qt::Key_Alt:
-			g_pMainWindow->showMenubar();
+			g_pMainWindow->toggleMenubar();
 	}
 
 	QMainWindow::keyPressEvent(event);
@@ -958,10 +957,15 @@ void KviMainWindow::updatePseudoTransparency()
 #endif
 }
 
-void KviMainWindow::showMenubar()
+void KviMainWindow::toggleMenubar()
 {
-	m_pMenuBar->show();
-	m_pMenuBar->setFocus();
+	if(m_pMenuBar->hasFocus())
+	{
+		m_pMenuBar->hide();
+	} else {
+		m_pMenuBar->show();
+		m_pMenuBar->setFocus();
+	}
 }
 
 void KviMainWindow::hideMenubar()
--- a/src/kvirc/ui/KviMainWindow.cpp
+++ b/src/kvirc/ui/KviMainWindow.cpp
@@ -133,10 +133,15 @@ KviMainWindow::KviMainWindow()
     m_pAccellerators = new KviPointerList<QShortcut>;
 	m_pMenuBar   = new KviMenuBar(this,"main_menu_bar");
 	setMenuWidget(m_pMenuBar);
+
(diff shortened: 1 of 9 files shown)


un1versal:
I dont mind the idea of hiding/unhiding menu bar by pressing Alt, but as is hidden by default is a -1 from me.

FYI works only to unhide it pressing Alt, if you press Alt again to hide it it does sod all. Clicking the main window bar hides it.

src/kvirc/ui/KviMainWindow.cpp
```diff
@@ -133,6 +133,7 @@ KviMainWindow::KviMainWindow()
     m_pAccellerators = new KviPointerList<QShortcut>;
 	m_pMenuBar   = new KviMenuBar(this,"main_menu_bar");
 	setMenuWidget(m_pMenuBar);
+	m_pMenuBar->hide();
```
un1versal: Should not be hidden by default imo,

wodim: Read the pull request.

src/kvirc/ui/KviMainWindow.cpp
```diff
@@ -917,6 +918,17 @@ void KviMainWindow::contextMenuEvent(QContextMenuEvent *)
 	// do nothing! avoids builtin popup from qmainwindow
 }
 
+void KviMainWindow::keyPressEvent(QKeyEvent *event)
+{
+	switch(event->key())
+	{
+		case Qt::Key_Alt:
+			g_pMainWindow->showMenubar();
```
un1versal: should probably also include

```
else
case Qt::Key_Alt:
            g_pMainWindow->hideMenubar();
```

So you can also hide by pressing Alt, as is you need to click to hide

src/kvirc/ui/KviMainWindow.cpp
un1versal: I have, and if this is proposed as is I disagree + also doesnt work well to hide if you already un-hidden it, its counter intuitive.

in #1915 you make sane comments and then propose the exact opposite. Sorry I dont get it.

ctrlaltca: > I have, and if this is proposed as is I disagree
> 
> This is a tentative implementation ... I just want to know if it works well for others and on OSes other than Linux.

Once @wodim's knows this works fine, he'll probably move on trying to implement the missing bits and shaping up this PR better. If this doesn't work on other OSes, he won't waste his time polishing up a not-working thing.

un1versal:
I suppose you read https://github.com/kvirc/KVIrc/issues/1917 which is the reason I suggested a statusbar indicator. not because of bikeshedding.

wodim:
I am thinking of adding an entry to the ircview popup called "show menubar", what do you think? Something that sets `AutoHideMenubar` to false.

un1versal:
that maybe useful but not as obvious perhaps, what about a toolbar icon (visible by default), similar to away/notaway that hides/unhides that would probably be more obvious?

I can make the icons 16x16 and 32x32 (4 icons total) if you think thats a good idea and send you a PR to this branch

wodim:
I don't want such a button by default on a toolbar.

un1versal:
What about it becomes visible in toolbar if you select option OR hide menubar?

that would be 2 icons only in that case

we already have something similar, the tray icon that becomes visible in tray if you minimize kvirc even if the option is unchecked

wodim:
Does not make much sense to me since toolbars are customisable by users...

un1versal:
well we need some **obvious** indicator that toolbar is hidden, however it makes more sense to do that.

Else support burden is a pita

IF someone complains (wheres my menu gone?), we ask for a screenshot, if the indicator is there we know its because of this option. If indicator not there we know its some other bug related to #1917 
it makes sense to me.

Without this PR, there are 2 causes (1 known 1 unknown) for menu disappearing with this PR there are 3 causes.  (2 known 1 unknown)

Issue #1917, opened 2016-04-02
No menu at the top anymore


DaVukovic:
### Expected behavior

The expected bahavior would be like this:

![kvirc1](https://cloud.githubusercontent.com/assets/7235787/14226093/7252b3e2-f8d8-11e5-85ec-a66d5bff7b67.png)
### Actual behavior

The actual one is like this:

![kvirc2](https://cloud.githubusercontent.com/assets/7235787/14226101/95114ef2-f8d8-11e5-9b20-d256b14f3ade.png)

And I don't know how to get it back. I just compiled the version from source and this is how it ends up.
### System information

about:

![aboutkvirc](https://cloud.githubusercontent.com/assets/7235787/14226128/4e0884ac-f8d9-11e5-9266-97120bc20120.png)

uname -a:

Linux david-kubuntu 3.13.0-83-generic #127-Ubuntu SMP Fri Mar 11 00:25:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

cat /etc/lsb-release:

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.4 LTS"

Let me know if something is missing.

Cheers

un1versal:
Thx David.

This is now a quite an old issue unreported in tracker till now.

A known cause for this issue (after many tests) is that `appmenu-qt5` when installed causes this... uninstalling it fixes it, however your issue doesn't seem related to that afaict.

I made a suggestion in https://github.com/kvirc/KVIrc/issues/1915#issuecomment-204617058 because of this issue because that will potentially add to confusion with users that have this problem here (I have 3 other reports over IRC for this)

un1versal:
**Somemore information:** related to `appmenu-qt5`its a valid component for other applications but it affects some apps + KVIrc in this manner as well. (menu disappears) 
reference https://bugs.launchpad.net/ubuntu/+source/appmenu-qt5/+bug/1307619

As for the cause for this issue #1917 reported, the cause is **unknown** afaik at this time.

In case of `appmenu-qt5`as cause we can ask ppl to remove it if it doesn't affect

**Question:** Can we do something in kvirc to prevent appmenu-qt5 from causing this issue?

un1versal:
@1977David you have KDE installed or something? Using Plasma5?

DaVukovic:
@un1versal 

[...]

PR #1919, opened 2016-04-02
Option to hide the menu bar

wodim:
Redux of #1916, now Ctrl+M toggles it, à la KDE. Cleaner and it works better


src/kvirc/ui/KviMenuBar.cpp
```diff
@@ -240,21 +250,13 @@ void KviMenuBar::setupMainPopup(QMenu *pop)
...
 
-	main->addSeparator();
-
 	// FIXME: Add a "Dock to tray" icon if the tray is not visible (or show tray icon or whatever)
 
 // Qt/Mac creates a Quit item on its own <= this is bad
 #ifndef COMPILE_ON_MAC
 	main->addSeparator();
 
-	main->addAction(
```
un1versal, line 251: Is there a particular reason why we need to remove the quit action from menu?
Not sure if this is necessary tbh.

src/kvirc/ui/KviMenuBar.cpp
```diff
@@ -174,6 +176,14 @@ void KviMenuBar::setupSettingsPopup(QMenu *pop)
 	QAction *pAction = opt->addAction(*(g_pIconManager->getSmallIcon(KviIconManager::Toolbar)),__tr2qs("Toolbars"));
 	pAction->setMenu(m_pToolbarsPopup);
 
+#ifndef COMPILE_ON_MAC
+	m_pMenuBarAction = opt->addAction(*(g_pIconManager->getSmallIcon(KviIconManager::StatusBar)),__tr2qs("Show Menu Bar"),m_pFrm,SLOT(toggleMenuBar()));
```
un1versal: icon wise its not a problem I can make a specific icon.

**edit** done in https://github.com/wodim/KVIrc/pull/1

src/kvirc/ui/KviMenuBar.h
```diff
@@ -56,6 +56,7 @@ class KVIRC_API KviMenuBar : public KviTalMenuBar
 	QMenu                       * m_pActionsToolsPopup;
 	KviPointerList<KviScriptMenuBarItem> * m_pScriptItemList;
     // Dynamic actions
+    QAction                     * m_pMenuBarAction;
```
un1versal: small nitpick, can we fix the indenting on separate commit for this file while we are here?

**Edit**
Fixed in https://github.com/wodim/KVIrc/pull/1

src/kvirc/ui/KviMainWindow.cpp
```diff
@@ -981,6 +989,41 @@ void KviMainWindow::toggleStatusBar()
...
+void KviMainWindow::toggleMenuBar()
+{
+	if(KVI_OPTION_BOOL(KviOption_boolMenuBarVisible) == true)
+	{
+		if(KVI_OPTION_BOOL(KviOption_boolWarnAboutHidingMenuBar))
+		{
+			// ah, we need an abstraction for this setCheckBox() qt5 shiny stuff
+			QMessageBox pMsgBox;
+			QCheckBox cb(__tr2qs("Do not show this message again"));
+			pMsgBox.setText(__tr2qs("This will hide the menu bar completely. "
```
un1versal, line 1001: The warning is a nice touch, but I still think we should have a visible indicator that menu bar was hidden, maybe somewhere in statusbar or something.

We really need this imo, dealbreaker for me.

**Edit**
The Icon I added in https://github.com/wodim/KVIrc/pull/1 can be used as a visibility indicator

src/kvirc/ui/KviMenuBar.cpp
un1versal: also...

`__tr2qs("Show Menu Bar")`

I think this should instead read
`__tr2qs("Hide Menu Bar")`

Thats what we are doing after all?

**edit**
Fixed in https://github.com/wodim/KVIrc/pull/1

un1versal:
FYI this works as desired, just the comments really.

Ill make an icon for this now that will suit bith the action and be usable for a visibility indicator at same time.

un1versal:
OK you have https://github.com/wodim/KVIrc/pull/1 waiting to merge, I added an icon, assigned it and addressed 2 of my own comments to save time.

Only one comment left after you merge my PR to your branch.

src/kvirc/ui/KviMenuBar.cpp
un1versal, line 251: I actually think this bit is only cleanup, cant see any difference at first glace

wodim, line 251: I have replaced it with the `ACTION_POPUP_ITEM` macro that does the same.

src/kvirc/ui/KviMenuBar.cpp
wodim: Negative options are better avoided.

src/kvirc/ui/KviMainWindow.cpp
wodim, line 1001: The warning should be enough. It's the same KDE does, I even took the text verbatim.

src/kvirc/ui/KviMenuBar.cpp
wodim: I see you've done the change in https://github.com/wodim/KVIrc/pull/1 , undo it because it's not even working properly now (checkbox always selected).

src/kvirc/ui/KviMenuBar.cpp
un1versal, line 251: ah... k

src/kvirc/ui/KviMenuBar.cpp
un1versal: This option doesnt show anything, it hides it, so logic fail.

Why also would the option name affect checkbox selection? what?

@ctrlaltca option hides it what do you think it should be named?

src/kvirc/ui/KviMainWindow.cpp
un1versal, line 1001: I disagree, especially because we have #1917 indicator would help us debug the bug vs option

src/kvirc/ui/KviMenuBar.cpp
wodim: The option is marked by default, like the one we already have for the status bar:

src/kvirc/ui/KviMainWindow.cpp
wodim, line 1001: If the user is stupid enough to bypass this warning and then not know what has happened... In any case you can always ask him to press Ctrl+M.

src/kvirc/ui/KviMenuBar.cpp
un1versal: @wodim, ive removed the label change from my PR, but with that this will not be OK to merge

Option does not SHOW menu, it hides it so it should be named accordingly

As is -1

DarthGandalf: @un1versal Absence of double negations is not confusing.

PR #1892, opened 2016-03-02
Hopefully fix IPC sentinel

ctrlaltca:
Fix #1885 restoring the old name of the kvirc4_ipc_sentinel

Also, remove old Qt3 workarounds

PR #1924, opened 2016-04-07
Remove all ifdefs related to older versions of Qt.

wodim:
Now that we only support Qt 5 we will no longer need these.

### DIFF
--- a/src/kvilib/core/KviQString.cpp
+++ b/src/kvilib/core/KviQString.cpp
@@ -38,10 +38,6 @@
 #include <stdio.h> // for sprintf()
 #include <QRegExp>
 
-#if (QT_VERSION < 0x050000)
-#include <QTextDocument> // for Qt::escape()
-#endif
-
 // kvi_string.cpp
 extern unsigned char iso88591_toLower_map[256];
 extern unsigned char iso88591_toUpper_map[256];
@@ -1306,10 +1302,6 @@ namespace KviQString
 
 	QString toHtmlEscaped(QString szData)
 	{
-#if (QT_VERSION >= 0x050000)
 		return szData.toHtmlEscaped();
-#else
-		return Qt::escape(szData);
-#endif
 	}
 }
--- a/src/kvirc/ui/KviInputEditor.cpp
+++ b/src/kvirc/ui/KviInputEditor.cpp
@@ -63,16 +63,11 @@
 #include <QFontMetrics>
 #include <QKeyEvent>
 #include <QDragEnterEvent>
-#if (QT_VERSION < 0x050000)
-	#include <QInputContext>
-#endif
 #include <QMenu>
 #include <QWidgetAction>
 #include <QTextBoundaryFinder>
 
-#if (QT_VERSION >= 0x050000)
-	#include <qdrawutil.h> // qDrawShadePanel
-#endif
+#include <qdrawutil.h> // qDrawShadePanel
 
 #if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
 	#include <windows.h>
--- a/src/kvilib/net/KviHttpRequest.h
+++ b/src/kvilib/net/KviHttpRequest.h
@@ -40,11 +40,7 @@
 
 #include <QObject>
 
-#if (QT_VERSION >= 0x050000)
-	#include <QtNetwork/QAbstractSocket>
-#else
-	#include <QAbstractSocket>
-#endif
+#include <QtNetwork/QAbstractSocket>
 
 class KviDnsResolver;
 class KviDataBuffer;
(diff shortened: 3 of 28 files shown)


staticfox:
May as well grab https://github.com/wodim/KVIrc/blob/qt5/src/kvirc/ui/KviInputEditor.cpp#L1775 as well

wodim:
I left it because of: https://github.com/wodim/KVIrc/blob/qt5/src/kvirc/ui/KviInputEditor.cpp#L1137-L1141

I also left the IPC stuff in place because god knows what it does (at least we are certain it's broken)

un1versal:
+1 for cleanups 

> I also left the IPC stuff in place because god knows what it does (at least we are certain it's broken)

IPC is broken in windows at least there was https://github.com/kvirc/KVIrc/pull/1892 but theres still no fix for it.
Also you know what it does, its already been explained to you in the above closed PR.

src/kvilib/core/KviQString.cpp
```diff
@@ -38,10 +38,6 @@
 #include <stdio.h> // for sprintf()
 #include <QRegExp>
 
-#if (QT_VERSION < 0x050000)
-#include <QTextDocument> // for Qt::escape()
```
un1versal, line 42: ~~Why not adding the `#include <QTextDocument>`?~~

nm I see reason now.

src/kvirc/ui/KviInputEditor.cpp
```diff
@@ -63,16 +63,11 @@
 #include <QFontMetrics>
 #include <QKeyEvent>
 #include <QDragEnterEvent>
-#if (QT_VERSION < 0x050000)
-	#include <QInputContext>
```
un1versal, line 67: why not add `#include <QInputContext>`? and also what about the https://github.com/wodim/KVIrc/blob/qt5/src/kvirc/ui/KviInputEditor.cpp#L1136-L1152 why leave the else and why not remove the `#if (QT_VERSION >= 0x050000)`

un1versal:
I was trying to understand some of these changes, can anyone explain and reply my questions?

wodim:
Told ya, I left it there because the Qt 5 helper functions are still not complete.

un1versal:
told me where, cant see a reply related to that.

> I left it because of: https://github.com/wodim/KVIrc/blob/qt5/src/kvirc/ui/KviInputEditor.cpp#L1137-L1141

that doesnt explain why you left the else for other qt version anyway

wodim:
https://github.com/kvirc/KVIrc/pull/1924#issuecomment-206819273

src/kvirc/ui/KviInputEditor.cpp
ghost, line 67: `QInputContext` was removed from Qt in Qt5, so we need this version check.

un1versal:
idk its probably me, but the https://github.com/wodim/KVIrc/blob/qt5/src/kvirc/ui/KviInputEditor.cpp#L1142-L1148 which you left will never be called for Qt5 so again, why leave it?

wodim:
I left it there because there's no equivalent of that code for Qt 5 for now, but in the future there probably will. So it's sort of a TODO. Read the comment above the code.

un1versal:
yes but that code isnt run, or is it run? I did read comment, I still dont understand it and you dont care to explain clearly, if the code I pointed to in https://github.com/wodim/KVIrc/blob/qt5/src/kvirc/ui/KviInputEditor.cpp#L1142-L1148 is run then youre correct else makes no sense.

ghost:
The code isn't run in Qt5, but since Qt5 doesn't have a solution for this, think of that block as a comment block for anyone that wants to update or check if newer Qt5/Qt6 versions have it built-in. Removing the code would hide a potential problem.

un1versal:
as a comment yes thats obvious helpful, as is just left makes no sense, which is why I asked trying to understand reasoning.

wodim:
It's not compiled so what is the difference between that and a comment? It's the same as comenting something by wrapping it inside `#if 0` tags.

un1versal:
the difference is some code sanity, it seems loads of stuff is just left in code waiting for years for some unknown reason that soon gets forgotten about.

if anyone wanted to make code cleanups its utterly messy.

Arfrever:
Qt 5 (at least 5.5.1) has `QPlatformInputContext`. Maybe it could be used?

un1versal:
Qt 5.5 isnt default in all distros, you be OK for windows, and I doubt anyone is going to package and compile Qt in all these Other OS/distros flavours specially..

Arfrever:
I have now checked and `QPlatformInputContext` class is present in Qt 5.0.0:
https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/kernel/qplatforminputcontext.h?h=v5.0.0

Issue #1854, opened 2016-01-25
Port to qtwebengine for Qt-5.6


Dessa:
~~Webkit is going away, building qtwebkit for 5.6 requires manual effort; i'm not sure which distros will do it (i didn't look exstensively, but i have seen it in Fedora Rawhide)~~

there is a porting document available at https://doc.qt.io/qt-5.15/qtwebenginewidgets-qtwebkitportingguide.html

see https://github.com/kvirc/KVIrc/issues/1854#issuecomment-325636363

annulen:
For distros there is absolutely no difference if QtWebKit is provided in official binaries or not, since they build all of Qt from source anyway.
I can also assure you that QtWebKit won't disappear from distros overnight, since a lot of applications depend on it. BTW, for 5.6 release we've prepared a number of fixes and new QML APIs, it won't be just a token release copying 5.5.1.
In case you are interested there is a project to upgrade QtWebKit to modern WebKit code base, you can find more info at https://github.com/annulen/webkit/wiki (It is still very far from usable state but we are making progress)

wodim:
Several distros have refused to package QtWebEngine ~~because of the mental handicap of some maintainers~~ because how retarded some distro policies are.

https://marc.info/?l=kde-core-devel&m=142954900813235&w=2

annulen:
Please don't offend Lisandro, he is doing his best to make Qt packages meeting Debian policies.

wodim:
Yeah sorry, I did not mean to say the maintainer is at fault here; the Debian policies are retarded instead.

un1versal:
The good thing is that saner distros exist with slightly more relaxed policies, the problem is that many software titles suffer as a result of the strict unbending rules of the other distros who refuse to care about anything other than the OS and not the apps who depend on these things.

So _KVIrc_ has to do something about it if it wants to run everywhere smoothly.
With the lack of manpower coding for KVIrc changes like this are not going to come easy.

Dessa:
there was always some middle ground to be had here (just keep using the webkit distros provide) but I think that middle ground got slightly more upstream-y in the last couple of weeks

so that means qtwebkit is *probably* not going away but instead will get its own release cycle (?)
I did not test that version yet, so I can't say if it requires changes.
I also have no idea if there will be a windows version built and distributed with the qt installer, so it will in turn, land on the build server for windows again, then again I didn't hear complaints about the lack of webkit support in kvirc for windows either.

wodim:
[...]

PR #1926, opened 2016-04-08
[AV] switch to qt 5.6

Dessa:
see if it will suffer from #1854 or not


un1versal:
Those appveyor failures dont make sense to this change.

Dessa:
the 64 bit one does. closing until we can actually fix #1854 since we cannot rely on a self-built qtwebkit 5.6 without spending too much time for it.

PR #1930, opened 2016-04-13
Fix typos in Events/function/variables/options names Estabilished -> Established

un1versal:
#### Changes proposed

~~\- Fixes various typos and capitalization across varied files~~
~~\- Fixes remainder of bolierplating typos and omissions~~
- Fixes typos in Events/function/variables names containing Estabilished -> Established

@staticfox if you feel like having a looksee ;)

### DIFF
--- a/doc/FAQ
+++ b/doc/FAQ
@@ -545,7 +545,7 @@
   However, if you router's IP has a dns entry, you can ask KVIrc to fill
   this field with a simple script to be performed at every connection.
 
-  event(OnIRCConnectionEstabilished,updatedccsource)
+  event(OnIRCConnectionEstablished,updatedccsource)
   {
       ahost(yourrouterhostname.dyndns.org)
       {
--- a/src/kvilib/net/KviHttpRequest.cpp
+++ b/src/kvilib/net/KviHttpRequest.cpp
@@ -269,7 +269,7 @@ void KviHttpRequest::slotSocketConnected()
 		m_p->pConnectTimeoutTimer = NULL;
 	}
 
-	emit connectionEstabilished();
+	emit connectionEstablished();
 	emit status(
 			__tr2qs("Connected to %1:%2: sending request")
 					.arg(m_p->pSocket->peerAddress().toString())
--- a/src/kvilib/net/KviHttpRequest.h
+++ b/src/kvilib/net/KviHttpRequest.h
@@ -170,7 +170,7 @@ public:
 signals:
 	void resolvingHost(const QString &hostname);
 	void contactingHost(const QString &ipandport);
-	void connectionEstabilished();
+	void connectionEstablished();
 	void receivedResponse(const QString &response);
 
 	void terminated(bool bSuccess);
--- a/src/kvirc/kernel/KviIrcConnection.cpp
+++ b/src/kvirc/kernel/KviIrcConnection.cpp
@@ -316,7 +316,7 @@ void KviIrcConnection::start()
 	m_pLink->start();
 }
 
-void KviIrcConnection::linkEstabilished()
+void KviIrcConnection::linkEstablished()
 {
 	m_eState = Connected;
 
@@ -334,7 +334,7 @@ void KviIrcConnection::linkEstabilished()
 	}
 
 	// FIXME: With STARTTLS this is called TWICE!
-	context()->connectionEstabilished();
+	context()->connectionEstablished();
 
 	// Ok...we're loggin in now
 	resolveLocalHost();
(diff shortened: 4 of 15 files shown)


data/man/kvirc.fr.1
```diff
@@ -64,14 +64,14 @@ Use this port for connection.
 This option requires also a server name.
 .TP 8
 .B [ircurl]
-An url in the following form:
+An URL in the following form:
```
staticfox: It's `a URL` not `an URL`

staticfox:
I have reservations on changing the KVS event handler name since that may break existing scripts? (I rarely dive in to the inner mechanics of KVS). I'll look in to it in a bit to verify if it will or not, but someone can clarify in the mean time.

un1versal:
> I have reservations on changing the KVS event handler name since that may break existing scripts? (I rarely dive in to the inner mechanics of KVS). I'll look in to it in a bit to verify if it will or not, but someone can clarify in the mean time.

yes, I thought of that also, but 4.2.0 -> 5.x.x will break much anyway so in a sense one more one less...
That said sod it, Im not that invested in this correction **but either do all or none**

un1versal:
I split the other changes and merged them as they were straight forward fixes,

I left this one since its the one that is really needs clarification/investigation

un1versal:
@TheReign for advice/thoughts because he does or knows ppl that do scripting and probably be affected if this breaks things.

un1versal:
One more try, lets see if @OmegaPhil can provide some insight.

OmegaPhil:
I'm for fixing obvious typos in code, and as you're saying earlier, a major version change allows for breakage.

None of my scripts reference anything involving 'estabilished', so you could say that is unfair for me to comment here, however it still seems that KVIrc scripting is not a 'big thing' yet, so its not like mIRC where you risk breaking countless hours of work.

So I would go ahead with this change.

un1versal:
Thx for taking the time to reply. 

The reason for asking is even if your scripts dont use these, you are more familiar with this things and  that experience and knowledge in form of feedback or whatever, helps more than you might think.

> however it still seems that KVIrc scripting is not a 'big thing' yet

Lets hope this is closer rather than further away into the future.

wodim:
My vote is no. Unnecessary breakage.

HelLViS69:
Il 16/apr/2016 16:54, "un1versal" notifications@github.com ha scritto:

> Thx for taking the time to reply.
> 
> The reason for asking is even if your scripts dont use these, you are
> more familiar with this things and that experience and knowledge in form of
> feedback or whatever, helps more than you might think.
> 
> > however it still seems that KVIrc scripting is not a 'big thing' yet
> 
> Lets hope this is closer rather than further away into the future.

Maybe it's better to rename the function, and add a new function with the
old wrong name ad an alias, including in the doc that it's deprecated

HelLViS69

un1versal:
The functions are internal and are not exposed to or matter users.

@HelLViS69 

the only one which matters scripting wise is **one single event** OnIRCConnectionEstablished

FYI the internal KVIrc documentation referenced the right name already for this event **(you added that documentation according to git blame back in 2008)**, so if anything, anyone used doc, it wouldn't have worked and this is in fact a fix for a buggy event name.

Reference https://github.com/kvirc/KVIrc/blob/7bc5592c3a556d6fd8e1738d203a4e5b87a3a4ba/src/kvirc/kvs/event/KviKvsEventTable.cpp#L593-L597

wodim:
Ah, I thought it was the event name. You could fix this too: 

https://github.com/kvirc/KVIrc/blob/a589f834e10f8767cbfb7ac26485cfcd07ff2214/src/kvirc/kvs/KviKvsScript.cpp#L151

https://github.com/kvirc/KVIrc/blob/a589f834e10f8767cbfb7ac26485cfcd07ff2214/src/kvirc/kvs/KviKvsScript.cpp#L159

HelLViS69:
Il 16/apr/2016 17:15, "un1versal" notifications@github.com ha scritto:

> The functions are internal and dont matter to users.
> 
> the only one which matters scripting wise is one single event
> 
> @HelLViS69
> FYI the internal KVIrc documentation referenced the right name already
> (you added that documentation according to git blame), so if anything,
> anyone used that, it wouldn't have worked and this is in fact a fix for a
> buggy event name.

Ok, so just grep those name and in case sed on all files..

HelLViS69

un1versal:
> Ah, I thought it was the event name. You could fix this too:
> 
> https://github.com/kvirc/KVIrc/blob/a589f834e10f8767cbfb7ac26485cfcd07ff2214/src/kvirc/kvs/KviKvsScript.cpp#L151
> 
> https://github.com/kvirc/KVIrc/blob/a589f834e10f8767cbfb7ac26485cfcd07ff2214/src/kvirc/kvs/KviKvsScript.cpp#L159

thats a bug right there, fix incoming, thx

un1versal:
> Ok, so just grep those name and in case sed on all files..

Im sure this is all of them except on .po files those get handled automatically on next Transifex sync

PR #1932, opened 2016-04-14
Part 2 of 99925aa: Also check and conditionally disable buggy appmenu-qt5 against known buggy Qt version

un1versal:
After the discussion in https://github.com/kvirc/KVIrc/commit/99925aa796ad1319f6f374fedc9a6d5907294b36  this approach would solve @TheReign concerns

> TheReign commented on 99925aa 12 hours ago
> 
> This patch breaks this feature and I'm now forced to run custom patches to run KVIrc as a proper app. This is really disappointing.

---

Seems the ideal thing is to only disable appmenu-qt5 if the Qt version is known to be buggy this is what @ctrlaltca indicated the issue only happens with Qt 5.2.1 which is widespread against LTS releases.

I suggested https://github.com/kvirc/KVIrc/commit/99925aa796ad1319f6f374fedc9a6d5907294b36#commitcomment-17085495 sort of similar to this and then it occurred to me why not this? So here it is this try

 Im not even sure if this is 100% correct, but the idea seems sound, I rather do this that waste time chasing future support and also not disappoint @TheReign and when this version is phased out the revert is obvious and simple

I would need guidance to implement this properly if this happens to be a total botch.
#### Changes proposed
- Also check for the known buggy Qt version and only disable if condition is met.

@Thereign @ctrlaltca for thoughts and feedback.

### DIFF
--- a/src/kvirc/kernel/KviMain.cpp
+++ b/src/kvirc/kernel/KviMain.cpp
@@ -23,7 +23,7 @@
 //=============================================================================
 
 
-
+#include "KviRuntimeInfo.h"
 #include "KviApplication.h"
 #include "KviCString.h"
 #include "kvi_settings.h"
@@ -391,7 +391,9 @@ int main(int argc, char ** argv)
 	qputenv("QT_BEARER_POLL_TIMEOUT", QByteArray::number(-1));
 #endif
 
-	if(qgetenv("QT_QPA_PLATFORMTHEME") == QByteArray("appmenu-qt5"))
+	// appmenu-qt5 in Qt 5.2 is buggy and when installed breaks KVIrc's menu bar #1917
+	// Check if lib is installed and that Qt version is the known buggy and disable it only if those conditions are met.
+	if(qgetenv("QT_QPA_PLATFORMTHEME") == QByteArray("appmenu-qt5") && KviRuntimeInfo::qtVersion() == "5.2.1")
 		qunsetenv("QT_QPA_PLATFORMTHEME");
 
 	KviApplication * pTheApp = new KviApplication(argc,argv);


wodim:
We should check if the version is minor than 5.3 or whatever fixed it.

un1versal:
> We should check if the version is minor than 5.3 or whatever fixed it.

Sure, I agree with that, idk which fixed it tbh, not a clue.

Code wise is this OK though?

wodim:
I guess so.

ghost:
The approach you used is probably better than doing `#if` Qt version checks.  The only issue I see is figuring out which versions are affected and adding them.  I'm also curious if the 5.2.1 that we've seen having this issue applies to all 5.2.1 or just debian 5.2.1 releases.

On a side note, on my machine that has appmenu-qt5 installed, after the menu is hidden and moved to the gloabl menu bar, I can bring it back with `ctrl+m` so that both menus appear.  This fix may not be **as** critical as it was before we added the hotkeyed hide/unhide menu.

un1versal:
Yea, idk which Qt versions are specifically affected, I dont use debian and I never had this issue tbh. Also never used KDE Kubuntu or similar, my repro was done in Host OS Linux mint 17.3 Cinnamon.

If anyone has some idea what versions other before/after  other than 5.2.1 which are affected by bug please let me know here (I tried looking at Qt website but found nothing).

Any assistance is appreciated 

Hope that with this you dont need workarounds in your special builds, idk why you need special builds but hey its opensource. ;)

@1977David
Can you please compile this ontop and see if your problematic machine still works with this?

ghost:
I said Debian because Ubuntu pulls the Qt package from upstream Debian and likely doesn't make much modifications to it, so I would expect to see this issue on any Debian derivative using Qt5.2.1 if the problem is with the package and not Qt itself.

un1versal:
Yes that makes perfect sense, it wouldn't be the only time things like that happen and this not the only package mangled by _debianisms_.

Still no idea which version doesnt have this issue anymore to change the check for, waiting on @ctrlaltca for possible insights.

ctrlaltca:
This is mostly a problem with ubuntu releases.
Ubuntu 12.x : Qt 5.0.2  is unusable.
Ubuntu 14.4 LTS:  Qt 5.2.1 is usable, but the appmenu-qt5 package is buggy:
- hotkeys binded to menu items doesn't work;
- even if you don't use unity, it will create problems.

Ubuntu 14.10 has Qt5.3, but it's EOL.
Ubuntu >= 15.04 have Qt 5.4.1, and it should work fine.

wodim:
And how would we check the appmenu-qt5 version? Or work around it somehow. The QPA API is almost undocumented still...

un1versal:
We dont need to check against EOL versions

Problem I see here is non Ubuntu and do we need to check them? If so how...

@1977David said Kubuntu/KDE doesnt use universal menu (appmenu-qt5 uses) which @TheReign said this is and Linux mint 17.3 cinnamon also doesnt afaik.

:/

TBH soon as 16.04LTS is out with Qt 5.5.1 and 13.04LTS is EOL we dont need this check

ctrlaltca:
Maybe we can check the `XDG_CURRENT_DESKTOP` environment variable and disable if it's anything but Unity?

un1versal:
So this isnt an issue on unity? IDK never used it either

TBH soon as 16.04LTS is out with iirc Qt 5.5.1 and 14.04LTS is EOL we dont need this check sadly only till 2019

wodim:
#1917 is about KDE, although I installed Kubuntu 14.04 and I could not reproduce it. appmenu-qt5 worked properly.

un1versal:
I managed to repro in Linux Mint cinnamon 17.3 (14.04LTS) host OS only (not VM)

DaVukovic:
@wodim 

you are absolutely right. More or less it was my fault. So I have to apologize in a bit. I know of that global menu bar while using Unity and I googled a bit and found how to enable the global menu bar while using Kubuntu. It was disabled by default and I never saw that setting. After enabling nearly any app has lost its menu bar which is now at the top. I didn't test with KVIrc yet, but I will do that immediately tomorrow morning and will report as fast as I can. 

This is the link to it: 

http://www.webupd8.org/2013/02/how-to-enable-new-kde-410-appmenu-title.html

I didn't even know that a global menu bar for Kubuntu exists...so 1000 sorries for that ;)

un1versal:
Well that doesnt explain why I repro the issue in non KDE OS, Linux Mint Cinnamon, has no such setting

In any case here is the clincher, none of appmenu-gtk3 appmenu-gtk is required by KVIrc so if this PR is closed and *fix* reverted which is what should happen, who on earth will support that cluster fuck of missing packages in OS?

I wont.

un1versal:
@1977David you should check if  `appmenu-gtk3 appmenu-gtk` is installed if they are and you cant make it work then idk anymore since wodim says it works and Im not installing pos kubuntu.

GTK also has some history of breaking KVIrc, and support is disabled by default... idk what the consequences are of that and tbh Ive had enough of this **issue** which btw I never had but apparently @1977David is not alone there since more than one person moaned about it.

After @ctrlaltca confirmed the issue in any case we need a proper solution to at least 2019 thats when 14.04 LTS goes EOL that is unless we can get Qt 5.5.1 for current LTS in a ppa properly packaged.

un1versal:
SO Im slightly confused with this now, what do I need to do here?

un1versal:
while we wait for some instructions on what to do here there we go I asked alex if he would do this for us and in true form hes come through as usual....

https://github.com/kvirc/KVIrc/issues/1949

Issue #900, opened 2015-07-26
Window width properties are lost for userlist / topic / modes, upon KVIrc's restart.


DarthGandalf:
**Reported by Corin on 11 Aug 2010 17:11:41 UTC**
I've had this issue every since Insomnia and it's still happening in Equilibrium. I have noticed that this doesn't happen if you close each tab, individually, as long as each tab is currently selected, when you close it.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/900

DarthGandalf:
**Comment by HelLViS69 on 11 Aug 2010 22:58:49**
confirmed<br/>
I tried also to remove winproperties.kvc

DarthGandalf:
**Comment by Corin on 12 Aug 2010 22:10:52**
Yes, I forgot to mention that doing so hadn't worked.

DarthGandalf:
**Comment by anonymous on 03 Sep 2010 20:11:41**
I can confirm this annyoing bug.

It happens when trying to resize a user's list.
I attached several versions of winproperties.kvc
Hope this helps.

DarthGandalf:
**Comment by Stanzilla on 23 Sep 2010 09:47:32**
Confirmed, and it's annoying as hell.

DarthGandalf:
**Comment by Corin on 04 Feb 2011 06:45:10**
Disabling Preferences > Interface > Remember window properties and deleting the winproperties file seems to solve the issue entirely.

DarthGandalf:
**Comment by Corin on 04 Feb 2011 06:46:22**
Disabling Preferences > Interface > Remember window properties and deleting the winproperties file in the config folder seems to solve the issue entirely.

DarthGandalf:
**Comment by ctrlaltca on 12 Aug 2011 14:06:53**
Corin's workaround is fine if you just don't care about rkvirc remebering previous customizations to the interface. Anyway, r5923 reintroduces a "sizes handling"mechanism that emulates the old kvirc3's way, hopefully making kvirc4 more user-friendly.
The first times you run kvirc with this patch you could experience some oversized userlists; they are supposed to appear when an old (broken) size value had been saved inside the kvirc config. Just resize the userlist as you like it best and it should work from the second time on.
Please comment/feedback on this

---

Owner: pragmaware → ctrlaltca
Status: new → assigned

DarthGandalf:
**Comment by DarthGandalf on 17 Sep 2011 11:06:52**
[...]

PR #1942, opened 2016-04-19
Add option to hide the taskbar button.

Cizzle:
Adds an option to hide/show the taskbar button, useful for when you have a second screen for kvirc only. Recommended to be used in combination with the trayicon. Needs a restart for the option to take effect.
Tested on Win7 only.

### DIFF
--- a/src/kvirc/kernel/KviApplication.cpp
+++ b/src/kvirc/kernel/KviApplication.cpp
@@ -1605,7 +1605,14 @@ void KviApplication::createFrame()
 {
 	Q_ASSERT(g_pMainWindow == nullptr);
 
-	new KviMainWindow();
+#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
+	if(KVI_OPTION_BOOL(KviOption_boolShowTaskBarButton))
+		new KviMainWindow(0);
+	else
+		new KviMainWindow(new QWidget(0, 0));
+#else
+	new KviMainWindow(0);
+#endif
 
 	Q_ASSERT(g_pMainWindow != nullptr);
 
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -341,6 +341,7 @@ KviBoolOption g_boolOptionsTable[KVI_NUM_BOOL_OPTIONS]=
 	BOOL_OPTION("MenuBarVisible",true,KviOption_sectFlagFrame | KviOption_resetUpdateGui),
 	BOOL_OPTION("WarnAboutHidingMenuBar",true,KviOption_sectFlagFrame),
 	BOOL_OPTION("WhoRepliesToActiveWindow",false,KviOption_sectFlagConnection),
+	BOOL_OPTION("ShowTaskBarButton",true,KviOption_sectFlagFrame)
 };
 
 #define STRING_OPTION(_txt,_val,_flags) KviStringOption(KVI_STRING_OPTIONS_PREFIX _txt,_val,_flags)
--- a/src/kvirc/kernel/KviOptions.h
+++ b/src/kvirc/kernel/KviOptions.h
@@ -356,8 +356,9 @@ DECLARE_OPTION_STRUCT(KviStringListOption,QStringList)
 #define KviOption_boolMenuBarVisible 266
 #define KviOption_boolWarnAboutHidingMenuBar 267
 #define KviOption_boolWhoRepliesToActiveWindow 268                       /* irc::output */
+#define KviOption_boolShowTaskBarButton 269                              /* windows only, used in KviApplication::createFrame only */
 
-#define KVI_NUM_BOOL_OPTIONS 269
+#define KVI_NUM_BOOL_OPTIONS 270
 
 
 #define KVI_STRING_OPTIONS_PREFIX "string"
--- a/src/modules/options/OptionsWidget_interfaceFeatures.cpp
+++ b/src/modules/options/OptionsWidget_interfaceFeatures.cpp
@@ -53,14 +53,23 @@ OptionsWidget_interfaceFeatures::OptionsWidget_interfaceFeatures(QWidget * paren
 #else
 	addBoolSelector(0,5,0,5,__tr2qs_ctx("Require Ctrl to be held down to copy text","options"),KviOption_boolRequireControlToCopy);
 #endif
+#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
+	addBoolSelector(0,6,0,6,__tr2qs_ctx("Show taskbar button (change requires restart)","options"),KviOption_boolShowTaskBarButton);
+	g = addGroupBox(0,7,0,7,Qt::Horizontal,__tr2qs_ctx("Open Dialog Window for","options"));
+#else
 	g = addGroupBox(0,6,0,6,Qt::Horizontal,__tr2qs_ctx("Open Dialog Window for","options"));
+#endif
 	addBoolSelector(g,__tr2qs_ctx("Preferences","options"),KviOption_boolShowGeneralOptionsDialogAsToplevel);
 	addBoolSelector(g,__tr2qs_ctx("Registered users","options"),KviOption_boolShowRegisteredUsersDialogAsToplevel);
 	addBoolSelector(g,__tr2qs_ctx("Identity","options"),KviOption_boolShowIdentityDialogAsToplevel);
 	addBoolSelector(g,__tr2qs_ctx("Servers","options"),KviOption_boolShowServersConnectDialogAsToplevel);
 	addBoolSelector(g,__tr2qs_ctx("Join channels","options"),KviOption_boolShowChannelsJoinDialogAsToplevel);
 
+#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
+	addRowSpacer(0,8,0,8);
+#else
 	addRowSpacer(0,7,0,7);
+#endif
 }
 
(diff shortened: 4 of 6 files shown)


staticfox:
In the future, please push your changes to a separate branch other than master.

po/modules/options/options.pot
```diff
@@ -2002,6 +2002,10 @@ msgstr ""
 msgid "Require Ctrl to be held down to copy text"
 msgstr ""
 
+#: src/modules/options/OptionsWidget_interfaceFeatures.cpp:55
```
un1versal: Theres no point adding the po changes, please drop

See https://github.com/kvirc/KVIrc/wiki/Updating-KVIrc's-translations

un1versal:
Looks excellent thanks

**Edit**
I cant really test this so someone with dual monitor setup that can test would be ideal to provide feedback and thoughts.

staticfox:
Can you screenshot the result with the option enabled and disabled? Can't get a visual difference on Windows 10.

Cizzle:
@un1versal sure you can test, it just hides the taskbar button, you don't need a second monitor for that.
@staticfox win10 still has a taskbar right, which has buttons for applications that are running? It's like when you minimize KVIrc to the systray (when enabled), the taskbar button disappears then too. But now you can have the same but with the window always open.

un1versal:
so what happens here say, if a user disables tray icon 100% and also disables this, where does kvirc goto when you minimize it?

Cizzle:
Minimizes to a corner (lowerleft) above the taskbar to the smallest size the titlebar can be resized to.

un1versal:
Ill test this in Linux and see...

un1versal:
Sorry to say this seems to have no effect in Linux whatsoever. the taskbar entry is present immediately as soon as KVIrc its restarted.

> Minimizes to a corner (lowerleft) above the taskbar to the smallest size the titlebar can be resized to.

I would be curious to see how this looks like, can you grab a screenshot?

Cizzle:
See attached image.
kvirc minimized wo taskbar button

un1versal:
Ah I see thx,

Do you experience any side effect caused by this? Im thinking about known issues like https://github.com/kvirc/KVIrc/issues/900 for instance.

Cizzle:
Not that I can see, though I haven't been using this enhancement that long.
Will see if I can test this on Linux.

un1versal:
Any more advances here Linux side?

Cizzle:
Have tested this on Ubuntu 16.04 with default Gnome where I suppose the sidebar is equivalent with the taskbar in Windows, and doesn't hide it from there. There isn't a systray there either though.
Could be a Windows-only option, though I can't test on win8 or win10.

un1versal:
well, then ifdef the option so it works on windows only, as Linux I doubt it is even any similarity here.

@staticfox mentioned something that windows 10? this is controlled by the OS so you need to ifdef then for specific windows versions...

maybe he can clarify this more.

Cizzle:
Confirmed to be working on Windows 7, 8.1 and 10. I don't know why it doesn't work for @staticfox.
For Linux I've seen methods with altering the NET_WM_STATE though I couldn't get this working.

Is an ifdef really needed though, as it didn't seem to affect kvirc's normal working? Might be it does work on other window managers, as I've only tested it on GNOME.

un1versal:
It doesnt work in Linux mint Cinnamon) and tbh having options on UI that dont do anything or are incompatible with any particular OS would be a no no imo.

This would also have to be tested in OSX to see if the ifdef or ifndef has to include Linux/OSX or just Linux in general.

src/kvirc/kernel/KviApplication.cpp
```diff
@@ -1605,7 +1605,14 @@ void KviApplication::createFrame()
 {
 	Q_ASSERT(g_pMainWindow == nullptr);
 
-	new KviMainWindow();
+#ifdef COMPILE_ON_WINDOWS
```
un1versal: #if defined(COMPILE_ON_WINDOWS) || (defined(COMPILE_ON_MINGW)

src/kvirc/kernel/KviApplication.cpp
```diff
@@ -1605,7 +1605,14 @@ void KviApplication::createFrame()
...
+	if(KVI_OPTION_BOOL(KviOption_boolShowTaskBarButton))
+#endif
+		new KviMainWindow(0);
+#ifdef COMPILE_ON_WINDOWS
```
un1versal: #if defined(COMPILE_ON_WINDOWS) || (defined(COMPILE_ON_MINGW)

src/kvirc/kernel/KviOptions.cpp
```diff
@@ -341,6 +341,9 @@ KviBoolOption g_boolOptionsTable[KVI_NUM_BOOL_OPTIONS]=
 	BOOL_OPTION("MenuBarVisible",true,KviOption_sectFlagFrame | KviOption_resetUpdateGui),
 	BOOL_OPTION("WarnAboutHidingMenuBar",true,KviOption_sectFlagFrame),
 	BOOL_OPTION("WhoRepliesToActiveWindow",false,KviOption_sectFlagConnection),
+#ifdef COMPILE_ON_WINDOWS
```
un1versal: and again

src/modules/options/OptionsWidget_interfaceFeatures.cpp
```diff
@@ -53,14 +53,23 @@ OptionsWidget_interfaceFeatures::OptionsWidget_interfaceFeatures(QWidget * paren
 #else
 	addBoolSelector(0,5,0,5,__tr2qs_ctx("Require Ctrl to be held down to copy text","options"),KviOption_boolRequireControlToCopy);
 #endif
+#ifdef COMPILE_ON_WINDOWS
```
un1versal: and again

DarthGandalf: un1versal, it's not defined on mingw?

2016-04-25 14:34 GMT+01:00 un1versal notifications@github.com:

un1versal: See discussion on IRC

src/kvirc/kernel/KviOptions.cpp
```diff
@@ -341,6 +341,9 @@ KviBoolOption g_boolOptionsTable[KVI_NUM_BOOL_OPTIONS]=
 	BOOL_OPTION("MenuBarVisible",true,KviOption_sectFlagFrame | KviOption_resetUpdateGui),
 	BOOL_OPTION("WarnAboutHidingMenuBar",true,KviOption_sectFlagFrame),
 	BOOL_OPTION("WhoRepliesToActiveWindow",false,KviOption_sectFlagConnection),
+#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
+	BOOL_OPTION("ShowTaskBarButton",true,KviOption_sectFlagFrame),
```
un1versal: Im not sure but...

```
#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
    BOOL_OPTION("ShowTaskBarButton",true,KviOption_sectFlagFrame),
  };
#endif
  };
```

staticfox: you need an `#else` otherwise the count won't match up

staticfox: you could just remove the `#if defined` since the option will only be able to be used via other preprocessor guards and make a comment next to it saying it's windows specific.

un1versal: so
```
#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
    BOOL_OPTION("ShowTaskBarButton",true,KviOption_sectFlagFrame),
#else
  };
#endif 
```

staticfox: no..

src/kvirc/kernel/KviOptions.h
```diff
@@ -356,8 +356,14 @@ DECLARE_OPTION_STRUCT(KviStringListOption,QStringList)
 #define KviOption_boolMenuBarVisible 266
 #define KviOption_boolWarnAboutHidingMenuBar 267
 #define KviOption_boolWhoRepliesToActiveWindow 268                       /* irc::output */
+#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
+	#define KviOption_boolShowTaskBarButton 269
```
staticfox: I'd rather not have platform specific option count definitions here

src/kvirc/kernel/KviApplication.cpp
```diff
@@ -1605,7 +1605,14 @@ void KviApplication::createFrame()
 {
 	Q_ASSERT(g_pMainWindow == nullptr);
 
-	new KviMainWindow();
```
staticfox, line 1608:
```
#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
    if(KVI_OPTION_BOOL(KviOption_boolShowTaskBarButton))
        new KviMainWindow(0);
    else
        new KviMainWindow(new QWidget(0, 0));
#else
    new KviMainWindow(0);
#endif
```

staticfox:
jenkins: build this please

src/kvirc/kernel/KviApplication.cpp
un1versal, line 1608:
```
        new KviMainWindow();
#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
    if(KVI_OPTION_BOOL(KviOption_boolShowTaskBarButton))
        new KviMainWindow(new QWidget(0, 0));
#endif
```

PR #1955, opened 2016-04-23
Add clang-format configuration

DarthGandalf:
And reformat everything


src/kvilib/core/KviError.h
```diff
@@ -44,128 +44,129 @@ namespace KviError
...
-		CantSendAZeroSizeFile                 =  94,
-		MissingPopupName                      =  95,
-		ItemPopupOrSeparatorExpected          =  96,
-		SelfModificationNotAllowed            =  97,
-		FeatureNotAvailable                   =  98,
-		UnexpectedCharactersInArrayIndex      =  99,
-		UnexpectedEndInExpression             = 100,
-		UnexpectedEndInArrayIndex             = 101,
-		ProxyHttpFailure                      = 102,
+	enum Code
```
un1versal, line 47: this bit doesnt look as tidy

DarthGandalf, line 47: Yep, see the commit message.
When bugs of clang-format are fixed, the alignment options can be enabled, and this will automatically come back.
I didn't surround this enum by "// clang-format off", because even though it's not very tidy, it's not very dirty either, and it's easily revertable by fixed clang-format from future.

un1versal, line 47: Fair enough. looks good though. impressive by any means.

un1versal:
This seems to have gone down badly with travis & co

is there any reason in particular? these changes seemed pretty straight forward tbh. (seemed, being the operative word)

DarthGandalf:
Yep, I'm debugging it. Looks like it's because of order of includes.

src/kvilib/core/KviError.h
un1versal, line 47: if you like, I can have a look at excluding these and pushing to your branch... tbh waiting for a fix, could be in a week or a year and its easier to undo the exclusions.

DarthGandalf, line 47: I think the non-aligned version is good enough.
The less manual work the better. Cost of requiring human to format code
manually is higher than cost of seeing a non-aligned declaration for a
while.

I'm working on fixing this. Probably a few days?

PR #1960, opened 2016-04-29
Try fix some coverity CID's

un1versal:
These looked fairly straight forward, but its my first time on these waters so have a look.

### DIFF
--- a/src/kvirc/kvs/object/KviKvsObject.cpp
+++ b/src/kvirc/kvs/object/KviKvsObject.cpp
@@ -1262,6 +1262,7 @@ bool KviKvsObject::function_setProperty(KviKvsObjectFunctionCall * c)
 				WRONG_TYPE("unsigned integer")
 			m_pObject->setProperty(szName.toUtf8().data(), QVariant((unsigned int)i));
 		}
+		break;
 		case QVariant::Double:
 		{
 			kvs_real_t i;


src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
```diff
@@ -1787,11 +1787,12 @@ void KviIrcServerParser::parseLiteralNick(KviIrcMessage *msg)
 			old->mergeQuery(q);
 			g_pMainWindow->closeWindow(q); // deleted path
 			if(!msg->haltOutput())
-					old->output(KVI_OUT_NICK,__tr2qs("\r!n\r%Q\r [%Q@\r!h\r%Q\r] is now known as \r!n\r%Q\r"),
-					&szNick,&szUser,&szHost,&szNewNick);
+			old->output(KVI_OUT_NICK,__tr2qs("\r!n\r%Q\r [%Q@\r!h\r%Q\r] is now known as \r!n\r%Q\r"),&szNick,&szUser,&szHost,&szNewNick);
```
DarthGandalf: This definitely looks wrong.

DarthGandalf:
Static analysis has false positives.
How are you going to test that these bugs are genuine, and your fix is actually a fix, and not a new bug, which is visible only in certain circumstances?

un1versal:
> How are you going to test that these bugs are genuine, and your fix is actually a fix, and not a new bug, which is visible only in certain circumstances?

How are you going to test its not?

These comments dont help me at all, its wrong its wrong and tell me why its wrong, when you ask these type questions they dont help.

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
un1versal: the only option here is to

```
            if(!msg->haltOutput())
                old->output(KVI_OUT_NICK,__tr2qs("\r!n\r%Q\r [%Q@\r!h\r%Q\r] is now known as \r!n\r%Q\r"),&szNick,&szUser,&szHost,&szNewNick);
```

since it was only a wrapped line but as is without this its definitely wrong

DarthGandalf:
I don't know if any of these changes are correct. Maybe they are.
If you blindly inserted `break;` in places where Coverity thinks `break;` was missing, you're likely to be wrong, because sometimes missing `break;` is actually correct.

> How are you going to test its not?

The point is, it's your job to prove it. Not mine.
That depends on YOU to understand what exactly are the effects of your change.

In the end, static analysis only hints, that something is probably wrong. It's human's job to triage its findings, and to determine which of them are actual bugs, and which are false positives.

un1versal:
It wasnt per say a blind insert

I looked at all the other entries and more or less the logic made sense. BUT

If you think this is wrong completely and want to be difficult and vague the close button is just below.

src/kvirc/kvs/object/KviKvsObject.cpp
```diff
@@ -1243,6 +1243,7 @@ bool KviKvsObject::function_setProperty(KviKvsObjectFunctionCall * c)
 			if(i < 0)WRONG_TYPE("unsigned integer")
 			m_pObject->setProperty(szName.toUtf8().data(),QVariant((unsigned int)i));
 		}
+		break;
```
DarthGandalf, line 1265: This is the only line from the whole PR, which looks valid for me.

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
```diff
@@ -1787,11 +1787,13 @@ void KviIrcServerParser::parseLiteralNick(KviIrcMessage *msg)
 			old->mergeQuery(q);
 			g_pMainWindow->closeWindow(q); // deleted path
 			if(!msg->haltOutput())
-					old->output(KVI_OUT_NICK,__tr2qs("\r!n\r%Q\r [%Q@\r!h\r%Q\r] is now known as \r!n\r%Q\r"),
+				old->output(KVI_OUT_NICK,__tr2qs("\r!n\r%Q\r [%Q@\r!h\r%Q\r] is now known as \r!n\r%Q\r"),
```
un1versal: @DarthGandalf this is also right, if its not then look at the rest of the file and if this is wrong then the rest just like these are also wrong

there was one to many tabs here

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
```diff
@@ -1787,11 +1787,13 @@ void KviIrcServerParser::parseLiteralNick(KviIrcMessage *msg)
...
 					&szNick,&szUser,&szHost,&szNewNick);
+
 			if(!_OUTPUT_MUTE)
-				old->output(KVI_OUT_SYSTEMWARNING,__tr2qs("End of merged output"));
-				old->userAction(szNewNick,szUser,szHost,KVI_USERACTION_NICK);
+			old->output(KVI_OUT_SYSTEMWARNING,__tr2qs("End of merged output"));
```
DarthGandalf: OTOH, this one has not enough tabs now.

un1versal: fixed, but generally speaking these files are fruit salad

un1versal:
> In the end, static analysis only hints, that something is probably wrong. It's human's job to triage its findings, and to determine which of them are actual bugs, and which are false positives.

Correct! hence I done these most _obvious_ ones.Tthough to be fair coverity scan is somewhat perplexing, there are similar issues that werent picked up or event hinted.
I see on our coverity page that there's some setup still to be done. Would that in any way improve anything that side, or is it irrelevant?

PR #1961, opened 2016-04-30
old clang format had a farm

un1versal:
This doesnt look right if these were meant to be spaces then all would be spaces right?

havent looked at other files but we should pick one way of doing these things and stick to it.
edit (more files have this mix of spaces vs tabs)

is this something clang format can handle or?

### DIFF
--- a/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
+++ b/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
@@ -96,8 +96,8 @@ void KviIrcServerParser::parseLiteralPing(KviIrcMessage * msg)
 	if((!msg->haltOutput()) && KVI_OPTION_BOOL(KviOption_boolShowPingPong))
 	{
 		msg->console()->output(KVI_OUT_SERVERPING,
-		    __tr2qs("Received ping from \r!s\r%Q\r (PING %Q), replied pong"),
-		    &szPrefix, &szAllParams);
+			__tr2qs("Received ping from \r!s\r%Q\r (PING %Q), replied pong"),
+			&szPrefix, &szAllParams);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


un1versal:
@DarthGandalf would it be possible to stick to tabs only or not use spaces at all on these type things?

Ill close this in any case but would be nice to fine tune these things so we dont end up with a soup again.

DarthGandalf:
That's intentional. It comes from `UseTab: ForIndentation`. There is a difference between indentation in `{}` block, and for non-first lines of function arguments to be righter.
Sometimes it's nice to align such arguments, and that's impossible to do with tabs:

``` c++
function_call(lorem, ipsum, dolor, sit,
              amet, consectetur, adipiscing,
              elit);
```

DarthGandalf:
> Ill close this in any case but would be nice to fine tune these things so we dont end up with a soup again.

The best would be for everyone to install clang-format, and instead of reformatting code manually, just press Ctrl-K in vim.
As fallback, it's possible to rerun it on the whole code from time to time, like I did, to cleanup whatever recent modifications became not formatted.

DarthGandalf:
> Sometimes it's nice to align such arguments,

Though I'm not sure if any function call became formatted like that, probably not. In that case spaces are not very useful...

pragmaware:
Spaces should never be used for indentation.
Moreover, mixing tabs and spaces will make the code look bad in someone's editor.

DarthGandalf:
@pragmaware it depends on how they are mixed :)

If there are defined rules when exactly it's spaces, and when exactly it's tabs, it can look good in every editor. As with lorem ipsum example above: that whole function can be idented using tabs in a while loop, so in a single line it's first tabs, then spaces.

Anyway, I'll prepare another PR.

pragmaware:
As soon as one uses a different tab size the alignment is lost.

DarthGandalf:
Nope.

``` c++
while (true) {
⟷function_call(lorem, ipsum, dolor, sit,
⟷              amet, consectetur, adipiscing,
⟷              elit);
}
```

(I replaced tab with ⟷)
Regardless of length of tab, they are still aligned, because of spaces between tabs and text.

DarthGandalf:
Also https://github.com/kvirc/KVIrc/blob/3114dbe7badb8eb65c21d1b20639b4784cdb7c43/src/kvirc/sparser/KviIrcServerParser_tables.cpp#L124 would look broken if to replace spaces with tabs, because it would be different number of tabs on different lines, and value won't be aligned anymore...

(it was using spaces even before I started touching clang-format)

un1versal:
> Spaces should never be used for indentation.
> Moreover, mixing tabs and spaces will make the code look bad in someone's editor.

:) my point exactly.

The main thing is it is also inconsistent the spaces used, most entries have 4 and some have 8 spaces and any of them are mixed with tabs...

imo would be preferable now that clang is doing the heavy lifting to stick to one absolute rather than mixing several inconsistencies.

thx @DarthGandalf will look at other PR, Im all in favor of this type of work.

PR #1962, opened 2016-04-30
Apply several fixes of clang-tidy

DarthGandalf:


un1versal:
Im sure @staticfox will love the nullptr changes, personally I dont understand a great deal of these improvements so cant comment on accuracy or causes. But I trust you especially know what youre doing.

Dont mean to nag but are the spaces + tabbing thing going to be revised and made into a non competing standard?

DarthGandalf:
They are not competing, when used this way...
If to have tabs _only_, size of tab will matter, and it will look ugly depending on editor settings.
I've tried that.

> personally I dont understand a great deal of these improvements

The loop-convert one is more interesting.

un1versal:
Sorry yes competing and very much so

e.g. https://github.com/kvirc/KVIrc/blob/master/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp#L1336-L1337

those have 15 spaces give or take a space...

https://github.com/kvirc/KVIrc/blob/master/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp#L1458

that has 4 spaces

and that's 2 random examples of what I call competing,

> If to have tabs only, size of tab will matter, and it will look ugly depending on editor settings.
> I've tried that.

If we say we need tabs to be 4 spaces or 8 spaces or whatever then thats the standard we all have to use and document and make sure thats what we expect from everyone.

I disagree with this soup, I dont know much, but this bad standard to follow.

then to add more wood into the pyre there's this type of block indenting
https://github.com/kvirc/KVIrc/blob/master/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp#L1740-L1742

which directly conflicts with thing of the same exact nature https://github.com/kvirc/KVIrc/blob/master/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp#L1780-L1782

I thought the point of clang etc. was to reduce spaghetti soup, to hopefully one standard only not add more.

un1versal:
These spaces are silly and inconsistent, the more I look at other files the more dislike this automated organized chaos. Not counting the tabs here is another example of inconsistency.

24 spaces https://github.com/kvirc/KVIrc/blob/master/src/modules/options/OptionsWidget_avatar.cpp#L78-L79

17 spaces https://github.com/kvirc/KVIrc/blob/master/src/modules/options/OptionsWidget_avatar.cpp#L80 and clang format decided to break `"options"));` to a whole new line

-1 for using spaces

DarthGandalf:
> 24 spaces https://github.com/kvirc/KVIrc/blob/master/src/modules/options/OptionsWidget_avatar.cpp#L78-L79

That's for alignment of continuation of the same string.

> clang format decided to break "options")); to a whole new line

It didn't change how arguments are splitted to lines. The `"options"));` was on separate line before that.
Clang-format could have changed that, but I tried to keep diffs at minimum.

DarthGandalf:
> 17 spaces

This one looks strange indeed, even though I think I understand how it happened.

un1versal:
> That's for alignment of continuation of the same string.

Its inconsistent with everything else and sadly you dont seem inclined to agree this is bad.

> It didn't change how arguments are splitted to lines. The "options")); was on separate line before that.
> Clang-format could have changed that, but I tried to keep diffs at minimum.

Not true, I fixed this file myself manually in https://github.com/kvirc/KVIrc/commit/83d5cf501df303e207d2830066193e1cbd1d988b

> This one looks strange indeed, even though I think I understand how it happened.

None match, there's absolutely no logic being followed, and you still haven't said if this is going to be made consistent or not, given you are driving this and its unlikely anyone else will bother with clang format.

DarthGandalf:
> there's absolutely no logic being followed

There is logic :) just it's not obvious, and confusing.

> said if this is going to be made consistent or not

I'll try various options.

> Not true

Hm, probably I remembered that from one of previous iterations of running it which were broken in other ways.

un1versal:
> There is logic :) just it's not obvious, and confusing.

then its failed :)

> I'll try various options.

thx much, and /me crosses fingers.

> Hm, probably I remembered that from one of previous iterations of running it which were broken in other ways.

:) yea, Ideally you need an empty line to separate tooltips

I was trying to do in https://github.com/kvirc/KVIrc/commit/83d5cf501df303e207d2830066193e1cbd1d988b was to make all tooltips in that exact format so as to make things like https://github.com/kvirc/KVIrc/blob/cde3fc69a17ba51e35f0e3ce365d10809cb94165/src/modules/options/OptionsWidget_alertsAndHighlighting.cpp#L124 consistent with one style only...

PR #1963, opened 2016-05-02
Fix text appearing outside the buttonboundary

Cizzle:
When using the classic window list, text on the buttons that is longer than the button itself, goes outside the right margin and buttonboundary. This PR fixes this.


wodim:
lgtm

PR #1970, opened 2016-05-06
Add a close button to the buttons in the classic windowlist.

Cizzle:
This adds a closebutton to the buttons used in the classic window list, resembling tabs in browsers.

### DIFF
--- a/src/modules/options/OptionsWidget_windowList.cpp
+++ b/src/modules/options/OptionsWidget_windowList.cpp
@@ -48,11 +48,12 @@ OptionsWidget_windowList::OptionsWidget_windowList(QWidget * parent)
 	                "options"));
 
 	addBoolSelector(0, 4, 0, 4, __tr2qs_ctx("Show IRC context indicator in window list", "options"), KviOption_boolUseWindowListIrcContextIndicator);
-	addBoolSelector(0, 5, 0, 5, __tr2qs_ctx("Enable window tooltips", "options"), KviOption_boolShowWindowListToolTips);
-	addBoolSelector(0, 6, 0, 6, __tr2qs_ctx("Allow the window list to be moved", "options"), KviOption_boolShowTreeWindowListHandle);
-	addBoolSelector(0, 7, 0, 7, __tr2qs_ctx("Show user flag for channels", "options"), KviOption_boolShowUserFlagForChannelsInWindowList);
+	addBoolSelector(0, 5, 0, 5, __tr2qs_ctx("Show close button on window list items", "options"), KviOption_boolUseWindowListCloseButton);
+	addBoolSelector(0, 6, 0, 6, __tr2qs_ctx("Enable window tooltips", "options"), KviOption_boolShowWindowListToolTips);
+	addBoolSelector(0, 7, 0, 7, __tr2qs_ctx("Allow the window list to be moved", "options"), KviOption_boolShowTreeWindowListHandle);
+	addBoolSelector(0, 8, 0, 8, __tr2qs_ctx("Show user flag for channels", "options"), KviOption_boolShowUserFlagForChannelsInWindowList);
 
-	addRowSpacer(0, 8, 0, 8);
+	addRowSpacer(0, 9, 0, 9);
 }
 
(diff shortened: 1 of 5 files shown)


un1versal:
I have to wonder if this wont make the button too busy, Icons + progress indicators + name + this, especially when the popups handle this perfectly well currently.

capture Right clik on button

capture1 right click on window

un1versal:
cap1

Too busy perhaps..., real estate here is at a premium, longer channel names that are faded in under lose out to this.

If you like to try to make this less clogged up by adding an option to the theme settings to make this optional, it may help your suggestion. + also

see  capture

minimum length button in themes to accommodate most cases is best between 150 to 160 pixels so aside from optional setting, in addition all themes need adjusting defaults to fit the minimum button width., **I can do this last one on the internal themes**

Cizzle:
This only takes one leftclick though to close, instead of a rightclick, move to correct contextmenuitem, leftclick. Or the SHIFT + leftclick which needs a hand on the keyboard. Or the keyboard-only CTRL+W.

It can get crowded, but there's the minimum width for? Enlarge the listbar and you get more space for the buttons.

On that third pic, it looks like you have a maximum width set?

un1versal:
> This only takes one leftclick though to close, instead of a rightclick, move to correct contextmenuitem, 

On the window right click itself is the top option. as opposed to the second  (right-click on menu), its not bad considering not long (two months) ago you had no such options at all. :)

> On that third pic, it looks like you have a maximum width set?

Yes, I set minimum and maximum...

> It can get crowded, but there's the minimum width for?

Minimum width doesnt work as prescribed when you have many many channels (and some KVIrc users do have dozens on one network) using this view. Its squashes them together and they become tiny. and really overcrowed in short, unusable for those users anyway.

>  Enlarge the listbar and you get more space for the buttons.

You run out sooner or later. Anyway the point is we have to make things like this optional
For this option. you should add a checkable boolean box to enable or disable this close button
disabled by default.

With this optional we can merge this asap and then Ill adjust the internal themes a bit.

Cizzle:
Added the option as requested. Added it to the main Window list settings as I'm working on adding a closebutton in the tree list as well.

un1versal:
Excellent :) will you add that to this PR?

Cizzle:
No, can't give an ETA for this as first attempts gave funny results. So decided to split this up PR-wise.

src/modules/options/OptionsWidget_windowList.cpp
```diff
@@ -48,11 +48,12 @@ OptionsWidget_windowList::OptionsWidget_windowList(QWidget * parent)
 	                "options"));
 
 	addBoolSelector(0, 4, 0, 4, __tr2qs_ctx("Show IRC context indicator in window list", "options"), KviOption_boolUseWindowListIrcContextIndicator);
-	addBoolSelector(0, 5, 0, 5, __tr2qs_ctx("Enable window tooltips", "options"), KviOption_boolShowWindowListToolTips);
-	addBoolSelector(0, 6, 0, 6, __tr2qs_ctx("Allow the window list to be moved", "options"), KviOption_boolShowTreeWindowListHandle);
-	addBoolSelector(0, 7, 0, 7, __tr2qs_ctx("Show user flag for channels", "options"), KviOption_boolShowUserFlagForChannelsInWindowList);
+	addBoolSelector(0, 5, 0, 5, __tr2qs_ctx("Show close button in window list", "options"), KviOption_boolUseWindowListCloseButton);
```
un1versal: :+1: I would maybe just be a little more explicit here with label wording to avoid confusion.

`Show close button in window list items` would make it obvious that this isnt for the window list itself.

Or some other wording to the same effect.

un1versal:
OK pending the label wording revision and some tests (latest tomorrow) I see no reason to hold this up.

Thx

un1versal:
works as intended so merging. thx for making it optional.

As for the close button on treeview window list items, one thing that pops to mind is the close button is black and with dark/black themes, the visibility maybe not ideal. sot so much of an issue when you select the item as the selection color tends to be lighter.

Or may not be a problem at all, just putting it out there.

PR #1974, opened 2016-05-08
:appveyor: openssl - just DO IT!

un1versal:
"DO IT, JUST DO IT!"


Dessa:
you are effectively caching what is already present.

un1versal:
I thought that everytime the file was changed the caching was nuked... so what to do? remove -> .appveyor.yml and merge then commit another change adding it?

un1versal:
@Dessa this seems not to work either, should be a simple way to bump these things.

PR #1976, opened 2016-05-09
kvioptions: reuse unused bools

un1versal:
**KviOptions.cpp** and **KviOptions.h** get cluttered over time and no one is reusing the unused options.

So this is my first part on fixing this. **tested and works** but want to test the windows options hence PR.

Im working on actually removing the really unused ones and adjust both files so it works both compile and runtime but this is my proof of concept to show theres nothing to fear.
#### Changes proposed
- Resuse unused bool entries

this is what we need to do from now on
and do away with these leftovers once and for all.


un1versal:
tested windows/linux and works

PR #1977, opened 2016-05-09
KviOptions: remove unused options + adjust numbering

un1versal:
See part 1 #1976 
Here However instead of reusing Ive bitten the bullet and went all out with cleaning this up.
Advantages are whoever adds something new ideally wont have to figure out how to reuse anything
#### Changes proposed
- remove unused options and adjust numbering etc 

Slightly tricky due to possible PEBCAKs but its possible.

now need to do same across the KviOptions.h/.cpp each one in order
always making sure each section compiles and is runtime tested and works before moving on to next.

Ill keep pushing updates to here as I go.

**Helpful feedback welcome.**

Tested runtime and compile in Linux seems OK


un1versal:
unrelated clang build failure

this seems to have gone better than I expected, tests come out OK with nothing major going on, both on Linux and Windows

The only minor glitch was I had to reapply theme and that seemed to fix up whatever darker bg on transfer window a toolbar applet, I didnt notice anything other than whats to be expected of dev builds. (famous last words)

Cant imagine this breaking anything so here goes nothing...... 

`un1versal pushes the merge button and runs for cover`

PR #1984, opened 2016-05-17
KviIrcToolBar: IRC Contextt Widget tooltip

un1versal:
#### Changes proposed
- Move Context counter to IRC Context Widget tooltip header
- Simplify contents of tooltip
- Make tooltip content presentation consistent with other similar tooltips

Todo:
    add connection in progress to tooltip

This works as intended when fully connected and looks good.
capture

capture

However, when connection is in progress things dont look so great and after many failed tries to add in progress that worked with non IRC context + disconnected, here is hoping to get some pointers on how to do this.


un1versal:
as expected...

PR #1994, opened 2016-06-02
KviMediaManager: add default file handlers Win, OS X + Linux

### LINKED ISSUE #1992: add default file handlers for Windows, OS X and most Linux
See https://github.com/kvirc/KVIrc/blob/fd8f4cd45463aef328c3a258e37ffccd0253e661/src/kvilib/ext/KviMediaManager.cpp#L377-L393

Whats is there is probably OK only for KDE

This isnt really a feature request because as is its a bug imo.

un1versal:
> ## **Notice**
> 
> These changes will only be useful to new installs or reset configs **since KVirc cannot handle any such changes properly** or to advanced users who know which files to overwrite.
> ### **Workaround**
> 
> 1) Close KVIrc
> 2) Delete **mediatypes.kvc** file  from config folder
> 3) Start KVIrc
> 4) Close KVIrc (magic happens a empty mediatypes.kvc is created)
> 5) Start KVIrc (more magic happens, the contents of the file are populated from new global values.
#### Changes proposed
- add minimal file handlers for supported OSs
- fix old typo in applicatoin -> application

This is at best a quick fix since nothing else but KDE was considered till now.

resolves #1992

~~This is incomplete at best, its also 100% untested on OS X and
doesnt account for executable files on any OS other than Windows
which is also partially untested until appveyor builds this~~

If anyone wants to add something or suggest something simple please do so in detail.

soon as some minimal tests are concluded this should be merged.

### DIFF
--- a/src/kvilib/ext/KviMediaManager.cpp
+++ b/src/kvilib/ext/KviMediaManager.cpp
@@ -374,23 +374,55 @@ KviMediaType * KviMediaManager::findMediaTypeForRegularFile(const char * szFullP
 	return mtd;
 }
 
-// FIXME : add default handlers for Windows, OS X and Non KDE??
+// FIXME : add proper default handlers for Windows
+// e.g. a KVS function which calls https://msdn.microsoft.com/en-us/library/windows/desktop/bb762153(v=vs.85).aspx
 
+#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
 static KviDefaultMediaType g_defMediaTypes[] = {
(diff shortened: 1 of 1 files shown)


un1versal:
Windows tests:

re-added zip to the Windows since the default command works on everything from XP to Win 7.

Things like tar/bz and bz2 etc have no business in Windows by default so thats gone from there.

OS X: Untested
has no handlers for executable run if someone wants it can add a default themselves.

Linux: These seems to work
has no handlers for executable run if someone wants it can add it themselves

> **Note:** tests were ran from cmd line only

PR #1998, opened 2016-06-05
 OptionsWidget_interfaceFeatures: remove broken options

### LINKED ISSUE #1978: Open Windows as toplevel
### Expected behavior

That there would be a difference in how windows open with toplevel or without

Could be a a Mdi remnant, but I suspect we could make these be modal vs not modal instead or something better.
### Actual behavior

There's no difference whatsoever
1. open capture
2. check and try to see what it does checked and what it does unchecked
   3.try running the commands via command line with -t or without

reference https://github.com/kvirc/KVIrc/blob/690dd7a33ddc90678367d73adf313533536e10f2/src/kvirc/kernel/KviInternalCommand.cpp#L54-L90
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Linux 3.13.0-85-generic
System version: #129-Ubuntu SMP Thu Mar 17 20:50:15 UTC 2016
Architecture: x86_64
Qt version: 5.5.1
Qt theme: fusion

Build Info:
Build date: 2016-05-09 21:56:46 UTC
Sources date: 20160509
Revision number: git-6614-g0658fc0-dirty
System name: Linux-3.13.0-85-generic
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler name: /usr/bin/c++
Compiler flags: N/A
Qt version: 5.5.1
Features: IRC, Debug, Crypt, SSL, IfAddr, IPC, OSS, Transparency, Webkit, Perl, Python, Enchant, Qt5, KVS
```

System:    Kernel 3.13.0-85-generic x86_64 (64 bit) Desktop Cinnamon 2.8.8  Distro Linux Mint 17.3 Rosa

un1versal:
#### Changes proposed
- Remove broken options from UI - Open Windows at toplevel
- Mark removed options Unused in KViOptions.h/cpp and add reuse notice

Pragma expressed that all these type things should open in dialogs only
and I agree see https://github.com/kvirc/KVIrc/issues/1978

these options have been broken and are unlikely ever to be fixed

This has been tested in LInux / Windows

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -59,6 +59,7 @@ KviRectOption g_rectOptionsTable[KVI_NUM_RECT_OPTIONS] = {
 
 #define BOOL_OPTION(_txt, _val, _flags) KviBoolOption(KVI_BOOL_OPTIONS_PREFIX _txt, _val, _flags)
 
+// NOTICE: REUSE EQUIVALENT UNUSED KviOption_bool in KviOptions.h ENTRIES BEFORE ADDING NEW ENTRIES BELOW
 KviBoolOption g_boolOptionsTable[KVI_NUM_BOOL_OPTIONS] = {
 	BOOL_OPTION("WindowsRememberProperties", true, KviOption_sectFlagWindows),
 	BOOL_OPTION("IrcViewShowImages", false, KviOption_sectFlagIrcView | KviOption_resetUpdateGui | KviOption_groupTheme),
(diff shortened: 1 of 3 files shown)


un1versal:
tested now on both Windows and Linux

wodim:
Wouldn't it be better to fix them?

Well, no, we have far too many options already.

Thinking out loud.

un1versal:
This is probably broken due to MDI removal and theres no benefit I can see to any of these options, good riddens I say,

Im just wondering what to do about https://github.com/kvirc/KVIrc/pull/2000 (related cleanup to these chnages) since that cleanup I identified a ton of unused stuff.

un1versal:
Im going to take more time trying to see if I have a better proposal for this and #2000

un1versal:
nope there are other things like manage addons, manage themes, toolbars editor and some others that work at toplevel and this convoluted way of opening inside kvirc) but to be honest I still dont see any advantage to any of this, where possible they should be dialogs like pragma says.

PR #1999, opened 2016-06-05
 KviMenuBar: help menu entries; have one help entry only

un1versal:
There is no good practical reason to have multiple entries on main menu especially when one entry is not searchable.
#### Changes proposed
-  Drop non searchable help options -- have one help entry only on MenuBar
-  make F1 entry also open an undocked window by default (like most apps use)
### Know issues

If closing help window from main window controls, at next reopen teh window size is shrunk to a square
**Hence WIP status and not merge recommendation is set.**


un1versal:
no longer opening undocked since clearly cant set minimum window sizes

so known issue is less visible

PR #2000, opened 2016-06-05
KviInternalCommand: clenup internal commands: windows toplevel etc

un1versal:
## **do not merge** until after https://github.com/kvirc/KVIrc/pull/1999 and https://github.com/kvirc/KVIrc/pull/1998 are in

There are many many unused options here, I removed 2 for testing side effects and see none
Suggestion is when https://github.com/kvirc/KVIrc/pull/1998 is merged the unused options should be considered for full removal.
#### Changes proposed
- Cleanup KviInternalCommand.h/cpp entries and mark unused
- Cleanup  Part 2 of https://github.com/kvirc/KVIrc/pull/1998

this cleans up the options to open at toplevel in internal commands.


un1versal:
Im wondering, if I should just remove all the unused options identified here, they dont show up anywhere in  source.

un1versal:
I added https://github.com/kvirc/KVIrc/pull/2000/commits/147b7f72454c3942a3675d71db915544a0333b34 but these look like unused since at least KVIrc 3.x

thers no reason to believe they will ever be used either

un1versal:
Ive tried reusing these unused ones into the coreactions but that requires more changes then Im prepared to do at this time.

So, Will just be the original cleanup and leaving the unused in kvirc for now as they are only internal code not exposed to users anywho.

At another time I may reuse them, else ripping them out is just as easy.

un1versal:
removed **dirbrowser.open** entries since this module no longer exists see https://github.com/kvirc/KVIrc/issues/2005

PR #2015, opened 2016-06-19
gpl: get rid of incompatible licences,

un1versal:
This always made me cringe somewhat, Im not a laywer, but according to GPL FAQ this cant be done since these licenses are incompatible.

We either upgrade to gplv3 or stay at gplv2(+) but both cant coexist.
#### Changes proposed
- get rid of gplv3 from source, gplv2 and gplv3 are strictly incompatible

If theres reasons for this or some special permission


un1versal:
So Qt is licensed under GPLv3

I will close this after having read https://www.gnu.org/licenses/gpl-faq.html.en#v2v3Compatibility once again and making note that KVIrc is GPLv2 or later.

However the GPLv2 on website should be revised and updated since the Qt exception no longer applies to GPLv3 Qt.
https://github.com/kvirc/kvirc-www/blob/ba751394dedebfc16d15acea54fef08e8a9e05fa/content/license.php#L303-L309

PR #2018, opened 2016-06-21
KviIrcView_loghandling: add spaces to logging after timestamps

un1versal:
SM0TVI requested this on IRC, its here for SM0TVI to test and confirm this works for all timestamps and user action msg type etc...
#### Changes proposed
- Add spaces after timestamp in log events

### DIFF
--- a/src/kvirc/ui/KviIrcView_loghandling.cpp
+++ b/src/kvirc/ui/KviIrcView_loghandling.cpp
@@ -56,12 +56,15 @@ void KviIrcView::stopLogging()
 				// this is the equivalent to an empty date.toString() call, but it's needed
 				// to ensure qt4 will use the default() locale and not the system() one
 				szDate = QLocale().toString(date, "ddd MMM d hh:mm:ss yyyy");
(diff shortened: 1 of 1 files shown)


SM0TVI:
By request, uploading "before" views of the log view in ISO8601 (10:06:59 local), Classic (10:41:50 local) and System Default (10:43 local)mode (the screenshots were acquired in that order). Log and window lists censored a bit, otherwise the screenshots are as they appeared on my Win64 box.

Just looking at the screenshots, it appears it is only the ISO8601 version which lacks the spacer between the time stamp and message text. 

kvirc-logview-win64-classic-before
kvirc-logview-win64-iso8601-before
kvirc-logview-win64-sysdflt-before

SM0TVI:
"After" view screenshot.
Message at 10:58 local is the system default log view
Message at 11:00 local is the classic log view
Message at 11:01 local is the ISO8601 log view (which I use by default).
kvirc-logview-win64-iso8601-after
Appears FIXD.
_congrats un1versal_

PR #2029, opened 2016-06-27
KviOptions: make San Serif the font familly in Linux

un1versal:
#### Changes proposed
- The fonts for various parts all ~~monospace~~ San Serif in Linux

Windows defines Arial for all, OS X uses Lucida Grande, and this seems logical and works better to get things started.

> #### Worthy note:

These fonts are only used if by **at first run** user decides not to apply any themes

and overridden by when any themes are applied.

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -667,15 +667,15 @@ KviFontOption g_fontOptionsTable[KVI_NUM_FONT_OPTIONS] = {
 	FONT_OPTION("Notifier", "Lucida Grande", 12, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier),
 	FONT_OPTION("NotifierTitle", "Lucida Grande", 12, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier)
 #else
-	FONT_OPTION("IrcView", "Monospace", 10, KviOption_sectFlagIrcView | KviOption_resetUpdateGui),
-	FONT_OPTION("Input", "Sans Serif", 12, KviOption_sectFlagInput | KviOption_resetUpdateGui),
-	FONT_OPTION("UserListView", "Sans Serif", 10, KviOption_sectFlagUserListView | KviOption_resetUpdateGui),
-	FONT_OPTION("Label", "Sans Serif", 10, KviOption_sectFlagLabel | KviOption_resetUpdateGui),
-	FONT_OPTION("Application", "Sans Serif", 10, KviOption_sectFlagGui | KviOption_resetUpdateAppFont),
-	FONT_OPTION("WindowList", "Sans Serif", 10, KviOption_sectFlagWindowList | KviOption_resetUpdateWindowList),
-	FONT_OPTION("TreeWindowList", "Sans Serif", 10, KviOption_sectFlagWindowList | KviOption_resetUpdateWindowList),
-	FONT_OPTION("Notifier", "Sans Serif", 9, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier),
-	FONT_OPTION("NotifierTitle", "Sans Serif", 9, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier)
+	FONT_OPTION("IrcView", "sans-serif", 10, KviOption_sectFlagIrcView | KviOption_resetUpdateGui),
+	FONT_OPTION("Input", "sans-serif", 10, KviOption_sectFlagInput | KviOption_resetUpdateGui),
+	FONT_OPTION("UserListView", "sans-serif", 10, KviOption_sectFlagUserListView | KviOption_resetUpdateGui),
+	FONT_OPTION("Label", "sans-serif", 10, KviOption_sectFlagLabel | KviOption_resetUpdateGui),
+	FONT_OPTION("Application", "sans-serif", 10, KviOption_sectFlagGui | KviOption_resetUpdateAppFont),
+	FONT_OPTION("WindowList", "sans-serif", 10, KviOption_sectFlagWindowList | KviOption_resetUpdateWindowList),
+	FONT_OPTION("TreeWindowList", "sans-serif", 10, KviOption_sectFlagWindowList | KviOption_resetUpdateWindowList),
+	FONT_OPTION("Notifier", "sans-serif", 10, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier),
+	FONT_OPTION("NotifierTitle", "sans-serif", 10, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier)
 #endif
 };
 


CyberShadow:
Why monospace? If the default is variable-width fonts on one platform, it should be the same on all platforms. Maybe you meant `sans-serif`?

HelLViS69:
On Linux there is Dejavu family

CyberShadow:
On Linux you can specify "sans-serif" or "monospace" and it will use whatever font the user configured to be by default, like in CSS. I think this is what programs should be doing instead of making a guess at what fonts might be pre-installed.

> On Linux there is Dejavu family

Most distributions have it by default, but on some you have to install it explicitly (e.g. Arch Linux).

wodim:
I think `monospace` (lowercase) is the best idea. The default is configured, systemwide, using fontconfig.

wodim:
Anyway this is overwritten when a theme is applied during setup. The user can choose not to apply a theme but I doubt many users do that. The default is, in fact, pretty awful. Maybe that choice should be reviewed or removed.

CyberShadow:
Yes but why `monospace` (which is fixed-width) if the default on other platforms is variable-width? Wouldn't `sans-serif` be a better fit?

wodim:
yes

un1versal:
> Anyway this is overwritten when a theme is applied during setup. The user can choose not to apply a theme but I doubt many users do that. The default is, in fact, pretty awful. Maybe that choice should be reviewed or removed.

Why cant we use fixed width for all?

Visually alignment of all things is generally better

Removing default is not really a good idea, since you can also define not to have any theme applied which then uses these defaults + default colors etc.

I dont mid assigning variable widths but they are awful imo.

> On Linux you can specify "sans-serif" or "monospace" and it will use whatever font the user configured to be by default, like in CSS. I think this is what programs should be doing instead of making a guess at what fonts might be pre-installed.
> 
> ```
> On Linux there is Dejavu family
> ```
> 
> Most distributions have it by default, but on some you have to install it explicitly (e.g. Arch Linux).

I discussed this a while back with Pragma, and even suggested themes carry own fonts so they look the same on all OS the way theme builders intended.

so you guys want crappy variable widths (for all OS) instead so UI and all IRC things are all over then place right?

CyberShadow:
> Visually alignment of all things is generally better
> 
> so you guys want crappy variable widths (for all OS) instead so UI and all IRC things are all over then place right?

If something is not aligned correctly with variable-width fonts, then it's a bug.

Mildly related: http://nickgravgaard.com/elastic-tabstops/

> I dont mid assigning variable widths but they are awful imo.

I've talked with people about their font preferences for IRC, it seems to be highly dependent on personal preference.

> I discussed this a while back with Pragma, and even suggested themes carry own fonts so they look the same on all OS the way theme builders intended.

I don't think this is a good idea:
1. Most fonts are copyrighted. Redistributing them implicitly with KVirc themes facilitates copyright violations.
2. The same font file can look great or awful depending on what OS/library renders it.
3. Edit: There's also the issue of formats (TTF/OTF/etc.) and compatibility.
4. Edit: Any non-standard fonts are likely to have poor Unicode coverage, which is not un-important for IRC.

un1versal:
> Most fonts are copyrighted. Redistributing them implicitly with KVirc themes facilitates copyright violations.
> The same font file can look great or awful depending on what OS/library renders it.

Welcome to 2016 and Opensource! There are **many** GPL or MIT and other GPL License compatible fonts outthere, many right here on GitHub.

> If something is not aligned correctly with variable-width fonts, then it's a bug

Yes, welcome to KVIrc, where bugs are more important than features :) `</sarcasm>`

> I've talked with people about their font preferences for IRC, it seems to be highly dependent on personal preference.

Yes, true, but these are default fonts that only applied when and if a user decides not to apply any theme by default at first KVIrc setup...

The two main themes both offer variable IIRC. and if you use your own themes then it overrides these + if you dont like it you can always change which works both ways.

I only suggested fixed width due to the **now revealed bug** that all alignment in UI and else where is lost.

so OK , variable with it is and bugs be dammed :-}

CyberShadow:
> Welcome to 2016 and Opensource! There are many GPL or MIT and other GPL License compatible fonts outthere, many right here on GitHub.

So will KVirc check the license of the font before including it in the theme file? Please tell me how you intend to implement that :)

Just because there are many libre fonts, doesn't change that most fonts than a user is likely to use / stumble upon will be non-free.

Edit: How would this work anyway? Would KVirc pull the font file from the OS font directory, or ask the user to attach a font file manually?

src/kvirc/kernel/KviOptions.cpp
```diff
@@ -667,15 +667,15 @@ KviFontOption g_fontOptionsTable[KVI_NUM_FONT_OPTIONS] = {
 	FONT_OPTION("Notifier", "Lucida Grande", 12, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier),
 	FONT_OPTION("NotifierTitle", "Lucida Grande", 12, KviOption_sectFlagNotifier | KviOption_resetUpdateNotifier)
 #else
-	FONT_OPTION("IrcView", "Monospace", 10, KviOption_sectFlagIrcView | KviOption_resetUpdateGui),
-	FONT_OPTION("Input", "Sans Serif", 12, KviOption_sectFlagInput | KviOption_resetUpdateGui),
+	FONT_OPTION("IrcView", "Sans Serif", 10, KviOption_sectFlagIrcView | KviOption_resetUpdateGui),
```
CyberShadow: It's `sans-serif`, lower case and through a dash.

un1versal: sorry I was only using whats in there before, gimme a sec.

un1versal:
> So will KVirc check the license of the font before including it in the theme file? Please tell me how you intend to implement that :)

We only care if themes are published in our repo/website and in that case, thats what reviews are for. I would have thought it was obvious a programmatic solution is not possible,

We already have that issue with theme icons and personally speaking, all we can do is if there are any complaints due to something slipping the review net, deal with such complaints diligently should they ever arise. In meanwhile the themes readme.md covers this, something I though wise to implement after it was approved by pragma https://github.com/kvirc/kvirc-themes#themes-and-icon--artwork-licensing

CyberShadow:
Fair enough. There's still serious technical issues with bundling fonts. Even for web pages, you generally use stylesheet generators which link to multiple formats.

un1versal:
> Fair enough. There's still serious technical issues with bundling fonts. Even for web pages, you generally use stylesheet generators which link to multiple formats.

Yes thats a huge problem currently as you are well aware. Sadly.

src/kvirc/kernel/KviOptions.cpp
un1versal: fixed in https://github.com/kvirc/KVIrc/pull/2029/commits/fd7803b455ffcbb79221dd4899512d96b129a899

wodim:
> Why cant we use fixed width for all?

I don't see a reason why a chat program should use a monospaced font by default.

> Fair enough. There's still serious technical issues with bundling fonts. Even for web pages, you generally use stylesheet generators which link to multiple formats.

That's funny. The Aria theme for example comes with a DejaVu monospaced font by default which won't work anywhere else, so in Windows for example the default font is not monospaced.

CyberShadow:
> That's funny. The Aria theme for example comes with a DejaVu monospaced font by default which won't work anywhere else, so in Windows for example the default font is not monospaced.

Perhaps a better solution is to allow themes to specify a list of fonts to fall back on, like CSS. E.g.: "DejaVu Sans Mono, Menlo, Consolas, Lucida Console, Courier New, monospace"

un1versal:
> Perhaps a better solution is to allow themes to specify a list of fonts to fall back on, like CSS. E.g.: "DejaVu Sans Mono, Menlo, Consolas, Lucida Console, Courier New, monospace"

That sounds reasonable, but that brings me back to shipping a selection of fonts with KVIrc (not themes this time)

Some IRC clients dont seem to have this issue according to Pragma, they look very similar from one OS to the Other (so they must be doing something right)

so... as for this PR, its good now? fell free to merge.

wodim:
> That sounds reasonable, but that brings me back to shipping a selection of fonts with KVIrc (not themes this time)

Unnecessary, if it's a list of fonts, just make sure in that list you have fonts that are available to all three OSes

un1versal:
comic sans anyone? damm, its not in all OS's

Issue #2017, opened 2016-06-21
Setting output font as Italic will not work


un1versal:
Reported by DarthGandalf over in IRC, confirmed by myself after a few tests
### Expected behavior

That any output font is set to italic in output when selected.
### Actual behavior

Font stays with its normal glyph type
### Steps to reproduce the issue
1. Open Theme Editor 
2. Select output
3. Click the font button and select Italic
4. Click Apply
### Other things tried

Saving a Theme with all italic fonts everywhere and restarting Kvirc and even re-applying theme.
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Linux 3.13.0-88-generic
System version: #135-Ubuntu SMP Wed Jun 8 21:10:42 UTC 2016
Architecture: x86_64
Qt version: 5.5.1
Qt theme: fusion

Build Info:
Build date: 2016-06-20 17:52:43 UTC
Sources date: 20160620
Revision number: git-6745-g0377245
System name: Linux-3.13.0-88-generic
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler name: /usr/bin/c++
Compiler flags: N/A
Qt version: 5.5.1
Features: IRC, Debug, Crypt, SSL, IfAddr, IPC, OSS, Transparency, Webkit, Perl, Python, Enchant, Qt5, KVS
```

AndrioCelos:
What should happen, then, if the font is set to italic and italic codes are used in the message? You can't make it doubly italic...

un1versal:
No change then, I would imagine.

AndrioCelos:
That's what HexChat does... it's at odds with what I see in literature though.

un1versal:
eh...

wodim:
What I've seen in literature is that two italics cancel each other :camel: 

Camel unrelated

PR #2031, opened 2016-06-27
Respect bold and italic font settings. Fixes kvirc/KVIrc#2017.

AndrioCelos:
#### Changes proposed

This commit changes the way italic codes are handled to respect an italic font setting. If the normal font is italic, italic codes will make the text normal style, which is the formatting style seen in literature when emphasis occurs in a quote which is written in italic.
Bold and underline codes stack with font settings.
### Known issues
- The XOR behaviour of italic codes and stacking of bold codes don't seem to work as intended in the topic widget.
- Italic codes may fail to render if you change to an italic font, until KVIrc is restarted.

### DIFF
--- a/src/kvirc/ui/KviInputEditor.cpp
+++ b/src/kvirc/ui/KviInputEditor.cpp
@@ -853,6 +853,8 @@ void KviInputEditor::drawContents(QPainter * p)
 
 	int iTextBaseline = iBottom - fm->descent();
 
+	QFont::Style normalFontStyle = p->font().style();
+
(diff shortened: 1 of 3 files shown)


un1versal:
Works well here, despite your known issues notes, not at all surprised given the issues previously present.

Lets see what DarthGandalf says.

un1versal:
some positive feedback and community/dev engagement are always desirable... but since that fails we merge it and opened https://github.com/kvirc/KVIrc/issues/2033

Issue #1950, opened 2016-04-20
Feature Request: line numbers in script editors


un1versal:
That title should be enough.

But everywhere where you can script into should have optional line numbers, default on if possible.

Main reason is that seems somewhat silly a KVS error point you to a line number, you code has a few dozen lines or hundred lines or more and your alternative is use a likely edit this in a external editor to get to it easier

AndrioCelos:
There's an area on the status bar that tells you what line the cursor is on (though that's also inconsistent as those numbers are zero-based, and should possibly be changed too).

un1versal:
yea tbh until you said it (numbers on the counter), really just missed it entirely, goes to show how non intuitive that is.
Ones expects line numbers in the editor area type thing.

un1versal:
> There's an area on the status bar that tells you what line the cursor is on (though that's also inconsistent as those numbers are zero-based, and should possibly be changed too).

When the script error/warning pops up and says e.g error in line  **line 1** in this counter its **Row 0**
Like you say, inconsistent for sure....

```
[KVS]   In script context "OnJoin::default", line 1, near character 1
```

So this is already misleading, We can rename the Row: to Line: but the 0 based stuff vs error needs to also be fixed like you said

PR #2040, opened 2016-07-02
Row/Col: Rename row to line and dont use 0 base for first line/col

un1versal:
This works fine @ Try 2

Noticed something weird going on with Event Editor, probably unrelated.
Open event editor and you can type into the editor area and even type into the event name field
The problem is there's no actual event selected so this should be possible :/

Reference https://github.com/kvirc/KVIrc/issues/1950 the feature request uncovered an issue

There's a inconsistency between row/col numbers in reference to errors printed and names used
Row and Column counter position starts at 0 while if there's an error it references line/col starts at 1
Also The errors mention lines not rows

e.g.

```
[02:33:46] [KVS] Compilation error: Unexpected end of line in string constant (missing " character or unescaped newline)
[02:33:46] [KVS]   In script context "sys", line 163, near character 22
[02:33:46] [KVS] Code listing:
[02:33:46] [KVS]   161 ...
[02:33:46] [KVS]   162  {
[02:33:46] [KVS]   163      echo $tr("Usage:")"
[02:33:46] [KVS]   164      echo "      "/%aliasname <cpu> %S $tr("Output CPU information.")
[02:33:46] [KVS]   165 ...
```

capture

However when you opened the script and goto line 8 you were actually in line 9 since counter is 0 based.

~~The counter also actually showed Row, not line as the error suggests.~~
~~so we try to make this consistent by using more common place terminology.~~
#### Changes proposed
-  ~~Rename Row -> Line on counter~~ fixed
-  Make Line and Column count start at 1

Let AV build and maybe with some luck get some feedback

### DIFF
--- a/src/modules/editor/ScriptEditorImplementation.cpp
+++ b/src/modules/editor/ScriptEditorImplementation.cpp
@@ -762,7 +762,7 @@ ScriptEditorImplementation::ScriptEditorImplementation(QWidget * par)
 	pLab->setAlignment(Qt::AlignRight | Qt::AlignVCenter);
 	g->addWidget(pLab, 1, 1);
 
-	m_pRowColLabel = new QLabel(QString(__tr2qs_ctx("Line: %1 Col: %2", "editor")).arg(0).arg(0), this);
+	m_pRowColLabel = new QLabel(QString(__tr2qs_ctx("Line: %1 Col: %2", "editor")).arg(1).arg(1), this);
 	m_pRowColLabel->setFrameStyle(QFrame::Sunken | QFrame::Panel);
 	m_pRowColLabel->setMinimumWidth(80);
 	g->addWidget(m_pRowColLabel, 1, 3);
@@ -937,8 +937,8 @@ void ScriptEditorImplementation::updateRowColLabel()
 {
 	if(m_lastCursorPos == m_pEditor->textCursor().position())
 		return;
-	int iRow = m_pEditor->textCursor().blockNumber();
-	int iCol = m_pEditor->textCursor().columnNumber();
+	int iRow = m_pEditor->textCursor().blockNumber() + 1; 
+	int iCol = m_pEditor->textCursor().columnNumber() + 1;
 	QString szTmp = QString(__tr2qs_ctx("Line: %1 Col: %2", "editor")).arg(iRow).arg(iCol);
 	m_pRowColLabel->setText(szTmp);
 	m_lastCursorPos = m_pEditor->textCursor().position();


wodim:
Are the changes in KvsObject_multiLineEdit.cpp necessary? They will break existing scripts and, as far as I guess, they should not be user-facing

un1versal:
yes thats broken as well as counter is wrong as well...

also 

just .arg(1) wont work,

to be more specific, the row and column numbering goes off to 0 in cases e.g. say in multiline editor and the others it also behave wrong...

this works. or seems to work...

Im not sure if the multiline.cpp changes are necessary but Ill check without as well

How will this break scripts? Can you be more specific pls.

wodim:
Read the docs for _multiline class_:

> <array> $cursorPosition()
> 
> Returns current cursor position in "line:uinteger, col:uinteger" format. Both numbers start counting from zero.

un1versal:
Well fix docs then and change that, 0 is not what error messages say.

The alternative being for the error messages to be 0 based just like nothing else does so at least its wrong all the way.

wodim:
You don't understand what I mean, that's fine, but don't break the multiline class, thank you.

un1versal:
I dont think its fine at all that I dont understand what you mean.. Explain it to me better or some other way... Im trying to make sense of it and as is I cant since nothing I know of with lines/columns counts from 0 editor wise.

wodim:
The "multiline class" is a KVS class for scripting. The "$cursorPosition()" function returns the cursor position for a "multiline class". Scripts that rely on that function will break if you arbitrarily sum 1 to the return values.

un1versal:
OK I reverted `KvsObject_multiLineEdit.cpp`

I didnt realize the "multiline class" has nothing to do with these fixes and that leaving this part wont cause the same issue in this "multiline class"

I tested without the `KvsObject_multiLineEdit.cpp` and it works UI side both on editors and multiline input now it counts from 1 so if errors and warning reference a line you actually check the correct line.

un1versal:
OK some more testing and I found a small discrepancy between error and warning on the column side.

made it fail on purpose.

```
[00:41:26] [KVS] Warning: Unterminated string constant
[00:41:26] [KVS]   In script context "sys", line 171, near character 71
[00:41:26] [KVS] Code listing:
[00:41:26] [KVS]   169 ...
[00:41:26] [KVS]   170      echo "      "/%aliasname <hdd> %S $tr("Output HDD/SSD information.")
[00:41:26] [KVS]   171      echo "      "/%aliasname <sen> %S $tr("Output Sensors information.")"
[00:41:26] [KVS]   172      echo "      "/%aliasname <share> <nick> %S $tr("Outputs a link to share %scriptname with a specific KVIrc user.")

[00:41:26] [KVS] Compilation error: Unexpected end of line in string constant (missing " character or unescaped newline)
[00:41:26] [KVS]   In script context "sys", line 171, near character 72
[00:41:26] [KVS] Code listing:
[00:41:26] [KVS]   169 ...
[00:41:26] [KVS]   170      echo "      "/%aliasname <hdd> %S $tr("Output HDD/SSD information.")
[00:41:26] [KVS]   171      echo "      "/%aliasname <sen> %S $tr("Output Sensors information.")"
[00:41:26] [KVS]   172      echo "      "/%aliasname <share> <nick> %S $tr("Outputs a link to share %scriptname with a specific KVIrc user.")
```

**Warning points to** 

[00:41:26] [KVS]   In script context "sys", **line 171**, near character **71**

**Error points to:**
[00:41:26] [KVS]   In script context "sys", **line 171**, near character **72**

Character 72 is the one after **"** while character 71 is exactly n the character

I cant remember is the desired pointer **near** or the **actual** character. (aside from what warning suggests?

this is the code for those bits
https://github.com/kvirc/KVIrc/blob/5f0292c32b336be0bb279a7511a9ee37ba9df3fa/src/kvirc/kvs/KviKvsRunTimeContext.cpp#L120
https://github.com/kvirc/KVIrc/blob/5f0292c32b336be0bb279a7511a9ee37ba9df3fa/src/kvirc/kvs/KviKvsRunTimeContext.cpp#L139
https://github.com/kvirc/KVIrc/blob/4d188c1afce5f87c4d3c8b7d0b44695324759b10/src/kvirc/kvs/parser/KviKvsParser.cpp#L96

#### **NOTE** but report at https://github.com/kvirc/KVIrc/issues/2041

un1versal:
I removed the cosmetic changes from here since there is a discrepancy with these fixes + warnings vs error column lines see https://github.com/kvirc/KVIrc/pull/2040#issuecomment-230368209

I dont think that issue lies in `ScriptEditorImplementation` and I doubt it can be fixed in this file and s I been told not to break whats already broken, so I'll leave it to the experts.

Ill open tickets with information and bury my head in the sand.

staticfox:
added in https://github.com/kvirc/KVIrc/commit/c67b91eb53f88d79daaed98765634bd3758c209a

un1versal:
> added in c67b91e

Yea github doesnt like it when PRs arent in sync with master and wont close PR's on merge via command line even if theres no conflicts...

This was reported to GitHub  2 years ago, and still unfixed.

Only mentioned that to say that it makes me feel better that (KVIrc) bug fixing is awful, but we just dont have the monkeys for it, while GitHub do have the monkeys and are paid to probably be on facebook/twitter all day. (or insert whatever site it is they actually spend time working hard on hardly working)

PR #2043, opened 2016-07-07
Use real bold when it is safe to do so

### LINKED ISSUE #2022: Ugly rendering of bold text when using bitmap X core fonts
1. On Linux, open Theme Preferences, and in Output -> Look -> Font, set the font to "Fixed [Misc]", "Regular" style, with size 10.
2. Send some bold text (with Ctrl+B) to a channel.
### Expected behavior

KVirc should be using the bold variant of the font.
### Actual behavior

The font KVirc uses appears to be an algorithmically "boldened" version of the regular style, with each glyph rendered twice with a 1-pixel horizontal offset.

This causes bold text to have a much lower readability (many letters are conjoined, and glyphs such as "m" and "w" become difficult to recognize.

Example screenshot - above is correct, below is KVirc rendering:

Worth noting that the problem does not manifest in Quassel, also a Qt application, which leads me to believe that it's KVirc-specific (as opposed to being a problem in Qt or TrueType).
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Linux 4.6.2-1-ARCH
System version: #1 SMP PREEMPT Wed Jun 8 08:40:59 CEST 2016
Architecture: x86_64
Qt version: 5.6.1
Qt theme: fusion

Build Info:
Build date: 2016-06-24 14:41:56 UTC
Sources date: 20160624
Revision number: git-6749-g08a66ad
System name: Linux-4.6.2-1-ARCH
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=
   Threads=POSIX
Compiler name: /usr/bin/c++
Compiler flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong 
Qt version: 5.6.1
Features: IRC, IPv6, Crypt, SSL, IfAddr, IPC, KDE, OSS, Audiofile, Transparency, Phonon, Webkit, DCCVoice, Perl, Python, Enchant, Qt5, KVS
```

CyberShadow:
Revive b612e36ce2581fdf1f0df5e4bee99bcc70cc548c, but only enable it when
the metrics of printable ASCII characters match between bold and non-bold.

Fixes #2022.

CC @wodim, @Dessa

### DIFF
--- a/src/kvirc/ui/KviIrcView.cpp
+++ b/src/kvirc/ui/KviIrcView.cpp
@@ -1072,7 +1072,7 @@ void KviIrcView::paintEvent(QPaintEvent * p)
 		// note that QFontMetrics(pa.font()) may be not the same as QFontMetrics(font())
 		// because the painter might effectively use an approximation of the QFont specified
 		// by font().
-		recalcFontVariables(QFontMetrics(pa.font()), pa.fontInfo());
+		recalcFontVariables(pa.font(), pa.fontInfo());
 	}
 
 #ifdef COMPILE_PSEUDO_TRANSPARENCY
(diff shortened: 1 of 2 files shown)


Dessa:
+1

un1versal:
Well thats all we needed, thx Dessa.

in you go

PR #2045, opened 2016-07-07
Improve text selection accuracy

CyberShadow:
Make the horizontal point delimiting which characters are to be selected
be in the middle of the character, i.e. if the user starts selecting
from the left half of a character, select that character too.

This mimics the behavior of most other GUI applications.

### DIFF
--- a/src/kvirc/ui/KviIrcView.cpp
+++ b/src/kvirc/ui/KviIrcView.cpp
@@ -2612,15 +2612,25 @@ int KviIrcView::getVisibleCharIndexAt(KviIrcViewLine *, int xPos, int yPos)
 						return l->szText.size();
 				}
 				// now, get the right character inside the block
-				int retValue = 0;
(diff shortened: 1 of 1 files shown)


un1versal:
good stuff...  tests here give it a green light

lets see what others say or do with these pulls, but from me its :+1:

un1versal:
the only thing I notice is the direction of selection.

This is OK when selection is being dragged right, when selection is dragged left (from the left half of a character) that character was and continues to be ignored... I suppose this is intentional?

CyberShadow:
The direction of the selection shouldn't matter, this affects translation of coordinates into character offsets (so affects both selection start/end/left/right equally). The change might be subtle because the actual character coordinates seem to shifted a bit compared to how they are rendered. It's easier to compare side-by-side or with some debugging printfs.

un1versal:
if side by side you mean a build with and a build without the patch, thats what I observe

I guess the issue probably lies elsewhere and not with patch, KVirc custom input is a pinata of surprises.

PR #2046, opened 2016-07-07
Add Ctrl+Home/End to scroll to top/bottom

CyberShadow:
Consistent with other apps, where Ctrl+Home/End jumps to the top/bottom
of the document.

This is mainly useful to quickly jump to the very bottom of a buffer
after a search, as well as UX parity with other clients.


un1versal:
does what it says on tin on Linux and Win so like the others in it goes

Issue #2019, opened 2016-06-24
Scrolling Channel Treelist skips one channel per scroll tick


gizmore:
### Expected behavior

Using the mouse wheel on channel tree list should move one channel/context per scroll tick.
### Actual behavior

Since a few weeks it is skipping one channel/context, marking the context read but skipping it.
### Steps to reproduce the issue
1. Have more than 2 channels
2. Use the mouse wheel to scroll through the channels
3. One channel is skipped per mouse wheel tick, moving 2 channels at once.
### System information

net-irc/kvirc-5.0_pre20160401144043::gentoo
KVIrc 4.9.2 'Aria'

Laufzeitinformationen:
System name: Linux 4.6.2-gentoo-gizmo-462e
System version: #2 SMP PREEMPT Fri Jun 24 02:15:40 CEST 2016
Architektur: x86_64
Qt version: 5.6.1
Qt theme: fusion

Erstellungsinfo:
Build date: 2016-05-28 00:56:57 UTC
Sources date: 20160102
Revision number: a559c50d8ff1ba2c265d22cb45ce0b79b26584d9
System name: Linux-4.5.2-gentoo-gizmo-452b
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   VERBOSE
   LIB_SUFFIX=64
   Threads=POSIX
Compiler name: /usr/bin/x86_64-pc-linux-gnu-g++
Compiler flags: -O2 -pipe -march=native -fomit-frame-pointer -fno-strict-aliasing 
Qt version: 5.6.0
Features: IRC, IPv6, Crypt, SSL, IfAddr, IPC, Transparency, Phonon, DCCVideo, DCCVoice, Perl, Python, Qt5, KVS

```
Thanks for kvirc!
```

un1versal:
Its not mouse related, Im able to duplicate with `Ctrl+Tab` or `Ctrl+Shift+Tab``

gizmore:
Fun Fact: Ctrl+Tab works correct here, one channel per tab

un1versal:
Not if you press and hold here, eventually it goes nuts

alisonken1:
Hmm - Having this issue on Fedora 24 as well as of 2016-08-08
KVirc 4.9.1 'Aria'
Fedora 24 x86_64 (fresh install)

Maybe the fedora 24 repo hasn't gotten the fix yet?

[...]

Issue #1301, opened 2015-07-26
WIndows List Scrolling


DarthGandalf:
**Reported by Cronus on 05 Jul 2012 01:31:04 UTC**
Pertaining to the setting boolWheelScrollsWindowsList Which either when using the mouse wheel you scroll the active window list, or the scrollbar is the one scrolling.  If you do these specific steps it will spazz out and try to scroll the active window and scrollbar at the same time..

I am running windows 7 64bit FYI

Minimize KVIRC
Go to another application (by clicking on it and bringing it into the active window
Now Click on kvirc to maximize it again and try to use your mouse wheel in the windows list on the left.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1301

DarthGandalf:
**Comment by ctrlaltca on 16 Jul 2012 21:45:34**
You forgot to mention the actual value of the boolWheelScrollsWindowsList option you're using. Anyway i tried to perform the steps you described with both that option turned on or off:
- when it's off, only the scrollbar scrolls;
- when it's on, the active window in the window list changes; when the last visible item in the window list gets selected, but there are some others more after it, the scrollbar scrolls to make the very next item visible so that it can gets selected and become the active window. This is the intended behaviour to ensure that the currently selected window list item is visible.<br/>

Is this what you mean or did i miss the point?

DarthGandalf:
**Comment by Cronus on 16 Jul 2012 23:04:00**
I have it turned off. And no that is not I mean. But like I said it basically tries to do BOTH actions at the same time (Change the active window AND move the scrollbar) I showed someone from the freenode support channel. I can't remember who it was, it might have been you? I showed them over join.me session and they said to make this ticket and maybe someone on windows can assist.

un1versal:
cannot duplicate this currently, nothing unusual about this as is, imo this unlikely scenario only a edge case user sees and also its on a outdated buggy build.

un1versal:
I can actually dup this after a couple of tries... its odd after doing this a channel is both selected and you can have some scrolling selection at same time. looks and behaves oddly and looks like arse too.

But like everything, this in particular is a very edge case scenario you wouldn't normally see or experience unless trying to find this issue.

un1versal:
> Arfrever commented on 3 Mar 

If I press long "rectangle" used to scroll list of channels and move cursor up or down AND to the left so that it points to name of any channel and if I stop pressing mouse button in this moment, then name of that channel will become semi-highlighted.

[...]

Issue #1940, opened 2016-04-16
Tree window list scrolling dies on OS X 


Cpuroast:
### Expected behavior

Tree window list scrolling on OS X works when "Mouse wheel changes the active window instead of scrolling the scrollbar" is unchecked.
### Actual behavior

Tree window list scrolling on OS X dies when "Mouse wheel changes the active window instead of scrolling the scrollbar" is unchecked.
### Steps to reproduce the issue
1. Uncheck "Mouse wheel changes the active window instead of scrolling the scrollbar"
2. Try to scroll the Tree window list
### System information

```
OS X 10.10.5

KVIrc 4.9.2 'Aria'

Runtime Info:
System name: MacOS X
System version: Darwin Kernel Version 14.5.0: Mon Jan 11 18:48:35 PST 2016; root:xnu-2782.50.2~1/RELEASE_X86_64
Architecture: x86_64
Qt version: 5.6.0
Qt theme: macintosh

Build Info:
Build date: 2016-04-16 03:28:45 UTC
Sources date: 20160102
Revision number: 
System name: Darwin-13.4.0
CPU name: x86_64
Build command: /usr/local/Cellar/cmake/3.5.1/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/Users/travis/build/kvirc/KVIrc/build/KVIrc.app
   Threads=POSIX
Compiler name: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++
Compiler flags: N/A
Qt version: 5.6.0
Features: IRC, IPv6, Crypt, SSL, IfAddr, IPC, Transparency, Perl, Python, Enchant, Qt5, KVS
```

un1versal:
What I get in Linux is another bug something old thats never been fixed when trying to duplicate this https://github.com/kvirc/KVIrc/issues/1301 so the reason may or may not be somewhat related.

un1versal:
this issue should be retested after https://github.com/kvirc/KVIrc/pull/2047 is merged

PR #2047, opened 2016-07-07
Fix double-scrolling in tree-style window list

CyberShadow:
When KviOption_boolWheelScrollsWindowsList was on, scrolling the mouse
wheel over the tree-style window list would skip one item. This was
because the mouse wheel would fire both KviWindowListBase::wheelEvent
and KviTreeWindowListTreeWidget::wheelEvent. We fix this by overriding
KviWindowListBase::wheelEvent in KviTreeWindowList, thus preventing it
from firing.

### DIFF
--- a/src/kvirc/ui/KviTreeWindowList.cpp
+++ b/src/kvirc/ui/KviTreeWindowList.cpp
@@ -492,6 +492,13 @@ void KviTreeWindowList::updateActivityMeter()
 	}
 }
 
+void KviTreeWindowList::wheelEvent(QWheelEvent * e)
+{
+	// Override KviWindowListBase::wheelEvent.
+	// Mouse wheel handling is done in
+	// KviTreeWindowListTreeWidget::wheelEvent.
+}
+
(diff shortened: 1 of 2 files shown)


un1versal:
is this it? https://github.com/kvirc/KVIrc/issues/1301

CyberShadow:
I don't think so, but this patch might do something for that bug too.

un1versal:
I cant duplicate #1301 after this, Ive been trying for a bit now and no dice, so I think it fixes it... it also seems to fix https://github.com/kvirc/KVIrc/issues/2019

CyberShadow:
Ah yes, #2019 is the bug I tried to fix here.

un1versal:
would you mind adding to your commit msg

fixes #2019 
fixes #1301 (if you also cant dupe it after this)

un1versal:
There's another scrolling related issue report @ #1940 
staticfox is testing this patch on OS X, so I expect to hear something either way from him on that.

CyberShadow:
> would you mind adding to your commit msg

Done

PR #2048, opened 2016-07-07
Esc saves and clears input line

CyberShadow:
This implements a useful feature from mIRC: the ability to save the
currently typed line to the history, without actually sending it. This
can be occasionally very useful in situations such as when you're typing
a long reply, and in the middle of that you need to stop and send a
short reply to a question that appeared meanwhile.

Regardless, using Esc to clear the current input is a common pattern we
should follow anyway.


un1versal:
usually I use `ctrl+a` then  `ctrl + x` since input history is disabled and I hate logging anything internet related

un1versal:
This works so here goes nothing

PR #2049, opened 2016-07-07
Implement `znc.in/self-message` IRCv3 capability

CyberShadow:
This capability is used to indicate that the client supports receiving
messages previously sent by the user. In ZNC, it is required for proper
playback of private message conversations.

More information: http://wiki.znc.in/Query_buffers#Self_messages

The implementation is fairly straight-forward. In the relevant places,
we switch the method used to identify if a message is a private message
or belongs in a channel - instead of comparing the target to our
nickname, we check if it starts with one of the server-announced
CHANTYPES characters, which has the additional benefit of avoiding a
race condition when changing nicknames.

Possible forward improvements:
- Currently, if the first message received for a window is a
  self-message, we do not call setTarget. As far as I've seen, the
  effect is purely cosmetic (the target is not added to the member list,
  which is hidden by default anyway).
- CTCP code for each CTCP message type needs to be manually updated to
  handle self-messages. This patch only addresses ACTION CTCPs.

### DIFF
--- a/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
+++ b/src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
@@ -844,14 +844,17 @@ void KviIrcServerParser::parseLiteralPrivmsg(KviIrcMessage * msg)
 {
 	// PRIVMSG
 	// :source PRIVMSG <target> :<message>
-	QString szNick, szUser, szHost;
-	msg->decodeAndSplitPrefix(szNick, szUser, szHost);
+	QString szSourceNick, szSourceUser, szSourceHost;
+	msg->decodeAndSplitPrefix(szSourceNick, szSourceUser, szSourceHost);
 
 	QString szTarget = msg->connection()->decodeText(msg->safeParam(0));
 	QString szMsg = msg->connection()->decodeText(msg->safeTrailing());
 
+	QString szTargetNick, szTargetUser, szTargetHost;
+	msg->decodeAndSplitMask(szTarget.toLatin1().data(), szTargetNick, szTargetUser, szTargetHost);
+
 	KviConsoleWindow * console = msg->console();
-	KviRegisteredUser * u = msg->connection()->userDataBase()->registeredUser(szNick, szUser, szHost);
+	KviRegisteredUser * uSource = msg->connection()->userDataBase()->registeredUser(szSourceNick, szSourceUser, szSourceHost);
 	//Highlight it?
 
 	PrivmsgIdentifyMsgCapState eCapState = IdentifyMsgCapNotUsed;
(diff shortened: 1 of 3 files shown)


CyberShadow:
CC @DarthGandalf :)

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
```diff
@@ -954,9 +954,13 @@ void KviIrcServerParser::parseLiteralPrivmsg(KviIrcMessage * msg)
 		//			}
 		//		}
 
+		// "znc.in/self-message" capability: Handle a replayed message from ourselves to someone else.
+		bool bSelfMessage = IS_ME(msg, szNick);
+		QString szWindow = bSelfMessage ? szTarget : szNick;
```
DarthGandalf: This causes the window to be `nick!user@host` instead of `nick`

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
```diff
@@ -1002,15 +1006,16 @@ void KviIrcServerParser::parseLiteralPrivmsg(KviIrcMessage * msg)
...
 				else
 				{
 					// no query yet, create it!
 					// this will trigger OnQueryWindowCreated
-					query = console->connection()->createQuery(szNick);
+					query = console->connection()->createQuery(szWindow);
 					// and this will trigger OnQueryTargetAdded
-					query->setTarget(szNick, szUser, szHost);
+					if (!bSelfMessage)
+						query->setTarget(szNick, szUser, szHost);
```
DarthGandalf: It still should be called, but with `szTarget`

CyberShadow: We do not have the user/host of the receiver in the case of a self-message.

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
CyberShadow: Not for me. I'm guessing that in your case, the self-message PRIVMSG is addressed to the full nick!user@host, instead of just the nick?

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
DarthGandalf: Docs for that function say it supports `*` in such case. Like when you do `/query nick`

src/kvirc/sparser/KviIrcServerParser_literalHandlers.cpp
DarthGandalf: PRIVMSG command supports either.

CyberShadow: Yes, that's what I said. In my tests I didn't see the full variant, so I assumed that ZNC never sends it. I guess it sends whatever the client sent? In either case, how can I reproduce this?

DarthGandalf: I'm not even using ZNC to test this.
`/inject :mynick!a@b PRIVMSG someoneelse!c@d :hi`

CyberShadow: Oh, OK. Well, in that case, should I even bother updating this PR if no software exists that can generate such a message? (I'm on vacation and the soonest I can do that is over a week from now)

DarthGandalf: It's not about software, but about the spec. Even if you didn't see ZNC sending such messages, theoretically it _can_.
And take your time for vacation, it's not urgent anyway :)

CyberShadow:
Updated and rebased.

PR #2054, opened 2016-07-09
Give better names to console log files

### LINKED ISSUE #1710: Save console log files with an actual context name
Because

doesn't really mean much to me as I load my networks in a different order each time, making finding things within logs nearly impossible.

CyberShadow:
This patch names console log files after the IRC network.

Note that the console windows are created without a connection
initially. To avoid creating mostly-empty indexed log files, we do not
begin logging until a connection is created.

Fixes #1710.

CC @un1versal @staticfox

### DIFF
--- a/src/kvirc/ui/KviConsoleWindow.cpp
+++ b/src/kvirc/ui/KviConsoleWindow.cpp
@@ -143,9 +143,6 @@ KviConsoleWindow::KviConsoleWindow(int iFlags)
 
 	m_pTmpHighLightedChannels = new QStringList;
 
-	if(KVI_OPTION_BOOL(KviOption_boolAutoLogConsole))
-		m_pIrcView->startLogging();
-
 	applyOptions();
 }
 
@@ -396,7 +393,10 @@ void KviConsoleWindow::saveProperties(KviConfigurationFile * cfg)
 
 void KviConsoleWindow::getBaseLogFileName(QString & buffer)
 {
-	buffer = QString("CONSOLE_%1").arg(context()->id());
+	if(context()->connection())
+		buffer = context()->connection()->target()->network()->name();
+	else
+		buffer = context()->id();
 }
 
 void KviConsoleWindow::showNotifyList(bool bShow, bool bIgnoreSizeChange)
@@ -504,6 +504,14 @@ void KviConsoleWindow::connectionAttached()
 	connect(m_pContext->connection(), SIGNAL(chanListChanged()), this, SLOT(updateUri()));
 	updateUri();
 	m_pNotifyListView->setUserDataBase(connection()->userDataBase());
+
+	// Update log file name
+	if(KVI_OPTION_BOOL(KviOption_boolAutoLogConsole))
+	{
+		if (m_pIrcView->isLogging())
+			m_pIrcView->stopLogging();
+		m_pIrcView->startLogging();
+	}
 }
 
 void KviConsoleWindow::connectionDetached()


src/kvirc/ui/KviConsoleWindow.cpp
```diff
@@ -396,7 +393,11 @@ void KviConsoleWindow::saveProperties(KviConfigurationFile * cfg)
 
 void KviConsoleWindow::getBaseLogFileName(QString & buffer)
 {
-	buffer = QString("CONSOLE_%1").arg(context()->id());
```
un1versal, line 399: I still need a cup of coffee or 5, but Im not sure that replacing the ID number is a good thing especially since the ID is attached to each context in every other instance that is not log related..

capture

What happens if a specific console ID is reused, see comment in https://github.com/kvirc/KVIrc/issues/1710#issuecomment-167148440
And this still affected by having `boolShowMetworkNameForConsoleWindowListEntry` on or off?

is it possible to do this in the same format as channels etc? i.e.

> ##### console_ID_YEAR **on** NETWORKNAME

un1versal:
I think staticfox willprobably like this...

despite my comment, I dont use logging at all.

src/kvirc/ui/KviConsoleWindow.cpp
CyberShadow, line 399: > I still need a cup of coffee or 5, but Im not sure that replacing the ID number is a good thing especially since the ID is attached to each context in every other instance that is not log related..

Sorry, I don't understand what you mean here. Can you give a concrete example of a situation where this might be a problem?

> What happens if a specific console ID is reused, see comment in #1710 (comment)

The last hunk is what takes care of that.

> And this still affected by having boolShowMetworkNameForConsoleWindowListEntry on or off?

I don't see how that would affect log naming. What do you mean by "still"?

> is it possible to do this in the same format as channels etc? i.e.

Why would you want the ID? IDs are ephemeral, they have no meaning once that console window is closed.

un1versal, line 399: > Sorry, I don't understand what you mean here. Can you give a concrete example of a situation where this might be a problem?

its not so much of a problem as it maybe preferred to know which ID the console is attached to

idk, I dont care either way.

> I don't see how that would affect log naming. What do you mean by "still"?

**still** means I need more coffee... yes youre right, it shouldnt affect logging but it does for channels and queries so Im wondering if this plays nice the same way.
capture

> Why would you want the ID? IDs are ephemeral, they have no meaning once that console window is closed.

~~we're talking about the case were they arent closed or no?~~

ignore that, I was just asking because in my mind it was a more rounded approach., but to me it matters none, so Ill shut up and let you go back to your coffeee cup. :+1:  :coffee:

CyberShadow, line 399: > still means I need more coffee... yes youre right, it shouldnt affect logging but it does for channels and queries so Im wondering if this plays nice the same way.

No, it's not like channel/query windows. Look at `KviChannelWindow::getBaseLogFileName` and `KviQueryWindow::getBaseLogFileName`, they query `windowName()` directly. I guess I could change it to do that but why?

> were talking about the case were they arent closed or no?

OK, but when would that ever be helpful? What would adding that number in the filename allow you to do? As far as I can see it's just clutter that'll get in the way of sorting.

un1versal, line 399: > OK, but when would that ever be helpful? What would adding that number in the filename allow you to do? As far as I can see it's just clutter that'll get in the way of sorting.

I updated my comment above

> No, it's not like channel/query windows. Look at KviChannelWindow::getBaseLogFileName and KviQueryWindow::getBaseLogFileName, they query windowName() directly. I guess I could change it to do that but why?

for the same reason as the last updated comment on last post.

hit refresh...

un1versal:
as a sideways thought and my last comment on this.

Some clients and most software that does logging, assigns default filename variables, but allows configuration of said variables by user... I guess that would be ideal in the long term

un1versal:
capture

yes, this looks OK imo, Ill leave it for @staticfox to comment / merge since its his request.

staticfox:
Looks good, issue is I tried it a few times and the only log file I could get is `standalone server_2016` so it's as if it's not detecting the network itself

un1versal:
> Looks good, issue is I tried it a few times and the only log file I could get is standalone server_2016 so it's as if it's not detecting the network itself

is that a connection via znc?

staticfox:
> > Looks good, issue is I tried it a few times and the only log file I could get is standalone server_2016 so it's as if it's not detecting the network itself
> 
> is that a connection via znc?

Nope.

CyberShadow:
How did you configure KVIrc to connect to servers?

staticfox:
added in https://github.com/kvirc/KVIrc/commit/b489a366ba4f7903a4ff25bc21783d0a870bdab2 Thanks.

staticfox:
For the record, I thought this set the log to the name of the network as generated by numeric `005` i.e. `NETWORK=freenode`

un1versal:
> For the record, I thought this set the log to the name of the network as generated by numeric 005 i.e. NETWORK=freenode

Its always good to test first AV does build exe from PR's

staticfox:
I did test locally

CyberShadow:
> For the record, I thought this set the log to the name of the network as generated by numeric 005 i.e. NETWORK=freenode

We find that out too late - it would mean changing the target log file in the middle of a connection.

PR #2060, opened 2016-07-10
OptionWidget_*: having 0 is equivalent to turning this logging off

un1versal:
#### Changes proposed
- fix minimum log paste values

also theres was a discrepancy between tooltip minimum value and actual
minimum value.


un1versal:
Cybershadow gave blessing, thx.

PR #2070, opened 2016-07-20
actions/popups/events: add option controller dialog for CTCP ignore and joins/parts/quits

un1versal:
### please do not squash Ill merge this in properly

> ### NOTE Illustrations below are meant to show how options are layed out ONLY. The State of the Options in illustrations does not reflect defaults. I.E its for illustration purposes only.
#### Changes proposed

**popups**
- optionctrlrs ( this is used to control both of the above so its easily
  added anywhere and easily expanded )
- this adds suppress on join/part/quit control menus actions
- this now houses previous CTCP Protection control actions

**events**
- add default events for Joins / Parts / Quits
- add button for accessing Option Controllers from main UI ( similar to logging )

These events are disabled by default meaning you see joins parts and quits per default.

**Note** Whenever something is Enabled or Disabled there's textual feedback

demo
## Known Issues
- [x] Dialog can be opened multiple times
- [x] If user closes dialog via title bar close button (%PDisplay is not unset and dialog cant be reopened until unsetting this global) Feature requests allow close button on title bar to be disabled/removed?
- [ ] Dialog can be open in as panel but only left hand side, this looks awful
- [ ] When opening Dialog as in panel Window title in treeview is invisible until selected. (bug!!)

### DIFF
--- a/data/defscript/actions.kvs
+++ b/data/defscript/actions.kvs
@@ -1,0 +1,392 @@
+# Default actions file
+#
+# Coding style/curly brackets: Allman style
+
+action.create -w=xcqd -t=tools ("optionctrlr",$tr("Option controller"),$tr("Opens a control dialog for easy CTCP ignore and joins/parts/quits management."),"kvi_bigicon_settings.png",$icon(options))
+{
+	#################################################################
+	#                                                               #
+	#   File : actions.kvs (optionctrlr) v4.1.1.6810                #
+	#   Creation date : Fri Jul 22 16:57:41 2016 UTC by un1versal   #
+	#                                                               #
+	#################################################################
+
+	# Prevent dialog from opening multiple times
+	if(%PDisplay)
+		halt;
+	else
+		%PDisplay = $true;
+
+	# This prologue houses the variables for the option controller dialog
+	# Add CTCP Protection Variables
+	%ProtectPing = $option(boolIgnoreCTCPPing)
+	%ProtectFinger = $option(boolIgnoreCTCPFinger)
+	%ProtectVersion = $option(boolIgnoreCTCPVersion)
+	%ProtectUserinfo = $option(boolIgnoreCTCPUserinfo)
+	%ProtectClientinfo = $option(boolIgnoreCTCPClientinfo)
+	%ProtectSource = $option(boolIgnoreCTCPSource)
+	%ProtectTime = $option(boolIgnoreCTCPTime)
+	%ProtectPage = $option(boolIgnoreCTCPPage)
+	%ProtectAvatar = $option(boolIgnoreCTCPAvatar)
+	%ProtectDCC = $option(boolIgnoreCTCPDCC)
+
+	# Add Join Suppression Variables
+	%IsJoinEnabled = $isEventEnabled(OnJoin, suppressJoinMsgs)
+	%IsPartEnabled = $isEventEnabled(OnPart, suppressPartMsgs)
+	%IsQuitEnabled = $isEventEnabled(OnQuit, suppressQuitMsgs)
+
+	# Create the main widget as a dialog
+	%widget = $new(dialog)
+	%layout = $new(layout,%widget)
+	%widget->$setWindowTitle($tr("Option Controllers - KVIrc"))
+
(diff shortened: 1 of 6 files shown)


wodim:
I like the idea but I don't like the UI at all. I think this would be better as a panel like the panel you can use to change channel modes.

un1versal:
> I like the idea but I don't like the UI at all. I think this would be better as a panel like the panel you can use to change channel modes.

Yes, well thers two ways to do that...

:x:  Hard-coding in Qt,  not so good idea  then its not so easilly expandable in future and requires compiling the app also.

:radio_button:  Whole UI in KVS, and while I find that idea appealing, I have to say no way at this time.
  mabye later on its a lot to code into `OnChannelWindowCreated` for such a simple usage... way too much code if you ask me to keep t simple.

un1versal:
Ah yes and the popup editor in case I coded that in KVS into a popup, is unusable, it crashed several times and wiped the entire popups file, so the popups editor is dead pretty much for anything else than very light editing.

un1versal:
> I think this would be better as a panel like the panel you can use to change channel modes.

OK Ive entertained that idea in KVS but am stuck in how to make this open in a panel

https://gist.github.com/un1versal/f7ce11b4f14a8a5260b90fa82d4099ab

That dialog currently only shows what is set or not its enough as a proof of concept.

use script tester for testing.

any help would be apprecciated

* make dialog open as a panel on right hand side

OmegaPhil:
Sorry, I have no experience working with panel stuff in KVS - a quick search of the help shows its probably not possible.

From your comments - presumably joins/parts/quits will not be disabled by default for people starting up with a brand new configuration? If they were disabled, for me that would make KVIrc appear to not be a serious IRC client, if I came to it as a new user.

un1versal:
> From your comments - presumably joins/parts/quits will not be disabled by default for people starting up with a brand new configuration?

They will be disabled by default, meaning you can see by default joins / parts / quits. 

Currently they dont exist by default, so as is you dont have the option unless youre fiddling to add this to be enabled it makes sense to follow that logic.

> If they were disabled, for me that would make KVIrc appear to not be a serious IRC client, if I came to it as a new user.

I dont understand that statement not one bit.

So to you makes sense that by default your IRC client blocks joins/parts/quits (as in you dont see any activity at all) so you can consider it a serious client?

OmegaPhil:
The reverse - if the IRC client blocked those messages by default, it wouldn't be a serious client for me. Glad to see that KVIrc won't do that.

un1versal:
> The reverse - if the IRC client blocked those messages by default, it wouldn't be a serious client for me. Glad to see that KVIrc won't do that.

yes, wouldnt dream of it.

Are you available in IRC I could use some assist.

un1versal:
Ok dialog now fully operational and ready to go, it still doesn't open as a panel, but unless anyone knows how to, I wont bother at this time.

un1versal:
I managed to open this docked but it only docks to right hand side and its really really buggy, kvirc ftw

un1versal:
The dialog has been moved to a deafult action, its not glitchy to trigger and its really not a popup and doesnt work well as a popup.

un1versal:
~~yep this is a problem, the action wont load from default even though it doesnt exist. which means all you get is errors trying to trigger something that for all intents and purposes KVIrc thinks it doesnt exist.~~

~~restoring default script works fine since https://github.com/kvirc/KVIrc/pull/2070/commits/747125d741b35e0226b119f313ae2b4cd1bbbeb0 but, if a default script doesnt exist in userspace, then it should be loaded from default~~

~~This loading from default only happens at first run, imo it should run always,~~

there only so much wishful thinking allowed, the above has been solved

un1versal:
Sadly the actions file is not loaded on first run...

Can be restored ok, but it doesnt load on first run, anyone have a pointer where to look?

un1versal:
OK, the last thing holding this was the dialog could be opened multiple times even it it was already open,,,

~~thats now been fixed and this is now ready to go,~~

edit... some issue check open logic... :/ dammit 

Opening as a panel looks crap and cant really get it to work properly, so its a standalone

Issue #2088, opened 2016-08-04
Bug: Flag set on Actions has buggy Signal/Slot


un1versal:
### Expected behavior

That when setting flags for the action no previous flags that already have been set should be unset 
### Actual behavior

previous flags are unset as you go down ticking boxes and you need to go up and re set them
### Steps to reproduce the issue

![demo-a](https://cloud.githubusercontent.com/assets/3521959/17394185/a692e35e-5a1f-11e6-92c3-429a628d6c28.gif)

reference https://github.com/kvirc/KVIrc/blob/master/src/modules/actioneditor/ActionEditor.cpp#L193-L271
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Linux 4.4.0-31-generic
System version: #50-Ubuntu SMP Wed Jul 13 00:07:12 UTC 2016
Architecture: x86_64
Qt version: 5.5.1
Qt theme: fusion

Build Info:
Build date: 2016-08-04 04:59:12 UTC
Sources date: 20160803
Revision number: git-6852-g2797d25
System name: Linux-4.4.0-31-generic
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler name: /usr/bin/c++
Compiler flags: N/A
Qt version: 5.5.1
Features: IRC, Debug, Crypt, SSL, IfAddr, IPC, OSS, Transparency, Webkit, Perl, Python, Enchant, Qt5, KVS
```

PR #2091, opened 2016-08-07
Fixes Issue #2088

IceN9ne:
Separates the logic for check box enabling/disabling into their own functions.


un1versal:
Thx for this. ;)

Issue #2073, opened 2016-07-25
Multi-line pasting does not work in console


staticfox:
### Expected behavior

Pasting a block of test in to console would allow the IRCd to parse the block of test
### Actual behavior

KVIrc stops sending after the first command
### Steps to reproduce the issue
1. Copy a few lines of test in to your clip board
2. Paste it in to console output
3. It sends the first line but nothing more
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Windows Release : 6.2.9200
System version: Windows 8 Professional (x64) (Build 9200)
Architecture: Intel64 Family 6 Model 60 Stepping 3
Qt version: 5.5.1
Qt theme: windowsvista

Build Info:
Build date: 2016-07-14 10:29:35 UTC
Sources date: 20160714
Revision number: git-6807-g2fa46d9
System name: Windows-6.3.9600
CPU name: AMD64
Build command: C:/Program Files (x86)/CMake/bin/cmake.exe
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=C:/projects/kvirc/build/release
   Threads=Win32
Compiler name: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/amd64/cl.exe
Compiler flags: /DWIN32 /D_WINDOWS /W3 /GR /EHsc
Qt version: 5.5.1
Features: IRC, IPv6, Crypt, SSL, IPC, Transparency, Webkit, Perl, Python, Enchant, Qt5, KVS
```

As for the _why_, sometimes I get bored and like to copy/paste operwall ascii art spam :v. Example:

``````
OPERWALL :quu..__
OPERWALL : $$$b  `---.__
OPERWALL :  "$$b        `--.                          ___.---uuudP
OPERWALL :   `$$b           `.__.------.__     __.---'      $$$$"              .
OPERWALL :     "$b          -'            `-.-'            $$$"              .'|
OPERWALL :       ".                                       d$"             _.'  |
OPERWALL :         `.   /                              ..."             .'     |
OPERWALL :           `./                           ..::-'            _.'       |
OPERWALL :            /                         .:::-'            .-'         .'
OPERWALL :           :                          ::''\          _.'            |
OPERWALL :          .' .-.             .-.           `.      .'               |
OPERWALL :          : /'$$|           .@"$\           `.   .'              _.-'
OPERWALL :         .'|$u$$|          |$$,$$|           |  <            _.-'
[...]

PR #2092, opened 2016-08-08
Fixes Issue #2073

IceN9ne:
This fixes the issue in #2073.

However, I'm not sure if there was a reason for having the return in there preventing you from submitting more than one raw command in the console.  Maybe for preventing dumb mistakes and spamming the server with potentially dangerous raw commands?


un1versal:
> Maybe for preventing dumb mistakes and spamming the server with potentially dangerous raw commands?

You mean, because you checked all the boxes?

staticfox:
See https://github.com/kvirc/KVIrc/issues/2073#issuecomment-238955960, I appreciate the pull request though but you're right.

PR #2115, opened 2016-08-20
Add a line highlighter for the current line in the script editor widget.

IceN9ne:
When using the `ScriptEditorWidget` in any of the instances in KVIrc, it now includes a line highlight for the current line containing the cursor.  If you want to disable it, just set "Background Highlight" color to match the "Background" color in the Configure Editor dialog.

### DIFF
--- a/src/modules/editor/ScriptEditorImplementation.cpp
+++ b/src/modules/editor/ScriptEditorImplementation.cpp
@@ -67,6 +67,7 @@ extern std::set<ScriptEditorImplementation *> g_pScriptEditorWindowList;
 extern KviModule * g_pEditorModulePointer;
 
 static QColor g_clrBackground(0, 0, 0);
+static QColor g_clrBackgroundHighlight(50, 50, 50);
 static QColor g_clrNormalText(100, 255, 0);
 static QColor g_clrBracket(255, 0, 0);
 static QColor g_clrComment(0, 120, 0);
@@ -130,6 +131,8 @@ ScriptEditorWidget::ScriptEditorWidget(QWidget * pParent)
 	}
 	else
 		loadCompleterFromFile();
+
+	connect(this, &ScriptEditorWidget::cursorPositionChanged, this, &ScriptEditorWidget::highlightCurrentLine);
 }
 
 ScriptEditorWidget::~ScriptEditorWidget()
@@ -214,6 +217,24 @@ void ScriptEditorWidget::asyncCompleterCreation()
 		delete m_pListCompletition;
 		delete m_pListModulesNames;
 	}
+}
+
+void ScriptEditorWidget::highlightCurrentLine()
+{
+	if(isReadOnly())
+	{
+		setExtraSelections(QList<QTextEdit::ExtraSelection>());
+		return;
+	}
+
+	QTextEdit::ExtraSelection selection;
+
+	selection.cursor = textCursor();
+	selection.cursor.clearSelection();
+	selection.format.setBackground(g_clrBackgroundHighlight);
+	selection.format.setProperty(QTextFormat::FullWidthSelection, true);
+
+	setExtraSelections(QList<QTextEdit::ExtraSelection>({ selection }));
 }
 
 void ScriptEditorWidget::loadCompleterFromFile()
@@ -475,6 +496,7 @@ ScriptEditorWidgetColorOptions::ScriptEditorWidgetColorOptions(QWidget * pParent
 	gbox->setInsideSpacing(0);
 
 	addColorSelector(gbox, __tr2qs_ctx("Background:", "editor"), &g_clrBackground, true);
+	addColorSelector(gbox, __tr2qs_ctx("Background Highlight:", "editor"), &g_clrBackgroundHighlight, true);
 	addColorSelector(gbox, __tr2qs_ctx("Normal text:", "editor"), &g_clrNormalText, true);
 	addColorSelector(gbox, __tr2qs_ctx("Brackets:", "editor"), &g_clrBracket, true);
 	addColorSelector(gbox, __tr2qs_ctx("Comments:", "editor"), &g_clrComment, true);
@@ -784,6 +806,7 @@ void ScriptEditorImplementation::loadOptions()
 
 	KviConfigurationFile cfg(szTmp, KviConfigurationFile::Read);
 	g_clrBackground = cfg.readColorEntry("Background", QColor(0, 0, 0));
+	g_clrBackgroundHighlight = cfg.readColorEntry("BackgroundHighlight", QColor(50, 50, 50));
 	g_clrNormalText = cfg.readColorEntry("NormalText", QColor(100, 255, 0));
 	g_clrBracket = cfg.readColorEntry("Bracket", QColor(255, 0, 0));
 	g_clrComment = cfg.readColorEntry("Comment", QColor(0, 120, 0));
@@ -830,6 +853,7 @@ void ScriptEditorImplementation::saveOptions()
 
 	KviConfigurationFile cfg(szTmp, KviConfigurationFile::Write);
 	cfg.writeEntry("Background", g_clrBackground);
+	cfg.writeEntry("BackgroundHighlight", g_clrBackgroundHighlight);
 	cfg.writeEntry("NormalText", g_clrNormalText);
 	cfg.writeEntry("Bracket", g_clrBracket);
 	cfg.writeEntry("Comment", g_clrComment);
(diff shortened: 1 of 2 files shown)


src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -475,6 +496,7 @@ ScriptEditorWidgetColorOptions::ScriptEditorWidgetColorOptions(QWidget * pParent
 	gbox->setInsideSpacing(0);
 
 	addColorSelector(gbox, __tr2qs_ctx("Background:", "editor"), &g_clrBackground, true);
+	addColorSelector(gbox, __tr2qs_ctx("Background Highlight:", "editor"), &g_clrBackgroundHighlight, true);
```
un1versal, line 499: Wouldnt it be nicer if this was simply called **Highlight current line** with no Camel Case since it is a label?

Ide follow that format even for variables and configuration entry names too names as some already are.

edit see https://github.com/IceN9ne/KVIrc/pull/1

src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -67,6 +67,7 @@ extern std::set<ScriptEditorImplementation *> g_pScriptEditorWindowList;
 extern KviModule * g_pEditorModulePointer;
 
 static QColor g_clrBackground(0, 0, 0);
+static QColor g_clrBackgroundHighlight(50, 50, 50);
```
un1versal, line 70: Im not sure  #323232 Is that great now I see it implemented. It works as desired, but this means users have to select a dark color for highlight which feels counter intuitive (if user prefers a light theme the line highlight must be light as opposed to dark also counter intuitive).

Im not sure what a better alternative would be, but usually this is done with a transparency e.g.
`static QColor g_clrSomeTarget(198, 198, 198, .4);`

not sure how this would work for user set colors considering dark and light editor scenarios.

maybe we should use tooltips here too....

I found a bug though. when changing the editor colors to a light theme most colors were not applied until an editor restart. see screenshot

capture

**Normal text** and **Comments** were not applied at least.

IceN9ne, line 70: This doesn't force them into using a dark theme.  The default theme for the script editor is a dark theme, so the default for "Background highlight" should be for a dark theme as well.  It doesn't make sense to default it to match a light theme as the user would have had to change it from dark to something else.

I thought about doing `lighten(##)` but that would only make sense for dark themes and it would fail on light themes, hence just setting it as something that makes sense for default and allowing the pro themers to change it to what they want.  Maybe they want a pink highlight on a normally black background?

src/modules/editor/ScriptEditorImplementation.cpp
IceN9ne, line 499: I'd agree with changing it to "Background highlight", but I'm not in favor of "Highlight current line" since I am still deciding if I want to add to it a multi-line highlighter when you do a multi-line selection. So, labeling it as "current line" wouldn't make sense.

Also, It's a replacement background color for when the cursor is on that line, so labeling it as "Background highlight" makes more sense than "Highlight current line" which seems like it could be several different things like "transparency setting" or something.

Maybe you'd prefer "Highlight background"?

src/modules/editor/ScriptEditorImplementation.cpp
un1versal, line 70: > This doesn't force them into using a dark theme. The default theme for the script editor is a dark theme, so the default for "Background highlight" should be for a dark theme as well. It doesn't make sense to default it to match a light theme as the user would have had to change it from dark to something else.

No, of course not, it has to work with dark and light or any other combination ppl want.

My main concern here is theres only a very small range of dark/light colors that will work if you choose one color too light or dark depending you either cant see it or you cant read selected text.

I also thought about adding a toggle to enable disable highlighting current line.

So for this we should add some tooltip explaining limitation.

> I thought about doing lighten(##) but...

idk if its worth the extra work to spend more time on this but tooltip should help some.

src/modules/editor/ScriptEditorImplementation.cpp
un1versal: > Maybe you'd prefer "Highlight background"?

that would be a way to go, but which ever way you wanna go multiline or not but ide make it consistent with the code part cus atm you have a mix that are neither here nor there.

something like https://github.com/IceN9ne/KVIrc/pull/1 for however you want to go

IceN9ne: I fail to see any "mix" that you speak of.

src/modules/editor/ScriptEditorImplementation.cpp
IceN9ne, line 70: You can easily disable it by setting both background colors to be the same.  I fail to see the need for a toggle.

src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -216,6 +219,24 @@ void ScriptEditorWidget::asyncCompleterCreation()
 	}
 }
 
+void ScriptEditorWidget::highlightCurrentLine()
```
un1versal: +void ScriptEditorWidget::highlightCurrentLine() 

mix shoudl be then  +void ScriptEditorWidget::backgroundHighlight()  or no? why say highlightcurrent line in some and backgroundhighligh in others?

I know its code only and not visible to user...

 its exhausting trying to explain what I mean.

src/modules/editor/ScriptEditorImplementation.cpp
un1versal: you crack me up you do...

src/modules/editor/ScriptEditorImplementation.h
```diff
@@ -84,6 +84,7 @@ public slots:
 	void slotReplace();
 protected slots:
 	void asyncCompleterCreation();
+	void highlightCurrentLine();
```
un1versal: mix

src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -830,6 +853,7 @@ void ScriptEditorImplementation::saveOptions()
 
 	KviConfigurationFile cfg(szTmp, KviConfigurationFile::Write);
 	cfg.writeEntry("Background", g_clrBackground);
+	cfg.writeEntry("BackgroundHighlight", g_clrBackgroundHighlight);
```
un1versal: mix

src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -784,6 +806,7 @@ void ScriptEditorImplementation::loadOptions()
 
 	KviConfigurationFile cfg(szTmp, KviConfigurationFile::Read);
 	g_clrBackground = cfg.readColorEntry("Background", QColor(0, 0, 0));
+	g_clrBackgroundHighlight = cfg.readColorEntry("BackgroundHighlight", QColor(50, 50, 50));
```
un1versal: mix

src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -216,6 +219,24 @@ void ScriptEditorWidget::asyncCompleterCreation()
...
+	{
+		setExtraSelections(QList<QTextEdit::ExtraSelection>());
+		return;
+	}
+
+	QTextEdit::ExtraSelection selection;
+
+	selection.cursor = textCursor();
+	selection.cursor.clearSelection();
+	selection.format.setBackground(g_clrBackgroundHighlight);
```
un1versal: mix

src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -130,6 +131,8 @@ ScriptEditorWidget::ScriptEditorWidget(QWidget * pParent)
 	}
 	else
 		loadCompleterFromFile();
+
+	connect(this, &ScriptEditorWidget::cursorPositionChanged, this, &ScriptEditorWidget::highlightCurrentLine);
```
un1versal: mix

IceN9ne:
As to avoid conflicts with the KVIrc dev team, I'll remove this pull request.

un1versal:
> As to avoid conflicts with the KVIrc dev team, I'll remove this pull request.

thats just unecessary but thats your choice. next time Ill shudup and say nothing

sorry it came to this

un1versal:
actually I rather you didnt close it, I think its not worth getting to that.

PR #2116, opened 2016-08-20
KviMainWindow: Modernize control loop

IceN9ne:
A good opportunity to clean things up with a range-based for loop.


staticfox:
I would rather you use a reference here to avoid making a copy. Also please look up 'emplace_back' as it will automatically call the constructor so then all you'll need to do is supply it the arguments.

DarthGandalf:
Copy of an int isn't expensive.

staticfox:
Oh it's an int, didn't see. On a 60 foot telephone pole atm.

IceN9ne:
This is a vector of pointers (`std::vector<QShortcut*>`), so I would disagree with the usage of `emplace_back` as that is not what it was designed for. `push_back` would be clearer and more appropriate.  Even with `std::vector<std::unique_ptr<QShortcut>>` I wouldn't use `emplace_back`.

PR #2129, opened 2016-09-01
KviInputEditor cleanup

### LINKED ISSUE #2065: input line focus changes adds random spaces into input line
### Expected behavior

That no matter if you **click** away from input line when returning to input line to complete the remainder of nick, or message you are typing, that this is done so without weird features like random size spaces being added..
### Actual behavior

If you type 3 letters of a nickname and select the user on userlist, when returning to input line to tab complete the remainder of nick when pressing tab, some spaces are added between half completed nick and the new nick is then completed.

if you type something on input line click on desktop or any other area that makes input line loose focus when returning to complete what you were typing a random ammount of  spaces may be added in between what you are typing.
1. You must type some alias (doesnt matter if its fake) e.g `/something`
2. type first 3 characters of a nick then goto userlist select another nick and then the target nick you maybe want to complete (its important the cursor is no longer flashing on input line)
3. return to input line click near the 3 letters of nick and press tab

sling
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Linux 4.4.0-31-generic
System version: #50-Ubuntu SMP Wed Jul 13 00:07:12 UTC 2016
Architecture: x86_64
Qt version: 5.5.1
Qt theme: fusion

Build Info:
Build date: 2016-07-14 07:44:59 UTC
Sources date: 20160710
Revision number: git-6806-g5113bcd-dirty
System name: Linux-4.4.0-30-generic
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler name: /usr/bin/c++
Compiler flags: N/A
Qt version: 5.5.1
Features: IRC, Debug, Crypt, SSL, IfAddr, IPC, OSS, Transparency, Webkit, Perl, Python, Enchant, Qt5, KVS
```

IceN9ne:
- Updates the partial string matching for tab completion to use modern c++
- Cleanup on comments and code style
- Fixes HTML code in the context menu for clipboard preview
- Fixes issue of having too many newlines before the 60 character truncation which would spam the context menu potentially causing it to go to 2 or more columns
- Fixes uninitialized Slow Paste status flag that would randomly cause it to show in context menu depending on what was in that memory location previously
- Fixes issue of not properly removing $ prefix on kvs functions in tab completion causing an extra $ to be inserted before functions
- Disables auto appending of ' ' and '(' when in `/help` mode since you wouldn't want it in this context
- Fixes tab completion for kvs function $$
- Fixes #2065

### DIFF
--- a/src/kvirc/ui/KviInputEditor.cpp
+++ b/src/kvirc/ui/KviInputEditor.cpp
@@ -71,6 +71,8 @@
 
 #include <qdrawutil.h> // qDrawShadePanel
 
+#include <algorithm>
+
 #if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
 #include <windows.h>
 #endif //COMPILE_ON_WINDOWS || COMPILE_ON_MINGW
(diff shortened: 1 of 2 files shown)


IceN9ne:
Several more `KviInputEditor` fixes to come tomorrow.  Testing if appveyor will accept this way of doing lambdas or else I'll just remove the first commit of this pull request.

un1versal:
Is there any specific existing bug reports you would like testing against this?

IceN9ne:
I don't recall any open bug reports for any of these.  I found the bugs during the original code refactor.

src/kvirc/ui/KviInputEditor.cpp
```diff
@@ -2282,54 +2293,38 @@ void KviInputEditor::completion(bool bShift)
...
 		}
 		else
 		{
 			QString szAll;
 			szMatch = tmp.front();
-			int iWLen = szWord.length();
-			if(szMatch.left(1) == '$')
-				szMatch.remove(0, 1);
-			for(auto szTmpIter : tmp)
+			auto predicate = bIsDir
```
staticfox: Can't we just put the `bIsDir` inside the lambda? That removes the need for `std::function`, the ternary stuff, and it just makes the code look cleaner overall.

IceN9ne: That works too.  I try to stay away from lambda captures unless strictly necessary, but since MSVC (maybe older gcc too) requires a `std::function` wrapper, I'd agree with you on it looking better.

un1versal:
Is this OK to go now or you still working on this?

IceN9ne:
I'm done with it

un1versal:
thanks

PR #2132, opened 2016-09-05
Fixes keyboard shortcuts in split view IRC channel. Fixes #1319

### LINKED ISSUE #1319: "split view" notices pane takes actions instead of chat pane. e.g. scroll, font-size (^+, ^-), and search
**Reported by smashingx on 31 Aug 2012 19:11:09 UTC**
I was wondering why always when I press the key up page or down page the one that goes or up is the window where all the join/lefts are logged the one that moves and not the channel (I have the split view)

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1319

IceN9ne:
By clicking inside one of the two views in a split view IRC channel, keyboard shortcuts now apply to the last clicked view.

### DIFF
--- a/src/kvirc/ui/KviChannelWindow.cpp
+++ b/src/kvirc/ui/KviChannelWindow.cpp
@@ -2008,6 +2008,14 @@ void KviChannelWindow::checkChannelSync()
 	}
 }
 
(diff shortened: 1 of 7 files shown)


un1versal:
thanks

PR #2133, opened 2016-09-05
Fix undefined behavior when trying to find an invalid window id

IceN9ne:
When trying to find a window via an invalid id (ex: `/echo $window.type(0)`) will result in undefined behavior when returning from the `std::map.find()` and calling `second`.

Edit: Wording clarification.


un1versal:
Do you mean if you run `/echo $window.type(0)` it segfaults?

over here it outputs

```
[22:16:55] ��@
```

IceN9ne:
Reworded it for you.

un1versal:
thanks

failure unrelated.

PR #2142, opened 2016-09-19
popups: move user control prologue and options to own popup

un1versal:
This separates the channel popup prologue at
https://github.com/kvirc/KVIrc/blob/master/data/defscript/popups.kvs#L1614-L1661
and the control popup at
https://github.com/kvirc/KVIrc/blob/master/data/defscript/popups.kvs#L1702-L1808
into a its own usercontrol popup.

this allows:

1) Calling the usercontrol as a external menu instead from existing channel popup
2) Allows extending user control in future easily and reusing by calling as externalpopup
   and avoid duplicating code.


staticfox (approved):
I have no objections

staticfox:
@un1versal it's been 2 weeks, if this works then I say kick it in.

un1versal:
be my guest

PR #2146, opened 2016-09-28
Add back the X button to close the current window

wodim:
So this adds back the X button we had before we got rid of MDI. I think it looks good on Ubuntu, but no clue about other systems, so I'm posting it as a PR so appveyor works on it and I can test it on Windows.

Might be a bit hacky too.


wodim:
This is bad, it's too tall on Windows, we'd need a smaller icon with less padding on the button itself.

wodim:
That or make it a menu entry on the QMenuBar and align it to the right. This is what Notepad++ does, but I think it looks very ghettoish.

Cizzle:
Suggestions:
- Use a QToolButton with autoraise enabled
- m_pCloseActiveWindowButton->setFixedSize(m_pCloseActiveWindowButton->iconSize().width(), m_pCloseActiveWindowButton->iconSize().height() + 4);
  The height might need a better approach, using the pop->size().height() gave a larger number than the menubutton would indicate.

wodim:
By the way this is what they looked like before.





As you can see, those icons are from the Qt theme. I wonder if there's any way to recover those from the current Qt theme as QIcons?

ctrlaltca:
> I wonder if there's any way to recover those from the current Qt theme as QIcons?

``` cpp
QStyle* style = QApplication::style();
QIcon icon = style->standardIcon( QStyle::SP_TitleBarCloseButton );
```

Stanzilla:
Not sure why you want to bring it back? Right click and "close" not obvious enough?

Cizzle:
How is that obvious? How many programs use a rightclick+close method as only mouse-closing method? Plus it takes a rightclick on the channelnamebutton, movement to the close-menuitem and a leftclick on it just to close it.

If you use the classic window list you can enable a close button on each channelnamebutton (for the tree window list it's WIP), but having one closebutton always in the same spot is more intuitive.

wodim:
I don't think it's obvious. We have gotten several peeps asking how to close windows... And remember, the X button has only disappeared in nightlies. Imagine when the actual thing is released.

wodim:
Grrr, this looks taller too.



I guess I'll have to look into `setFixedSize` or something like that.

wchristian:
So it doesn't get lost: You can pull a fixed-up version of your branch from here, then force push it: https://github.com/wchristian/KVIrc/tree/closebutton

PR #2147, opened 2016-10-02
Shut up VC++ and Fix Install Launch Bug

MrHamel:
## 
## 
## 

Make the MS VC++ redist installer run silently, it's just an unnecessary popup.

Fixed the finish page so it will launch KVIrc as the user that ran the install before UAC privilege escalation.


staticfox:
👍 Logic seems good, I'm certain that AV is just missing the plugin. I'll pass it on to @Dessa for further review as he knows NSIS better than I.

staticfox:
@Dessa ?

Stanzilla:
https://github.com/kvirc/KVIrc/pull/2368

PR #2157, opened 2016-11-12
KviIpEditor: Change to a QLineEdit and update code.

IceN9ne:
What this changes:
  * Changes from a multi-QLineEdit box for each part of the IP address
    to a single QLineEdit which allows pasting of IP addresses rather
    than having to hand type them in.
  * Uses Qt5's QHostAddress to proccess the input and validate the IPs
  * Fixes the OptionsWidget_servers to enable/disable the KviIpEditor
    depending on if the Cache IP checkbox is checked or not.
  * Fixes a spot in OptionsWidget_servers that was verifying an IPv6 with an IPv4 checker.

Not sure if this is desirable, so I'll ask for feedback.  If people want the individual boxes instead, I can just update that code instead.

This validates IPv4 addresses as you type them. I also added a basic IPv6 input validator that only accepts 0-9a-fA-F and : for addresses, except when you have ::ffff: prefx which allows for IPv6 mapped IPv4 addresses (Example `::ffff:1.2.3.4`).  If someone wants to write a more complete IPv6 input validator, I'd be grateful.

### DIFF
--- a/src/kvirc/ui/KviIpEditor.cpp
+++ b/src/kvirc/ui/KviIpEditor.cpp
@@ -44,6 +44,8 @@ void KviIpEditor::setAddressType(AddressType addrType)
 	{
 		m_addrType = addrType;
 		QString tmp = text();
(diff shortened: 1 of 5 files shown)


wodim:
travis failed

un1versal:
> travis failed

timeout maybe thers no fail where it matters

wodim:
no no... it did actually fail. but it's good now.

and I haven't tried this, but I really really hated the old control, so :+1:

PR #2165, opened 2016-12-11
Add compatibilty to fish10 and mircryption keyx cbc mode.

Skeddie:
Mimics fish10 when getting a DH1080_INIT from a fish10 user so the other party will use CBC mode as well.
Also adds support for DH1080_FINISH_cbc being received from Mircryption or other clients.

### DIFF
--- a/src/modules/fish/libkvifish.cpp
+++ b/src/modules/fish/libkvifish.cpp
@@ -160,9 +160,15 @@ bool fish_event_onQueryNotice(KviKvsModuleEventCall * c)
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


IceN9ne (changes requested):
There won't be a situation where the KVIrc client receives a DH1080_FINISH_cbc, since it will never send a DH1080_INIT_cbc.  That part of the code should be removed before being merged.

If you can show me a situation that says otherwise, I'll approve it.

Cizzle:
Silently removed with a fixup rebase.

IceN9ne (approved):

Issue #2178, opened 2017-01-24
Drop QtWebKit support


staticfox:
It's deprecated and should be replaced with something like QtWebEngine or a text viewer for now. Regardless, once that's implemented it should make porting KVIrc to MSVC2015 a breeze.

IceN9ne:
This seems related to #1854

SolarAquarion:
debian sid and stretch has qtwebengine

wodim:
msys2 on the other hand does not.

I'm thinking about this change because it seems fairly simple, but is qtwebengine available in the version of qt that resides in appveyor?

Dessa:
it should be, we could try a few versions if we are not sure

a17r:
Any update here?

ctrlaltca:
Closing as dup #1854

PR #2179, opened 2017-01-24
Drop webkit support

staticfox:
There's some left over stuff in here that only relies on webkit... but it looks like it's basically a web browser? Not sure if it's worth porting over to something new, let alone worth keeping.

Closes https://github.com/kvirc/KVIrc/issues/2178


staticfox:
The only thing this appears to break is a dialog that pops up with a web browser to (what once was) the theme repository. Going to remove it and have it just link to our theme repo on GitHub unless someone objects.

Side note, @AlexanderP, I removed the dependency correctly from our PPA build script right? (Just want to make sure, I'm not versed with that stuff)

staticfox:
All that remains are updated openurl links to the corresponding things that were dropped (theme browser, script browser). ill do that in a few hours.

HelLViS69:
This will also break all scripts which uses webkit as a browser.. Instead of removing it, it was better to sign it as deprecated and port all stuff to the new way Qt5 uses.. just my 2 cents

staticfox:
Hmm.. As per the mini discussion that took place in #kvirc, it was brought up that there was no point to migrate to the new qtwebengine as it doubled binary size for less of a reward. Not going to merge this any time soon, just curious on the pros and cons.

HelLViS69:
So, why don't implement qtwebengine as an optional support? we have CMake to enable/disable stuff.. I think removing it at all is a bad idea.. KVIrc is so sexy because you can do almost anything with KVS..

staticfox:
Alright. As per an upgrade path, how should we go about that? Keep qtwebkit and implement qtwebengine on top of that and mark qtwebkit as deprecated? Or can we just migrate qtwebkit over to qtwebengine. qtwebengine supports 5.4 >= IIRC so it fits within our target

HelLViS69:
Mmm.. as we don't know which qt users have, we have 2 choices:

1. implement a TAL (Toolkit Abstraction Layer) to let user's system choose which one it has and include/exclude code using ifdefs
2. upgrade CMake requirements to use Qt >= 5.4 and use only QtWebEngine

staticfox:
Maybe it would make sense to use 2) as our CIs already use 5.5 >= anyway. And I'm pretty sure a good majority of users just use CI artifacts anyway.

DarthGandalf:
There are categories of users:
* Majority of users try to install 4.2.0, the last stable version
* Majority of Linux users install whatever is available in their favorite package manager
* The rest install from git; on Windows the majority of such users install the CI artifact

This is just my educated guess, I obviously didn't make any study what exactly majority of users do.

SolarAquarion:
https://wiki.qt.io/Porting_from_QtWebKit_to_QtWebEngine bunch of APIs that kvirc uses are not availible

SolarAquarion:
actually only one

wodim:
This is a tough nut to crack. The size difference of the CI artifact is big (43MB vs 27MB), but I'm not really sure that's too relevant nowadays.

On the other hand, the KVS object and HTML docs are important, and as I have said in the past, I believe the path forward should be about getting rid of all the custom widgets, which means getting rid of the ircview, which is probably to be replaced by a webkit view. That's the way I see it.

Stanzilla:
Did you guys look around GitHub/Google etc to check how many scripts would be broken by switching without a compat layer?

wodim:
That would be a really bad metric because I assume most of the scripts ever written are never shared.

PR #2182, opened 2017-01-28
Make channel list dialogs more consistent; fix #1790

### LINKED ISSUE #1790: labels in edit/set text fields
Its should be obvious which field does what but only by typing in each one you find out what they are for.
Left is Channel
Right is Password

capture

We should design dialogs that are consistent and follow same standards applied in other dialogs.
FYI I tried deciphering how to do this myself, but After banging my head against https://github.com/kvirc/KVIrc/blob/ad7de10f2e041f0368afac5ab60257761df6b5cf/src/kvirc/ui/KviSelectors.cpp#L891-L898 I gave up.

ctrlaltca:
The "join channels" dialog and the "auto join channel list" settings dialog both presented a list of channels and some fields to add/remove channels from the list.
They were quite inconsistent as stated in#1790.
This PR tries to make them more consistent.
As a bonus, the "add" and "remove" buttons are now correctly enabled/disabled when they are useful.

schermata 2017-01-28 alle 18 28 19
schermata 2017-01-28 alle 18 28 52

### DIFF
--- a/src/kvirc/ui/KviSelectors.cpp
+++ b/src/kvirc/ui/KviSelectors.cpp
@@ -34,6 +34,7 @@
 #include <QPainter>
 #include <QLayout>
 #include <QColorDialog>
+#include <QGroupBox>
 #include <QPalette>
 #include <QFontDialog>
 #include <QAbstractItemView>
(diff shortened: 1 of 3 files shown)


IceN9ne:
The new layout looks good to me.  I think this makes the default width of the "join channels" dialog a bit wider, but I don't see that as a problem.

staticfox:
+1

Issue #2191, opened 2017-02-19
have Appveyor build standard and debug builds for windows?


wchristian:
In order to get useful backtraces debug builds are necessary. However the Windows build process is not automated for a general case, and right now is done so only for Appveyor. Short of automating the build process for users with scripting in the repo, a useful idea would probably have Appveyor generate normal AND debug builds.

As far as i can tell this should be doable by way of adding to the cmake line in appveyor.yml: https://github.com/kvirc/KVIrc/blob/master/.appveyor.yml#L109

As per https://github.com/kvirc/KVIrc/wiki/Grabbing-a-useful-backtrace -DDEBUG=1 should probably be added; and DarthGandalf suspects -DCMAKE_BUILD_TYPE=Debug as well.

wchristian:
@DarthGandalf This build lacks the *.pdb files containing the actual symbols. Do you think you can get those added?

    D:\KVIrc\Qt5Cored.pdb: Cannot find or open the PDB file.
    D:\KVIrc\Qt5Cored.pdb: Cannot find or open the PDB file.
    c:\work\build\qt5_workdir\w\s\qtbase\lib\Qt5Cored.pdb: Cannot find or open the PDB file.
    C:\WINDOWS\Qt5Cored.pdb: Cannot find or open the PDB file.
    C:\WINDOWS\symbols\dll\Qt5Cored.pdb: Cannot find or open the PDB file.
    C:\WINDOWS\dll\Qt5Cored.pdb: Cannot find or open the PDB file.
    SYMSRV:  c:\Symbols\Qt5Cored.pdb\41560C78A04540CABA4EFFC868CE75BD4\Qt5Cored.pdb - file not found
    *** ERROR: HTTP_STATUS_NOT_FOUND
    *** ERROR: HTTP_STATUS_NOT_FOUND
    *** ERROR: HTTP_STATUS_NOT_FOUND
    SYMSRV:  c:\Symbols\Qt5Cored.pdb\41560C78A04540CABA4EFFC868CE75BD4\Qt5Cored.pdb not found
    SYMSRV:  http://msdl.microsoft.com/download/symbols/Qt5Cored.pdb/41560C78A04540CABA4EFFC868CE75BD4/Qt5Cored.pdb not found
    srv*c:\Symbols*http://msdl.microsoft.com/download/symbols: Symbols not found on symbol server.

---

    D:\KVIrc\kvirc.pdb: Cannot find or open the PDB file.
[...]

PR #2192, opened 2017-02-19
Build also in debug mode for windows

DarthGandalf:
See #2191


Donat76:
This would be nice. I've had a crash that has to do with adding and deleting events out of the event editor. Sometimes KVIrc crashes a few minutes later but usually not. I've been trying to reproduce it consistently. No luck so far.

Dessa:
this is as good as it can get at this point, we need qt 5.6 for having pdb files for distribution.

Donat76:
The x86 release build still failed for some reason? Could you remind me why qt 5.6 isn't possible at this time? Thanks.

wodim:
That build failed because we download the spell checking dictionaries from sourceforge when we build, and the download failed for some reason.

Donat76:
Is there any way to get this pull request into the main branch so the debug builds are built with every commit?  I would like to do what  wchristian did and use VS 2015 and track down a crash I have been getting.  If I follow his directions, it should give the exact line number in the source code where it crashes.

wodim:
If you mean a random crash on Windows, it's probably the same...

Donat76:
It's crashing at random times but only after adding or deleting event handlers.

wodim:
In raw events? A crash was fixed just yesterday about that

Donat76:
No, not the raw event editor but the regular event editor.  Could be a similar bug in the event editor module? Is the debug build going to be added to the main branch at some time instead of just being a pull request.  The one debug build is a couple commits behind the main branch already.  I download the builds directly from the Appveyor, so an up to date debug build would help.

PR #2195, opened 2017-02-19
make the repaint timer in the ircview faster to decrease the lag that occurs when selecting text

wodim:
just a test for appveyor


wodim:
30 looks like a good cpu/lag compromise, everybody else ok?

DarthGandalf:
You found a good compromise for your CPU. Someone with a weaker CPU wouldn't be so happy.

Why there need to be a timer at all?

wodim:
My CPU peaks at 1% with this. My CPU is really beefy, but how much would that be in a core 2 duo or some other age old processor? 50% tops? 

And we have a timer because the ircview is crappy and if we didn't have a timer to limit the times per second the selection is calculated, it would completely eat the CPU, even mine. Not having a timer would require a significant rewrite.

Donat76:
This is off topic but I remember a conversation that revolved around QTWebKit and wanting to switch to QTWebEngine so the project could move to the QT 5.8 branch but the trade off of binary size and cpu cycles wasn't worth it. Or something similar to that. I think it's used for the help html files and other things? I know it can be disabled in the build but the community at QT.io are still keeping it up to date with the 5.8 branch. I have a MSYS2 build built with MinGW-W64 that I got through pacman. It's source is from here. https://github.com/annulen/webkit/releases/tag/qtwebkit-tp5

And the official community sources QTWebKit from QT.io
http://download.qt.io/community_releases/5.8/5.8.0-final/

It's not packaged in the official builds or source but it's still being actively maintained by the community since it's embedded in so many projects.
```

Downloads
10.4 MB  qtwebkit-tp5-qt58-darwin-x64.tar.xz
12.2 MB  qtwebkit-tp5-qt58-linux-x64.tar.xz
30.3 MB  qtwebkit-tp5-qt58-mingw530-x86.zip
27.3 MB  qtwebkit-tp5-qt58-msvc2015-x64.zip
21.9 MB  qtwebkit-tp5-qt58-msvc2015-x86.zip
11.8 MB  qtwebkit-tp5.tar.xz
Source code (zip)
Source code (tar.gz)
```

wodim:
We already bundle webkit, it's used for the help dialog. Also, you can use webkit from KVS to add custom webkit widgets to your scripts.

Our problem with webkit was, if I remember correctly, that it was simply dropped from Qt. The discussion is there somewhere. Many times I have stated that I'd like to see the ircview replaced with a webkit widget, like Textual does.

Webkit really makes kvirc bigger, but I can't quantify the difference.

wodim:
A user on IRC throttled his CPU down to 0.8 GHz and didn't experience a considerable increase in CPU usage while selecting text, so I'm merging this.

PR #2207, opened 2017-03-01
add only a direct shortcut to kvirc in the windows start menu

wodim:
Microsoft guidelines say that only shortcuts to the application itself should be shown in the start menu.

http://web.archive.org/web/20120305022332/http://msdn.microsoft.com/en-us/library/aa511447.aspx

>Put only program shortcuts on the Start menu. Don't put shortcuts to the following items on the Start menu:
>Program uninstallers. Users access uninstallers through the Programs control panel item.

Everybody ok?


Dessa (approved):
since we are past xp, this is probably ok

PR #2210, opened 2017-03-15
add export for all logs (channel, etc.)

AlexP11223:
This line was always commented (since initial impl in 184e489) but this feature seems to be working fine.

And it is very useful for someone who wants to export all old channel logs into txt without control codes. 
I almost implemented my own kvirc log exporter because of lack of that feature :)


Stanzilla:
@AlexP11223 would you be open to developing the `Choose directory` feature you mentioned?

Issue #2211, opened 2017-03-18
KVS Class Function Editor Reminder Text creates class build failure if ( ) are used in reminder text


Cabji:
### Expected behavior

Enter a reminder text using ( ) such as: "$0 - string to parse in some particular format (stringFormat)"
to remind dev later on to pass correctly formatted strings to the function.

### Actual behavior

On attempting class build, throws compilation errors: 
[KVS] Compilation error: Found character '$' (Unicode 0x24) where '{' was expected ...
[KVS]   22 	function epChange($0 = season/episode string (sXXeXX) $1 = change type ...

KVS is expecting a { after the (sXXeXX) - but why?
The class fails to build.

[//]: # (Besides the text description, include any screenshot(s) that help us visualize the issue you're facing)

### Steps to reproduce the issue
1. Create a class
2. Create a member function in the class
3. Use ( ) in the reminder text for the function
4. Try to build the class

### System information
[//]: # (Type "/about.kvirc" (or click "Help → About KVIrc") and copy&paste the details from the "Executable Information" tab below between ``` marks)
[//]: # (Also Enter any OS information the "Executable Information" tab doesn't contain, e.g. KDE, Gentoo, Linux Mint, etc.)

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Windows Release : 10.0.14393
System version: Windows 10 (x64) (Build 14393)
Architecture: Intel64 Family 6 Model 42 Stepping 7
Qt version: 5.5.1
Qt theme: windowsvista

Build Info:
Build date: 2016-12-12 22:31:43 UTC
Sources date: 20161212
Revision number: git-7042-g90a846a
System name: Windows-6.3.9600
CPU name: AMD64
Build command: C:/Program Files (x86)/CMake/bin/cmake.exe
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=C:/projects/kvirc/build/release
   Threads=Win32
Compiler name: C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/amd64/cl.exe
Compiler flags: /DWIN32 /D_WINDOWS /W3 /GR /EHsc
Qt version: 5.5.1
Features: IRC, IPv6, Crypt, SSL, IPC, Transparency, Webkit, Perl, Python, Enchant, Qt5, KVS
```
[...]

PR #2213, opened 2017-03-20
fix: hide comments in issues in new posts

ltc-data:
The _issue_ happens because comments can only have one ( open and ) close bracket. so ``` [//]:# (this is a comment (and it) shouldn't show up)``` shows up unless the inner brackets are escaped like ```[//]:# (this is a comment \(and it\) shouldn't show up)``` or there's no inner brackets or ```<!-- xml comments -->``` is used.

This is probably a regression introduced by GitHub

this should fix the issue that comments show up in tickets like #2211

#### Changes proposed
- remove inner brackets from comments
-
-

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,7 +4,7 @@
 ### Actual behavior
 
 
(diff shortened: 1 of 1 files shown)


staticfox:
Thanks!

PR #2218, opened 2017-04-24
Fix string type in ShellExecute for URL handlers on Windows

IceN9ne:
An attempt at fixing opening of utf-8 links when opened with setting "Use system URL handlers" enabled on Windows builds.


IceN9ne:
Doesn't work.

PR #2221, opened 2017-05-08
appveyor: Update MSVC to 2015 and Qt to 5.8.0

IceN9ne:


.appveyor.yml
```diff
@@ -93,7 +93,7 @@ install:
 build_script:
     - mkdir build
     - cd build
-    - '"c:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/vcvarsall" %vc_arch%'
+    - '"c:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/vcvarsall" %vc_arch%'
     - path
     - ps: |
         $zlibinc = c:\cygwin\bin\find /cygdrive/c/deps/ -path '*v120*/zlib.h'
```
DarthGandalf: You might need to change this to something like v140

IceN9ne: Is that what the `v120` stands for?  I assumed it was version 1.2.0 of `zlib`.

DarthGandalf: Yeh, this is an unsupported way to use nuget, the proper one is through visual studio GUI; I didn't know what is the format of filenames there, but needed to select a single file, without hardcoding too much.

Maybe there is a better way to use nuget from cmake-generated nmake...

Stanzilla:
do it!

ctrlaltca:
As a side note, Qt 5.8 is quite broken under osx and we had to revert using 5.7 in [another project](https://github.com/Cockatrice/Cockatrice/issues/2647).

PR #2228, opened 2017-06-12
Add ucrtbased.dll to the release directory for Windows debug builds.

### LINKED ISSUE #2227: Debug builds git-7115-g2c89201e3: Missing ucrtbased.dll Kivirc fails to launch
### Expected behavior
That the program would launch

### Actual behavior
Fails to Launch with error

```
The program cant start because ucrtbased.dll is missing from your computer, Try reinstalling the program
to fix this problem.
```

Tried reinstalling kvirc/msvc no joy

capture

[//]:# (Besides the text description, include any screenshots that help us visualize the issue you're facing)

1. upgrade Kvirc from  git-7103-g48fd3bc45 to  git-7115-g2c89201e3 on Windows 7
2. Start Kvirc post upgrade
3.

EDIT: I suspect https://github.com/kvirc/KVIrc/commit/5b14c711675aa807ccf27ea2f67cf5116872f210 broke this for Windows 7

### System information
[//]:# (Type ```/about.kvirc``` or click "Help → About KVIrc" and copy & paste the details from the "Executable Information" tab below between ``` marks)  
[//]:# (Also Enter any OS information the "Executable Information" tab doesn't contain, e.g. KDE, Gentoo, Linux Mint, etc.)

```
Program doesnt start
```
[//]:# (If you experienced crashes or segfaults please also include a stacktrace below, For how-to read: https://github.com/kvirc/KVIrc/wiki/Grabbing-a-useful-backtrace.)

IceN9ne:
Closes #2227

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,6 +5,7 @@ cache:
 environment:
     matrix:
         - cmake_build_type: Debug
+          arch: x64
           target_arch: x86_64
           ssl_arch: Win64
           nuget_arch: x64
@@ -17,6 +18,7 @@ environment:
           mingw_libgcc: libgcc_s_seh-1.dll
 
         - cmake_build_type: Debug
+          arch: x86
           target_arch: x86
           ssl_arch: Win32
           nuget_arch: Win32
@@ -29,6 +31,7 @@ environment:
           mingw_libgcc: libgcc_s_dw2-1.dll
 
         - cmake_build_type: Release
+          arch: x64
           target_arch: x86_64
           ssl_arch: Win64
           nuget_arch: x64
@@ -41,6 +44,7 @@ environment:
           mingw_libgcc: libgcc_s_seh-1.dll
 
         - cmake_build_type: Release
+          arch: x86
           target_arch: x86
           ssl_arch: Win32
           nuget_arch: Win32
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -5,6 +5,7 @@ cache:
 environment:
     matrix:
         - cmake_build_type: Debug
+          arch: x64
```
DarthGandalf, line 8: There are so many "arch"es already, please rename to something foo_arch, like below.

IceN9ne:
A question of whether licensing will be an issue was brought up in IRC.  That falls outside of my expertise, so leaving as is until it can be verified, since as far as I can tell, it accomplishes the original goal as is.

ctrlaltca:
Have a read at https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/ , especially point n. 6. At first M$ didn't want people to distribute the new dlls, but they had to give up on devs complaints.

ltc-data:
That point 6 is pretty clear, it says in no uncertain terms that you need to copy the necessary dlls with your app for all supported versions of windows. meaning yes, its fine to do this what PR rpoposes since the update from September 11, 2015 cancels all the previous restrictions i.e. the struck out text..

>>Updated September 11, 2015:  App-local deployment of the Universal CRT is supported.  To obtain the binaries for app-local deployment, install the Windows Software Development Kit (SDK) for Windows 10.  The binaries will be installed to C:\Program Files (x86)\Windows Kits\10\Redist\ucrt.  You will need to copy all of the DLLs with your app (note that the set of DLLs are necessary is different on different versions of Windows, so you must include all of the DLLs in order for your program to run on all supported versions of Windows).<s>App-local deployment of the Universal CRT is not supported. The Universal CRT is a Windows operating system component. Eventually (long-term), the Universal CRT will always be present on every machine, just like other operating system components today. We realize that there are today existing operating system versions where this component is not present, and we recognize that you, our customers will need to support these operating systems for some time. We hope that using the Windows Update package or static linking will suffice. One of the primary goals in our effort to refactor the CRT for this release was to mitigate the runtime proliferation problem, where over time computers end up with a large number of copies of the runtime libraries.</s>

PR #2237, opened 2017-06-29
defscript: add global to option controller titles

### LINKED ISSUE #2202: option controller window does not clarify it changes global options
The option controller window is opened by clicking a button in the channel option toolbar, which implies it works per channel, when in reality it works for all windows.

The window should have a note stating that changes there apply globally.

image

ltc-data:
should make clearer that this is a global controller thingy.

fixes #2202

#### Changes proposed
- added word **Global** to all title instances for the option controller
- if I read these files right the default.kvs is also bumped versions
-


data/defscript/popups.kvs
```diff
@@ -742,7 +742,7 @@ defpopup(windowpopup)
 
 	separator
 
-	item($tr("&Option Controllers","defscript"),$icon("options"))
+	item($tr("&Global Option Controllers","defscript"),$icon("options"))
```
IceN9ne: I think the `&` makes more sense to be on the `O` instead of the `G` for the options menu.

ltc-data: ok

Issue #2216, opened 2017-04-10
Prioritize tab completion by last spoken order


staticfox:
Requested by Ziginox on Rizon. Makes sense and doesn't seem too hard to do.

AndrioCelos:
Looking through the code, I notice that the user lists are implemented as what appear to be doubly linked lists, and tab completion steps through that list until it reaches the last found nickname.
To do this, we'd need to keep a separate list, sorted by the last message timestamp (which [KVIrc already keeps track of](https://github.com/kvirc/KVIrc/blob/master/src/kvirc/ui/KviUserListView.h#L162)), or rebuild that list each time `Tab` is pressed.
I'd prefer the former solution...

AndrioCelos:
Got the option added with [this commit](https://github.com/AndrioCelos/KVIrc/commit/cb1cd3a44e39e4485fc2b62971753cf7edc3d91c), but it's currently hidden. May I have help adding it to the options dialog?

wodim:
You can look at another implementations of combo boxes. We don't have any of our "magic" combo boxes, but a very similar thing is implemented in `OptionsWidget_ircView.cpp`, where it is constructed and added to the layout:

https://github.com/AndrioCelos/KVIrc/blob/cb1cd3a44e39e4485fc2b62971753cf7edc3d91c/src/modules/options/OptionsWidget_ircView.cpp#L56-L58

Then the option is read and the combo box is given a value:

https://github.com/AndrioCelos/KVIrc/blob/cb1cd3a44e39e4485fc2b62971753cf7edc3d91c/src/modules/options/OptionsWidget_ircView.cpp#L85-L98

And here the value of the combo box is saved if the user clicks apply/ok (in your case the implementation will be simpler): 

https://github.com/AndrioCelos/KVIrc/blob/cb1cd3a44e39e4485fc2b62971753cf7edc3d91c/src/modules/options/OptionsWidget_ircView.cpp#L121-L134

AndrioCelos:
I actually worked out the problem was KVIrc using the globally installed version of the options module, instead of the new one.
I appreciate your help, though, @wodim.

pragmaware:
Currently the code in  #2242 has some bugs that need to be addressed.

- Users that leave the channel remain in the completion list. They should be removed.
- Users that change their nickname should be removed from the completion list
- The completion list is not properly cleared when a message is sent. When a different completion is requested the old one is re-proposed.
[...]

PR #2242, opened 2017-07-07
Prioritize tab completion by last spoken order

AndrioCelos:
This pull request addresses #2216.

#### Changes proposed
- Expand standard nickname completion to allow re-ordering the matches. (bash-like completion is unaffected.)
- Fix a bug whereby a nickname completion made using `Shift`+`Tab` in standard mode was always treated as the last match.
- Add a new option, `uintNickCompletionOrder`, that governs how the matches are ordered. The options are as follows:
  - `0`: The same order as in the user list: by status, then alphabetical (the previous behaviour).
  - `1`: Alphabetical, A to Z.
  - `2` (default): By last action time, most to least recent.
- Add `/option` tooltips for more controls in the options dialog.

K2j5R8n

### DIFF
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -636,7 +636,8 @@ KviUIntOption g_uintOptionsTable[KVI_NUM_UINT_OPTIONS] = {
 	UINT_OPTION("ToolBarButtonStyle", 0, KviOption_groupTheme), // 0 = Qt::ToolButtonIconOnly
 	UINT_OPTION("MaximumBlowFishKeySize", 56, KviOption_sectFlagNone),
 	UINT_OPTION("CustomCursorWidth", 1, KviOption_resetUpdateGui),
-	UINT_OPTION("UserListMinimumWidth", 100, KviOption_sectFlagUserListView | KviOption_resetUpdateGui | KviOption_groupTheme)
+	UINT_OPTION("UserListMinimumWidth", 100, KviOption_sectFlagUserListView | KviOption_resetUpdateGui | KviOption_groupTheme),
+	UINT_OPTION("NickCompletionOrder", 2, KviOption_sectFlagNone)
 };
 
 #define FONT_OPTION(_name, _face, _size, _flags) \
--- a/src/kvirc/kernel/KviOptions.cpp
+++ b/src/kvirc/kernel/KviOptions.cpp
@@ -637,7 +637,7 @@ KviUIntOption g_uintOptionsTable[KVI_NUM_UINT_OPTIONS] = {
 	UINT_OPTION("MaximumBlowFishKeySize", 56, KviOption_sectFlagNone),
 	UINT_OPTION("CustomCursorWidth", 1, KviOption_resetUpdateGui),
 	UINT_OPTION("UserListMinimumWidth", 100, KviOption_sectFlagUserListView | KviOption_resetUpdateGui | KviOption_groupTheme),
-	UINT_OPTION("NickCompletionOrder", 2, KviOption_sectFlagNone)
+	UINT_OPTION("NickCompletionOrder", 2, KviOption_sectFlagInput)
 };
 
 #define FONT_OPTION(_name, _face, _size, _flags) \
--- a/src/kvirc/ui/KviUserListView.cpp
+++ b/src/kvirc/ui/KviUserListView.cpp
@@ -507,48 +507,39 @@ bool KviUserListView::completeNickStandard(const QString & szBegin, QString & sz
 			//		if(kvi_strEqualCI(entry->m_szNick.ptr(),c->currentNickName())
 			//	}
 
-			if(pEntry->m_szNick.length() >= szBegin.length())
+			if(pEntry->m_szNick.length() < szBegin.length()) continue;
+			// No match if it's shorter than the prefix entered.
+
+			bool bEqual = KviQString::equalCIN(szBegin, pEntry->m_szNick, szBegin.length());
+			if(!bEqual && KVI_OPTION_BOOL(KviOption_boolIgnoreSpecialCharactersInNickCompletion))
 			{
-				bool bEqual = KviQString::equalCIN(szBegin, pEntry->m_szNick, szBegin.length());
-				if(!bEqual && KVI_OPTION_BOOL(KviOption_boolIgnoreSpecialCharactersInNickCompletion))
-				{
-					QString szTmp = pEntry->m_szNick;
-					szTmp.remove(QRegExp("[^a-zA-Z0-9]"));
-					bEqual = KviQString::equalCIN(szBegin, szTmp, szBegin.length());
-				}
+				QString szTmp = pEntry->m_szNick;
+				szTmp.remove(QRegExp("[^a-zA-Z0-9]"));
+				bEqual = KviQString::equalCIN(szBegin, szTmp, szBegin.length());
+			}
 
-				if(bEqual)
+			if(bEqual)
+			{
+				// Matching entry; insert it into the list.
+				std::list<QString>::iterator it;
+				switch(KVI_OPTION_UINT(KviOption_uintNickCompletionOrder))
 				{
-					// Matching entry; insert it into the list.
-					std::list<QString>::iterator it;
-					switch (KVI_OPTION_UINT(KviOption_uintNickCompletionOrder))
-					{
-						case 0:  // As listed
-							m_CompletionList->push_back(pEntry->m_szNick);
-							break;
-						case 1:  // Alphabetical
-							for (it = m_CompletionList->begin(); it != m_CompletionList->end(); ++it)
-							{
-								if ((KviQString::cmpCI(pEntry->m_szNick, *it,
-									KVI_OPTION_BOOL(KviOption_boolPlaceNickWithNonAlphaCharsAtEnd)) < 0))
-								{
-									break;
-								}
-							}
-							m_CompletionList->insert(it, pEntry->m_szNick);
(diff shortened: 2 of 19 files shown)


src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
 	{
-		if(pEntry->m_szNick.length() >= szBegin.length())
+		szSkipAfter = *it;
+		szBuffer = *it;
+		if(bAppendMask)
 		{
-			bool bEqual = KviQString::equalCIN(szBegin, pEntry->m_szNick, szBegin.length());
-			if(!bEqual && KVI_OPTION_BOOL(KviOption_boolIgnoreSpecialCharactersInNickCompletion))
+			KviUserListEntry * entry = m_pEntryDict->find(*it);
+			if (entry)
```
IceN9ne: In the KVIrc code style, there is no space between `if` and the `(...)`.

src/kvirc/kernel/KviOptions.cpp
```diff
@@ -636,7 +636,8 @@ KviUIntOption g_uintOptionsTable[KVI_NUM_UINT_OPTIONS] = {
 	UINT_OPTION("ToolBarButtonStyle", 0, KviOption_groupTheme), // 0 = Qt::ToolButtonIconOnly
 	UINT_OPTION("MaximumBlowFishKeySize", 56, KviOption_sectFlagNone),
 	UINT_OPTION("CustomCursorWidth", 1, KviOption_resetUpdateGui),
-	UINT_OPTION("UserListMinimumWidth", 100, KviOption_sectFlagUserListView | KviOption_resetUpdateGui | KviOption_groupTheme)
+	UINT_OPTION("UserListMinimumWidth", 100, KviOption_sectFlagUserListView | KviOption_resetUpdateGui | KviOption_groupTheme),
+	UINT_OPTION("NickCompletionOrder", 2, KviOption_sectFlagNone)
```
IceN9ne: Consider using `KviOption_sectFlagInput` instead.

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+		for(KviUserListEntry * pEntry = m_pHeadItem; pEntry; pEntry = pEntry->m_pNext)
+		{
+			// FIXME: completion should skip my own nick or place it as last entry in the chain (?)
+
+			//	if(KviConsoleWindow * c = m_pKviWindow->console())
+			//	{
+			//		if(kvi_strEqualCI(entry->m_szNick.ptr(),c->currentNickName())
+			//	}
+
+			if(pEntry->m_szNick.length() >= szBegin.length())
```
IceN9ne: Consider using this instead and dropping the indention level of the rest of the code in the for statement.
`if(pEntry->m_szNick.length() < szBegin.length()) continue;`

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+					std::list<QString>::iterator it;
+					switch (KVI_OPTION_UINT(KviOption_uintNickCompletionOrder))
+					{
+						case 0:  // As listed
+							m_CompletionList->push_back(pEntry->m_szNick);
+							break;
+						case 1:  // Alphabetical
+							for (it = m_CompletionList->begin(); it != m_CompletionList->end(); ++it)
+							{
+								if ((KviQString::cmpCI(pEntry->m_szNick, *it,
```
IceN9ne: Extra set of parentheses are only needed with assignment in the conditional.  Also, extra space between `if` and `(...)`.

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+				{
+					// Matching entry; insert it into the list.
+					std::list<QString>::iterator it;
+					switch (KVI_OPTION_UINT(KviOption_uintNickCompletionOrder))
+					{
+						case 0:  // As listed
+							m_CompletionList->push_back(pEntry->m_szNick);
+							break;
+						case 1:  // Alphabetical
+							for (it = m_CompletionList->begin(); it != m_CompletionList->end(); ++it)
```
IceN9ne: In the KVIrc code style, there is no space between `for` and the `(...)`.
Also, consider using `auto it = std::find(...);` instead.

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+					QString szTmp = pEntry->m_szNick;
+					szTmp.remove(QRegExp("[^a-zA-Z0-9]"));
+					bEqual = KviQString::equalCIN(szBegin, szTmp, szBegin.length());
+				}
+
+				if(bEqual)
+				{
+					// Matching entry; insert it into the list.
+					std::list<QString>::iterator it;
+					switch (KVI_OPTION_UINT(KviOption_uintNickCompletionOrder))
```
IceN9ne: In the KVIrc code style, there is no space between `switch` and the `(...)`.

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+								if ((KviQString::cmpCI(pEntry->m_szNick, *it,
+									KVI_OPTION_BOOL(KviOption_boolPlaceNickWithNonAlphaCharsAtEnd)) < 0))
+								{
+									break;
+								}
+							}
+							m_CompletionList->insert(it, pEntry->m_szNick);
+							break;
+						default:  // Last action time
+							for (it = m_CompletionList->begin(); it != m_CompletionList->end(); ++it)
```
IceN9ne: In the KVIrc code style, there is no space between `for` and the `(...)`.
Also, consider using `auto it = std::find(...);` instead.

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+							{
+								if (pEntry->m_lastActionTime > (*m_pEntryDict->find(*it)).m_lastActionTime)
```
IceN9ne: In the KVIrc code style, there is no space between `if` and `(...)`.

src/kvirc/ui/KviUserListView.cpp
```diff
@@ -485,60 +487,115 @@ void KviUserListView::completeNickBashLike(const QString & szBegin, std::vector<
...
+								{
```
IceN9ne: Unnecessary {}.

src/modules/options/OptionsWidget_input.h
```diff
@@ -54,9 +54,14 @@ class OptionsWidget_inputLook : public KviOptionsWidget
...
 {
 	Q_OBJECT
+private:
+	QComboBox * m_pCompletionOrderCombo;
+
 public:
 	OptionsWidget_inputFeatures(QWidget * parent);
 	~OptionsWidget_inputFeatures();
+
+	virtual void commit();
```
IceN9ne: Consider using the modern form of `void commit() override;`.

src/kvirc/ui/KviUserListView.cpp
AndrioCelos: Looks like `std::find` checks whether a value is present in the list. That can be used later in the function, actually, but not in an insertion sort.
This part is searching for the first nickname that comes after the one being inserted.

src/modules/options/OptionsWidget_input.h
AndrioCelos: Should the other implementations of `KviOptionsWidget` be changed, then?

src/kvirc/ui/KviUserListView.cpp
IceN9ne: `std::find` allows you to use a lambda as a unary predicate to do a custom find, which performs the same action as your loop.

src/modules/options/OptionsWidget_input.h
IceN9ne: Eventually, yes.

src/kvirc/ui/KviUserListView.cpp
AndrioCelos: It does? I found only `std::find_if`.
http://www.cplusplus.com/reference/algorithm/find_if/

AndrioCelos: Is there precedent for how to format this?
```
it = std::find_if(m_CompletionList->begin(), m_CompletionList->end(), [this, pEntry] (QString n) -> bool {
	return (pEntry->m_lastActionTime > (*m_pEntryDict->find(n)).m_lastActionTime);
});
```

IceN9ne: it = std::find_if(m_CompletionList->begin(), m_CompletionList->end(),
        [this, &pEntry] (const QString & n) {
            return pEntry->m_lastActionTime > m_pEntryDict->find(n)->m_lastActionTime;
        });

pragmaware:
I'm reverting this because it has some problems. See my comment in #2216 .
The idea itself is good, but the algorithm probably needs some more thinking.

PR #2252, opened 2017-07-28
Fix invalid IP integer for DCC RSEND request

### LINKED ISSUE #2251: Reverse DCC: rsendExecute sends invalid IP parameter for ZeroPortRequests
When sending a Reverse DCC request with 0 port, rsendExecute sends out literally "127.0.0.1" as the IP parameter, which creates problems with some clients (for now, namely irssi) where then the IP gets parsed as part of the file name, the port (0) as the IP, the file size as the port and the token as the file size.

CRC-Mismatch:
Fixes #2251 

#### Changes proposed
- Instead of hardcoding "127.0.0.1" in the string message, since the IP is supposed to be sent in integer form and not all clients are able to parse the message correctly if it remains like so, I propose to send 0 or 2130706433 (which would be 127.0.0.1) instead.


IceN9ne:
Remove the comment and change it to the integer value for `127.0.0.1` and I'll merge it.

PR #2269, opened 2017-08-26
Add support for Python 3

wodim:
**Not for merging yet**


Arfrever:
What still needs to be done before support for Python 3 is added to KVIrc?
Python 2.7 will be discontinued by upstream in 9 months, on 2020-01-01. See https://pythonclock.org/.

wodim:
iirc this was completely functional, but I left it here because I wanted to implement a 2/3 toggle in cmake. Not necessary anymore it seems, right?

Stanzilla:
I guess not

wodim:
Can someone fix the cmake rules so it will only detect python if it's 3.x? The way it is now, I believe it will detect both 2 and 3 but it won't work with 2.

wodim:
craps out on windows 32 bit

src/modules/python/libkvipython.cpp
```diff
@@ -502,13 +502,39 @@ static bool python_kvs_fnc_isAvailable(KviKvsModuleFunctionCall * c)
...
+		$python.version
+	@description:
+		Returns which major version of Python is KVIrc linked to ([b]2[/b] or [b]3[/b])
+		or [b]0[/b] if Python is not supported at all.
+*/
+
+static bool python_kvs_fnc_version(KviKvsModuleFunctionCall * c)
+{
+#ifdef COMPILE_PYTHON_SUPPORT
+	c->returnValue()->setInteger(PY_MAJOR_VERSION);
```
Arfrever: Maybe value containing also at least the minor version of Python would be more useful.
Python headers, beside `PY_MAJOR_VERSION`, `PY_MINOR_VERSION`, `PY_MICRO_VERSION`, `PY_RELEASE_LEVEL`, `PY_RELEASE_SERIAL`, already provide `PY_VERSION_HEX`:
```
/* Version as a single 4-byte hex number, e.g. 0x010502B2 == 1.5.2b2.
   Use this for numeric comparisons, e.g. #if PY_VERSION_HEX >= ... */
#define PY_VERSION_HEX ((PY_MAJOR_VERSION << 24) | \
                        (PY_MINOR_VERSION << 16) | \
                        (PY_MICRO_VERSION <<  8) | \
                        (PY_RELEASE_LEVEL <<  4) | \
                        (PY_RELEASE_SERIAL << 0))
```

src/modules/python/libkvipython.cpp
```diff
@@ -502,39 +502,13 @@ static bool python_kvs_fnc_isAvailable(KviKvsModuleFunctionCall * c)
...
-}
-
 static bool python_module_init(KviModule * m)
 {
 	// register the command anyway
 	KVSM_REGISTER_SIMPLE_COMMAND(m, "begin", python_kvs_cmd_begin);
 	KVSM_REGISTER_SIMPLE_COMMAND(m, "destroy", python_kvs_cmd_destroy);
 
 	KVSM_REGISTER_FUNCTION(m, "isAvailable", python_kvs_fnc_isAvailable);
-	KVSM_REGISTER_FUNCTION(m, "version", python_kvs_fnc_version);
```
Arfrever: This seems like a potentially useful information to expose to consumers of this API.
Perhaps it should be called `python_version` instead of `version`.

Arfrever: I withdraw my suggestions.
Something like the following already works:
```
python.begin
import sys
kvirc.setLocal("python_major_version", str(sys.version_info[0]))
kvirc.setLocal("python_minor_version", str(sys.version_info[1]))
kvirc.setLocal("python_micro_version", str(sys.version_info[2]))
python.end
echo %python_major_version
echo %python_minor_version
echo %python_micro_version
```

Stanzilla:
@wodim yo sup?

wodim:
Still the same error, seemingly.

Stanzilla:
oh the cmake rule, right?

wodim:
I haven't got a clue, honestly.

DarthGandalf:
@wodim I haven't tested runtime, but the windows build at least doesn't confuse x86 and x64 libraries anymore.

wodim:
Seems to work fine. If it's ready (seeing that you added another commit) then merge. Thank you for your help.

DarthGandalf:
The last commit was to fix the debug build - it tried to link to **both** python38.lib and python38_d.lib

Issue #1959, opened 2016-04-29
Event handler names with spaces in name are mishandled depending on how they are added.


un1versal:
### Expected behavior

event handler names added via script/command or manually via GUI, with or without space in name) should be handled and display in event editor in a consistent and non buggy manner 
### Actual behavior

[know trigger] The GUI event editor strips spaces form the handlers names under some conditions
[know trigger] adding handlers via GUI event editor strips spaces form the handlers names

Handler names with spaces in name can behave and cause **eventctl** commands to behave erratically especially when such events names with spaces are _magically_ gone

Reference: https://github.com/kvirc/kvirc-scripts/issues/14
### Steps to reproduce the issue
- Add a handler manually via GUI event editor **notice** does not allow spaces in handler name once click **apply**
  ![capture](https://cloud.githubusercontent.com/assets/3521959/14913062/349da6da-0df8-11e6-9200-252a31d5c168.PNG)
- **Close event editor**
- add a event handler via command line in any input line with:
  `/event(OnURLLinkPopupRequest,added via command in input line) { popup.show urlpopup $0; }`
  **notice** the handler manually via command allows spaces in handler name
  ![capture](https://cloud.githubusercontent.com/assets/3521959/14913199/3b8ca9ea-0df9-11e6-9f21-9d78d8f682d0.PNG)
- Selecting via event editor (via GUI) **OnURLLinkPopupRequest** then selecting the handler named **`added via command in input line`** then selecting **OnURLLinkPopupRequest** again causes spaces in event name to disappear. 
  ![capture](https://cloud.githubusercontent.com/assets/3521959/14913298/f777372e-0df9-11e6-9bf0-76e6faeae219.PNG)
### System information

```
KVIrc 4.9.2 'Aria'

Runtime Info:
System name: Linux 3.13.0-85-generic
System version: #129-Ubuntu SMP Thu Mar 17 20:50:15 UTC 2016
Architecture: x86_64
Qt version: 5.5.1
Qt theme: fusion

Build Info:
Build date: 2016-04-29 06:51:37 UTC
Sources date: 20160428
Revision number: git-6544-gfa33e18
System name: Linux-3.13.0-85-generic
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   LIB_SUFFIX=/x86_64-linux-gnu
   Threads=POSIX
Compiler name: /usr/bin/c++
[...]

PR #2271, opened 2017-08-26
Clean up the names of event handlers

wodim:
This is a tentative fix for #1959. Since there was some heated discussion in that ticket I'm not committing it without a pr. I'm not sure it's the best solution but I think it's a step in the right direction.


Stanzilla (approved):

Arfrever:
Modified version of this patch was applied in commit 63e9e485a7a924187d13dc44a7b576acad569e9c.

PR #2273, opened 2017-08-26
cache the Strawberry dir

Stanzilla:
I actually wanted to cache the Chocolatey dir first but the appveyor config is a bit weird, in the logs it looked like it should use some sort of cached/already existing strawberry installation but that is never true on appveyor so it always downloads and installs and then patches it, this caching should make this faster.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -116,9 +116,9 @@ install:
         if (Test-Path "c:/spellcheck-dicts/zip/en_US.zip") {
             echo "using en dict from cache"
         } else {
-            (New-Object Net.WebClient).DownloadFile('http://sourceforge.net/projects/wordlist/files/speller/2017.01.22/hunspell-en_US-2017.01.22.zip', 'c:\spellcheck-dicts\zip\en_US.zip')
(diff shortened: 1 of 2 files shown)


ltc-data:
this was removed in https://github.com/kvirc/KVIrc/commit/ae97a8ffad491124f93be0d38798cbbc8a4af6ce

Dessa (commented):
if you are touching these anyway, you could also update to 2017.08.24

Dessa (commented):
IIRC the cache isn't even re-used for PRs, but i could be wrong here

Stanzilla:
Yeah certainly looks like it. How to best test if it does not then? New branch?

Stanzilla:
Yup, still too large 

```
Cache 'c:\Strawberry' - Uploading (151,131,111 bytes)...100%
Error uploading cache entry to the cache storage: The maximum cache size of 1,000 MB will be exceeded.
```

Stanzilla:
@Dessa will experiment with https://github.com/Microsoft/vcpkg next when he has time.

PR #2274, opened 2017-08-27
revert fast finish and let other builds attempt and likely build.

ltc-data:
```
matrix:
    fast_finish: true
```
Causes all other builds after a failed build to be cancelled, which means if the first build fails all other 3 are cancelled and no artifacts and or builds for Windows.

e.g. fast fail 
1st build failed all cancelled https://ci.appveyor.com/project/DarthGandalf/kvirc/build/0.0.0.0.1-branch-master-build-2377
e.g. no fast_build
1st build failed all others ran and finished OK https://ci.appveyor.com/project/DarthGandalf/kvirc/build/0.0.0.0.1-branch-master-build-2366

#### I looked for **practical examples** where first build failed (a debug build) and all release builds where OK

#### Changes proposed
- remove fast_fail strategy, and allow other builds to be attempted so we can have artifacts and usable builds.

The ONLY logical argument for this is, builds fail more often for external unrelated code changes than for actual bad code, Users that depend on appveyor builds for their updated binaries at least stand a better change of getting the usual binaries that work rather than nothing at all.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -51,8 +51,6 @@ environment:
           mingw_fourple: mingw32/mingw-w64-i686
           mingw_root: c:/msys64/mingw32
           mingw_libgcc: libgcc_s_dw2-1.dll
-matrix:
-    fast_finish: true
 install:
     - systeminfo
     - c:\cygwin\bin\uname -a


Stanzilla:
The thing is, we are waiting ages for appveyor builds, 2.5h+ per commit (4 jobs) and if a mistake sneaks in, it takes that time until you can verify that your afterwards committed fix worked. I agree that release builds should be compiled before debug though, which is why I am going to switch the order around.

ltc-data:
Actually no, if any code fails to build you can see it immediately on ANY failed build log. That doesnt take 2.5 Hour . case and point https://github.com/kvirc/KVIrc/commit/82f44408f63075903c6fd9ff1fb4765caf840d47 BAD code it took 9000 lines of log to fail

Asuccessful log has 18001 lines of log lines so: if 4 builds take 2.5 hours to build thats 37.5 minutes per build, https://github.com/kvirc/KVIrc/commit/82f44408f63075903c6fd9ff1fb4765caf840d47 failed half way so thats  18.75 minutes.,

2.5 hours to 18 minutes is not the picture you paint.

the problem is that in 100 failed builds (in 90) the issue is out of your control and you are essentially saying its OK for no builds at all to be built for some unreasonable change that has no value to the project IMHO.

Stanzilla:
I don't quite understand why you want a build of a commit that broke something though.

ltc-data:
The broken commit was an example to show you how your 2.5 hours is not a real time and your argument is not based in real life.

## the POINT is today there are no builds whatsoever including the fix for that broken commit you just mentioned because of the fast_fail strategy

NO BUILDS TODAY of any consequence because of this fast_fail nonsense.

Stanzilla:
I get that my time estimation was wrong, still don't know why you are so angry. The build system still needs more work because it still takes ages. the fast_fail strategy does speed it up in the case of the first build already having a bug that needs fixing anyway so time is not wasted on the other 3 jobs that would (probably) run into the same problem anyway.

Dessa:
the only real solution here is to have own prebuilts. i was planning on having that but i got kind of sidetracked by broken hardware

ltc-data:
@Stanzilla Im not angry, Im frustrated because there are no builds including the fix after the broken commit and you are being utterly unreasonable and inflexible on something you are plain wrong in changing.

again it doesnt take ages, 38 minutes per build is not ages,

@Dessa no sorry, it was perfectly fine so far without this nonsense.

We could have already had release builds today, but thats OK because none is better in YOUR OPINION.

Dessa:
nope, it wasn't really fine.
choco likes to fail randomly when the cache gets invalidated, it usually fixed itself in a few tries, also SF used to have the same problem.

using own prebuilts gets around that, maybe they can made smaller so they can be cached, maybe something else as i really didn't wrap my head around it yet for now probably obvious reasons.

fast_finish gives the problem more exposure yes but the reason for its usage isn't really to take the load off the build server, but to expose working builds to the user faster; if there is another aspect which makes them fail it needs to be fixed instead but as said im not in a position to do so at the current time

ltc-data:
The proof is in the pudding as they say, this commit is the only successful build today.

Perhaps this fast_fail was committed at a bad time, Im not arguing that it needs to be done, but lets see how many days have to go by with no windows builds until someone decides to agree this was the wrong time to do it.

Like you say, you havent the time now, and the problem is in the immediate time.

What the hell is the problem with fixing the now if it works now?

Stanzilla:
Failure is because Appveyor has a problem atm https://appveyor.statuspage.io/incidents/2tjbzx1mtnz8

wodim:
Instead of disabling `fast_finish`, we should do as Dessa says.

ltc-data:
when will that be? my weekend ends today Ive no time, patience or inclination to deal with it during the week. So thank you for denying me a build when I needed one.

Stanzilla:
You make zero sense, if you wanted a build that badly, you could just have used the one from your pull request here?

wodim:
Builds rarely fail. I think it makes sense to keep `fast_finish`. If someone commits something that doesn't compile at all we don't want to wait until it's finished failing four times.

A recommendation if you need to test the last commit of KVIrc is to switch to a Linux distro, where you can easily build this software from git. For example, you could use Linux Mint.

PR #2279, opened 2017-08-28
Raise the main window when an irc link is opened from a browser

wodim:
Just a test to see if this compiles on Windows


wodim:
ty fast_finish

PR #2282, opened 2017-08-28
Improve the script tester a little bit

### LINKED ISSUE #1758: Add a dedicated output window for the script tester
AFAIK, if you run anything such as `echo "whatever"`, `whatever` will be printed to the first(?) console window. Perhaps maybe we should add a dedicated output window within the script tester itself?

EDIT: By window I mean something in the same pane. Like, maybe a buffer where my wonderfully drawn red box is:

wodim:
* Now the code editor is focused when the script tester is open, instead of the find bar. I have also added a find button to all code editors, and I have fixed the colour of the "find" qlineedit, which has been broken since forever.

* A dedicated ircview has been added for the script tester, which is the default target for `echo`/etc. This would close #1758

### DIFF
--- a/src/modules/codetester/CodeTesterWindow.cpp
+++ b/src/modules/codetester/CodeTesterWindow.cpp
@@ -75,8 +75,7 @@ CodeTesterWindow::CodeTesterWindow()
 
 	m_pIrcView = new KviIrcView(m_pSplitter, this);
 
-	QList<int> li;
-	li << width() / 2 << width() / 2;
+	QList<int> li { width() / 2, width() / 2 };
 	m_pSplitter->setSizes(li);
 }
 
--- a/src/modules/codetester/CodeTesterWindow.cpp
+++ b/src/modules/codetester/CodeTesterWindow.cpp
@@ -34,6 +34,7 @@
 #include "KviConsoleWindow.h"
 #include "KviKvsScript.h"
 #include "KviKvsVariantList.h"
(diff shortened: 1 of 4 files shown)


src/modules/codetester/CodeTesterWindow.cpp
```diff
@@ -61,36 +70,30 @@ CodeTesterWidget::CodeTesterWidget(QWidget * par)
 	m_pParams = new QLineEdit(this);
 	m_pParams->setToolTip(__tr2qs_ctx("Here you can specify a semicolon-separated list of parameters that will be available in the code as $0, $1, $2, ..", "editor"));
 	g->addWidget(m_pParams, 1, 2);
+
+	m_pSplitter->addWidget(l);
+
+	m_pIrcView = new KviIrcView(m_pSplitter, this);
+
+	QList<int> li;
+	li << width() / 2 << width() / 2;
```
IceN9ne: Can be reduced to `QList<int> li{width() / 2, width() / 2};`

the-j0k3r:
A challenge I face regularly is to find next / previous the find and replace popup is also not fantastic.
Glad to see someone is finally trying to improve these.

wodim:
An ircview for the script tester is the kind of thing that should've been there since day one tbh

wodim:
Merging because nobody complained. Feel free to comment or revert if something's wrong.

Issue #1855, opened 2016-01-27
Tabs in the IrcView not displayed properly


wodim:
![2016-01-27_15-34-21](https://cloud.githubusercontent.com/assets/1001553/12616455/8cb2a572-c50b-11e5-8284-5b2a641f2ce9.gif)

pls fix

un1versal:
https://github.com/kvirc/KVIrc/wiki/Submitting-a-bug-report#information-required-when-making-a-bug-report-via-github-issues

lol

un1versal:
Works fine here in trunk, if still and issue provide build information and reopen

un1versal:
[14:59:59]  <wodim>     yes hello   there
[15:00:04]  <wodim> click that
[15:00:04]  <wodim> /say $char(9)yes$char(9)hello$char(9)there
[15:00:07]  <staticfox> ok yeah that is broke

un1versal:
This issue is also visible present with some aliases Usage string

type 
1) `/usermodename`

2) then select the text below Usage:

![tabs](https://cloud.githubusercontent.com/assets/3521959/16673099/014cc27a-44a3-11e6-939a-a3525ae4e4bd.gif)

and clicking the test above

![tabs](https://cloud.githubusercontent.com/assets/3521959/16672770/338e70f6-44a0-11e6-8847-3c82cb62e49e.gif)

and this is bug #2051 (I know #1855 != #2051 its not same bug, but the way it presents itself is somewhat similar)
![unicode](https://cloud.githubusercontent.com/assets/3521959/16672865/d2485e14-44a0-11e6-91f2-ef2c3c163cc9.gif)

sloanonlinux:
I has this bug too, version: 4.9.2

wodim:
When we print text in the ircview, we make one call to QPainter.drawText for every line. If the text has tabs, the tabs are wrapped at regular intervals. You can check this writing an alias like this:

    echo $char(9)yes$char(9)hello$char(9)there
    echo $char(9)yes$char(9)hellofriend$char(9)there

![image](https://user-images.githubusercontent.com/1001553/29744330-426ab450-8aa3-11e7-8d0b-ccc6529dfc4d.png)

When you select text, on the other hand, three calls to drawText are made: one for the text before the selected block, another for the text in the selected block, and another for the text after the selected block. This means there are three different origin points for the text drawn with drawText (one for each chunk of the line), which in turn means the tabs wrap at three different places now.

[...]

PR #2285, opened 2017-08-29
Substitute tabs with spaces before they are painted.

wodim:
#1855

There's like one thousand places in the code where you could put this. Every place has its complications. Whatever you do you feel guilty knowing you're ruining the performance.

### DIFF
--- a/src/kvirc/ui/KviIrcView.cpp
+++ b/src/kvirc/ui/KviIrcView.cpp
@@ -1136,6 +1136,8 @@ void KviIrcView::paintEvent(QPaintEvent * p)
 	// And loop through lines until we not run over the upper bound of the view
 	while((curBottomCoord >= KVI_IRCVIEW_VERTICAL_BORDER) && pCurTextLine)
 	{
+		pCurTextLine->szText.replace('\t', ' ');
+
 		// Paint pCurTextLine
 		if(maxLineWidth != pCurTextLine->iMaxLineWidth)
 		{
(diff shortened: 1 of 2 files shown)


IceN9ne:
Consider doing this in a different place where the lines are being added so that it only needs to happen once per line and not on every paint.  Maybe `KviIrcView::appendLine` or before it gets here.

I can't think of a reason tabs would need to be preserved unless it somehow allows the tabs to be preserved when you highlight-copy.

Also, we have a configuration for how many spaces a tab is replaced with for input, maybe do this for output as well?  Use same setting or new setting?

wodim:
What about this steaming pile of shit?

I really don't know what to do. I can't do the replacement later because I'd have to recalculate all the places where the length of the string and the chunks and whatever are stored. I can't do the replacement before (when it still is a kvi_wchar_t) because everything is marked const until the origin, and then I'd have to do the replacement in several places.

the-j0k3r:
How about adding a default event script to [b]OnTextInput[/b] that does a string replacement of tabs and accounts for **option uintSpacesToExpandTabulationInput** userset value

Or did I misunderstand the problem?

IceN9ne:
`OnTextInput` is when the user types something in. This PR is for when you receive messages from other users that have tabs in them and are painted to the `IrcView` window causing odd display issues.

the-j0k3r:
There are OnChannelMessage and OnQueryMessage also to handle incoming if that's the problem. In either case, Im not volunteering :)

Issue #2295, opened 2017-09-01
Windows: Path to custom icon popup entry with single slash OK until context menu called.


the-j0k3r:
theres an issue with single slashes in paths in Windows with custom icons in popup entries. The slash is stripped when you trigger the popup

so the act of calling the context menu is the issue?

e.g.
```popup.addExtPopup querytextview popupname "Menu Entry" $file.localdir("pics/icon_small.png") popupnames```

In linux that adds OK to the popup icon field

home /username/.config/KVIrc/pics/icon_small.png

but in Windows it adds as

E:\Portable\KVIrc\Settings\pics\icon_small.png the issue is that when bringing up the popup it shows a warning,

#### It is important to make the stressed point that if you open the popup editor the image is shown just fine with the path containing a single slash, the slash is stripped only when you call the relevant context menu in that case **querytextview**

manually adding a second slash to the icon entry ipath in popups like E:\\Portable\\KVIrc\\Settings\\pics\\icon_small.png makes the issue go away. but you cant insert it automatically.

even 

```
popup.addExtPopup querytextview popupname  "Menu Item Label" $file.fixpath($file.localdir("//pics//icon_small.png")) popupname
```

[21:32:33] Can't find the icon 'E:PortableKVIrcSettingspicsicon_small.png': ignoring
```
### Expected behavior
this to work without fuss?

### Actual behavior
[21:32:33] Can't find the icon 'E:PortableKVIrcSettingspicsicon_small.png': ignoring

[//]:# (Besides the text description, include any screenshots that help us visualize the issue you're facing)

### Steps to reproduce the issue
1. as above
2.
3.

### System information
[//]:# (Type ```/about.kvirc``` or click "Help → About KVIrc" and copy & paste the details from the "Executable Information" tab below between ``` marks)  
[//]:# (Also Enter any OS information the "Executable Information" tab doesn't contain, e.g. KDE, Gentoo, Linux Mint, etc.)

```

```
[//]:# (If you experienced crashes or segfaults please also include a stacktrace below, For how-to read: https://github.com/kvirc/KVIrc/wiki/Grabbing-a-useful-backtrace.)

the-j0k3r:
Thanks, will test asap and have alerted the parties concerned to this update. Once again  thanks.

the-j0k3r:
Im sorry but its still broken.

So I run this..

[...]

PR #2299, opened 2017-09-04
[libkvifile docs] $file.localdir should document $escape due to popupadd.ExtPopup usecase slash stripping

the-j0k3r:
See discussion in #2295

#### Changes proposed
- add documentation to $file.localdir reference to $escape

### DIFF
--- a/src/modules/file/libkvifile.cpp
+++ b/src/modules/file/libkvifile.cpp
@@ -1236,7 +1236,7 @@ static bool file_kvs_cmd_writeLines(KviKvsModuleCommandCall * c)
 		[example]
 			echo KVIrc looks for pictures in $file.localdir(pics)
 			echo panic.png would be translated to $file.localdir(pics/panic.png)
-			echo panic.png would be translated to $escape($file.localdir(pics/panic.png
+			echo panic.png would be translated to $escape($file.localdir(pics/panic.png))
 		[/example]
 	@seealso:
 		[fnc]$escape[/fnc]
--- a/src/modules/file/libkvifile.cpp
+++ b/src/modules/file/libkvifile.cpp
@@ -1219,7 +1219,7 @@ static bool file_kvs_cmd_writeLines(KviKvsModuleCommandCall * c)
 		Returns the path to the KVIrc local data directory.
 		The KVIrc local data directory is always writable and contains
 		the various subdirectories that KVIrc uses internally: audio, avatars,
-		config, help, incoming, log, modules, msgcolors and pics.
+		config, help, incoming, log, modules, msgcolors, events, aliases, actions and pics.
 		[br][br]
 		If <relative_path> is passed, then it is appended to the local directory path.
 		The path is adjusted to contain single separators suitable for the platform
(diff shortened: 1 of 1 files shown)


src/modules/file/libkvifile.cpp
```diff
@@ -1216,19 +1216,30 @@ static bool file_kvs_cmd_writeLines(KviKvsModuleCommandCall * c)
...
+		you should escape the path [b]$escape($file.localdir(pics/panic.png))[/b]
+		this is multiOS friendly and adds a custom icon path to Windows that is not broken
+		when KVS parses the context menu entry due to Windows paths being backslashed,
+		which are special characters in KVS that require escaping.[br]
+		Overall [b]$escape($file.localdir(pics/panic.png)) is a more correct approach.
 	@examples:
 		[example]
 			echo KVIrc looks for pictures in $file.localdir(pics)
 			echo panic.png would be translated to $file.localdir(pics/panic.png)
+			echo panic.png would be translated to $escape($file.localdir(pics/panic.png))
```
wodim: Why both lines, without `$escape` and with it?

the-j0k3r: I dont know, maybe because you dont need to use with in every case.

wodim: You *need* to use it in every case, unless you want to check if you're on Windows or on Linux before adding a popup menu, which is something you should not do.

The best approach once again is to use `\$` because that way you can move your configuration folder between OSes and have it working, but otherwise `$escape` is a must.

the-j0k3r: Thats not what I found at all and I and others have tested this,

YOU dont need to use it in every case even if OS's are different, and when adding a popup you dont need to check either which docs say clearly what the exception is.

what you suggest is broken in some cases, and I wont add more broken docs.
feel free to close.

Stanzilla: That sounds like a un1versal solution.

the-j0k3r: > That sounds like a un1versal solution.

This is the second time youre making some implications that are false.
I know the person know here as un1versal, he was a very good friend of my father.

Not that, I owe you or anyone any explanations, but un1versal was a good person and did nothing afaik but help. He's passed away now for some time and so has my father, and unless you have something against me (which escapes me because I dont know you or you me), I have nothing else to say.

Stanzilla: I don't have anything against you, I just don't believe the story.

the-j0k3r: I have nothing to prove to you or anyone and have done nothing to you or this project but try to help a friend of my father with some of his projects, your implications are already a waste of time.

Your attitude here of making petty reports to Github are a waste of my time and hinder contributions. If you dont want my contributions just close them,

Stanzilla: I didn't report anything to GitHub and let's leave it at, we would not have a problem if un1versal returned, no need for games.

the-j0k3r: Tread carefully, he was a friend of my fathers, perhaps his best and oldest friend, dont disrespect the dead by making false implications.
I have great respect for the man he was and what he stood for, hes passed away long ago so let that rest already.

the-j0k3r:
@wodim, is that better? Im not adding ```\$escape``` mention because as test showed it breaks popup editor icon not showing for one. this solution works afaik everywhere even if not strictly needed in all cases.

ltc-data:
@the-j0k3r dont waste your time here, the people in charge here dont deserve your or my or anyones help.

fyi, turn on your phone.

ltc-data:
Your turn. Im out of this pit.

PR #2301, opened 2017-09-06
Make the issue/PR templates much simpler

wodim:
I think the current issue/PR templates are very complex. I believe some people might feel intimidated by them, fearing the tickets will be closed or they'll be yelled at if they don't follow the template to the lettter. A bad bug report is better than no bug report at all and if there's some missing info we can always ask for it.


ltc-data:
Whats worst? Making trivial pull requests that are left to fester and people feel ignored and unappreciated (like I do). 
Or making people feel unwelcome on irc?
Or accuse people on GitHub of being liars without reason?

wodim:
Faking your own death.

ltc-data:
hahahaha troll

I rest my case! Im out of here and will get my mates out too.

wodim:
Merging because nobody complained.

Issue #2300, opened 2017-09-06
No indication of CTCP request being made


wodim:
When you make a CTCP request, there's no indication anywhere of it. Not sure if a bug or a feature request.

IceN9ne:
I'd say it's intentionally this way. I can't think of a CTCP request that I'd want to see that isn't echoed. You can always see it in the Socket Spy window if you want to debug.

If you were going to add CTCP request echoing, it'd need to be configurable as to which types of CTCP messages are echoed.

PR #2303, opened 2017-09-10
show ctcp message on send

ltc-data:
there are no -i=$msgtype for the ctcp except ping, so am using that genericstatus which looks ugly and cant use $icon(name) in these which is a stupid limitation.

#### Changes proposed
-  show ctcp message on send
-
-

### DIFF
--- a/data/defscript/default.kvc
+++ b/data/defscript/default.kvc
@@ -7,6 +7,6 @@ AddonVersion=4.1.1
 AliasVersion=4.1.1.6923
 ClassVersion=4.1.1
 EventVersion=4.1.1.7125
-PopupVersion=4.1.1.7126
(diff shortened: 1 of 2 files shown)


wodim:
If this is in response to #2300, you know this is a horrible idea and not what I want.

If there's no "CTCP sent" message type then the solution is to add it.

ltc-data:
idk what you want then, I dont understand what else should happen, but this does the job and is what im prepared to do for this with the time I have.

ltc-data:
failure is caused by https://github.com/kvirc/KVIrc/blob/master/CMakeLists.txt#L76

ltc-data:
Thank wodim for and his fanboys for driving contributors away.

PR #2311, opened 2017-09-28
Fix typo in manual & bug in alias

solmon4devx:
The function' objects. clearObjects' is called in the manual and in the alias clearobjects as' objects. clear'


wodim:
gcs

solmon4devx:
a ti wodim :)

PR #2325, opened 2017-11-11
AppVeyor: update Qt to 5.9, upgrade to VS2017

craftwar:


Stanzilla:
Hey @craftwar thanks for the PR. For starters, are you done with this PR yet? Looks like you missed updating the x86 branch configs. I'll mark it as `in progress` for now.

craftwar:
@Stanzilla Do you mean x86 config "qt_ver: 5.9\msvc2015"? It's intended. There is no official Qt 5.9 32bit VS2017 lib. They say use VS2015 (binary compatible).

Stanzilla:
oh, hrm.

.appveyor.yml
```diff
@@ -124,7 +125,7 @@ install:
 build_script:
     - mkdir build
     - cd build
-    - '"c:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/vcvarsall" %vc_arch%'
+    - '"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars%ssl_arch:~3%.bat"'
```
DarthGandalf: Let's not mess with variables this way. What's wrong with `%vc_arch%`?
The matrix has different _arch variables because they have different formats.

.appveyor.yml
```diff
@@ -24,7 +25,7 @@ environment:
           nuget_arch: Win32
           vc_arch: amd64_x86  # cross-compile from amd64 to x86
           choco_arch: --x86
-          qt_ver: 5.8\msvc2015
+          qt_ver: 5.9\msvc2015
```
DarthGandalf: A comment would help

DarthGandalf (changes requested):

.appveyor.yml
craftwar: I just follow appveyor doc. There are only these two .bat files, no more .bat accepting arch argument. (I checked in rdp session)

DarthGandalf: But you are still abusing `ssl_arch` for non-SSL related stuff. After this change `vc_arch` is not used anymore.
What about changing values of `vc_arch` to match its new format?

DarthGandalf:
By the way, when saying something like "following the doc", please provide the link. Otherwise it's very hard to understand *which* doc you're following (or where the answer is)

craftwar:
Hope no problem anymore. I am too tired these days.

DarthGandalf:
Erm. So now you're saying there is such vcvarsall.bat file which accepts
that argument?

2017-11-30 10:23 GMT+00:00 craftwar <notifications@github.com>:

ctrlaltca:
You could let cmake handle the problem (example for msvc2015):
```
# appveyor.yml - build matrix section
cmake_generator: Visual Studio 14 2015 Win64
cmake_toolset: v140,host=x64
# or, for 32bit:
cmake_generator: Visual Studio 14 2015
cmake_toolset: v140_xp # use the windows XP compatible toolset
# build script section
cmake .. -G "$env:cmake_generator" -T "$env:cmake_toolset"
```

craftwar:
Either I am too tired (sleep only 4hours these days), miss that or they are different on different build servers.
That's why I keep appveyor style and doc for backup.

cmake/KVIrc.nsi.cmake
```diff
@@ -55,7 +55,7 @@ Var LocalDir
 
 ;--------------------------------
 !insertmacro MUI_RESERVEFILE_LANGDLL
-ReserveFile "${NSISDIR}\Plugins\UserInfo.dll"
+ReserveFile "${NSISDIR}\Plugins\x86-unicode\UserInfo.dll"
```
DarthGandalf, line 58: If this needed for both x86 and x86_64?

.appveyor.yml
```diff
@@ -124,7 +125,8 @@ install:
 build_script:
     - mkdir build
     - cd build
-    - '"c:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/vcvarsall" %vc_arch%'
+#    - '"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars%ssl_arch:~3%.bat"'  #according to https://www.appveyor.com/docs/lang/cpp/#visual-studio-2017
```
DarthGandalf, line 128: please remove

cmake/KVIrc.nsi.cmake
craftwar: It's needed on appveyor VS2017 image. If this breaks compatibility on other platform. I can create link on appveyor.

DarthGandalf:
@ctrlaltca that's possible, but that also means switch from nmake to msbuild, which need another way to tell it location of dependencies... does msbuild support nuget better now than before?

ctrlaltca:
Well, once the dependencies are on disk, you can instruct cmake on where to pick them up.
As an example, here's the config for another project based on qt: [appveyor.yml](https://github.com/Cockatrice/Cockatrice/blob/master/.appveyor.yml); i dropped all the chocolatey and nuget usage, since they tend to bring it dependencies on different vcredist runtimes.. hope it can help.

Issue #2335, opened 2017-12-21
AppVeyor failing: "Cannot find Visual Studio release redistributable files"


wodim:
>Command executed with exception: Warning: Cannot find Visual Studio release redistributable files in C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\redist\MSVC\14.12.25810.

https://ci.appveyor.com/project/DarthGandalf/kvirc/branch/master/job/irto9u17d6ce9pa5

craftwar:
WinDeployQt bug

https://bugreports.qt.io/browse/QTBUG-65015
https://i.imgur.com/7sK3Hd6.png

DarthGandalf:
@craftwar any idea why it succeeded on #2325 ?

craftwar:
@DarthGandalf newer VS2017 changes vcredist file name as I write in last reply.

DarthGandalf:
Let's keep this open until the workaround is removed/fixed properly.

craftwar:
![image](https://user-images.githubusercontent.com/16416710/35477669-53d92300-0403-11e8-9b34-b6275a3ac554.png)

VS2017 15.5.4 on my computer. This can be closed.

IceN9ne:
The temporary fixed was reverted yesterday in 8458d1b

PR #2336, opened 2017-12-22
Workaround to vc_redist #2335

craftwar:
This is not used as permanent solution. windeployqt is fixed in next Qt version.


DarthGandalf:
nsis would probably need to check existance of both files, because it doesn't know which one it should run

PR #2337, opened 2017-12-28
run CodeMaid

Stanzilla:
Ran CodeMaid (http://www.codemaid.net/)

Open for discussion on options/style


wodim:
I vote no because the current style is not bad enough to justify destroying the entire history.

Stanzilla:
CodeMaid can be adapted to the current style and then can apply it consistently. I just need to find the correct options

DarthGandalf:
What's wrong with clang-format?
P.S. When I've run clang-format on the whole KVIrc, I also tuned its settings to match existing style, more or less.

Stanzilla:
Nothing, I'm just playing around

PR #2346, opened 2018-01-26
KviModeWidget: Fix two bugs and refactor class

IceN9ne:
* Bug fix: Unable to change previously set key in one step
    * Bug fix: Unable to set mode that has a parameter when no parameterless modes are set
    * Made the mode changing more intuitive for the user

Let me know if the style and changes look fine and I'll merge.

### DIFF
--- a/src/kvirc/ui/KviChannelWindow.cpp
+++ b/src/kvirc/ui/KviChannelWindow.cpp
@@ -103,9 +103,9 @@ KviChannelWindow::KviChannelWindow(KviConsoleWindow * lpConsole, const QString &
 	connect(m_pTopicWidget, SIGNAL(topicSelected(const QString &)),
 	    this, SLOT(topicSelected(const QString &)));
 	// mode label follows the topic widget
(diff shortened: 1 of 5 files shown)


IceN9ne:
I'm not happy with the length of `KviModeWidget::processModeChanges`, but it's much shorter and cleaner than it was.  C++17 structured bindings would clean it up even more when we get to that, but it should be able to be reduced further.  I'm considering breaking out the mode string building into it's own function and doing this in `processModeChanges`:

```
for(const auto & szLine : buildModeChangeCommands(removeModes, addModes, iModes))
    emit setMode(szLine);
```

with these added to the header:

```
using ModeList = std::vector<std::pair<QChar, QString>>;
QStringList KviModeWidget::buildModeChangeCommands(const ModeList & removeModes, const ModeList & addModes, const int iMaxModeChangesPerCommand) const;
```

Thoughts?  I know this is an infrequently used and minor widget, but it's still part of our codebase.

PR #2351, opened 2018-02-19
CMakeLists: Detect and use libenchant 2.x

### LINKED ISSUE #2350: Support libenchant 2.0
A new major version of the enchant spell-checking library is out, with an incompatible ABI:

https://github.com/AbiWord/enchant/releases/tag/v2.0.0

Distributions such as Arch Linux have already moved on to version 2.

Though version 1 [can be installed side-by-side with version 2](https://github.com/AbiWord/enchant/issues/168), Arch Linux does not package it.

It would be nice if KVIrc supported version 2 directly (i.e. detected its presence, enabled it during configuration, and used the library). I think supporting both 1.x and 2.x should not be too much of a burden, in the same way we support OpenSSL pre- and post-1.1.0.

CyberShadow:
Fixes #2350.

Looks like no source code changes are needed in KVIrc - just including/linking against the new library is enough.


CMakeLists.txt
```diff
@@ -780,8 +780,14 @@ option(WANT_SPELLCHECKER "Compile spell-checker support" ON)
 if(WANT_SPELLCHECKER)
 	find_package(PkgConfig)
 	pkg_check_modules(Enchant enchant)
-	if(Enchant_FOUND)
-		set(CMAKE_STATUS_SPELLCHECKER_SUPPORT "Yes")
+	pkg_check_modules(Enchant2 enchant-2)
```
DarthGandalf: Why not `pkg_search_module(Enchant enchant enchant-2)` instead of this whole change?

CyberShadow: Good idea! I thought we'd need to do some source modifications, but it doesn't seem to be the case.

PR #2360, opened 2018-03-23
doc: Allow to override build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

### DIFF
--- a/admin/gendoc.pl
+++ b/admin/gendoc.pl
@@ -28,8 +28,8 @@
 # GLOBAL CONFIGS
 #################################################################################################
 
-$g_currenttime=gmtime;
-$g_currentuser = getlogin || getpwuid($<) || "Unknown";
+$g_currenttime = gmtime($ENV{SOURCE_DATE_EPOCH} || time);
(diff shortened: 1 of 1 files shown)


admin/gendoc.pl
```diff
@@ -28,8 +28,9 @@
 # GLOBAL CONFIGS
 #################################################################################################
 
-$g_currenttime=gmtime;
-$g_currentuser = getlogin || getpwuid($<) || "Unknown";
+$g_currenttime = gmtime($ENV{SOURCE_DATE_EPOCH} || time);
+$g_currentuser = ($ENV{SOURCE_DATE_EPOCH} && "reproducible") ||
```
DarthGandalf: I think it should have a separate config knob for "reproducible build"; making username depend on some unrelated environment variable looks less than obvious

bmwiedemann: OK, how about using the already existing `$USER` environment variable?

PR #2361, opened 2018-03-23
Use cmake TIMESTAMP

bmwiedemann:
This greatly simlifies the code
and helps to allow reproducible builds.
See https://reproducible-builds.org/ for why this is good.


DarthGandalf:
Is combination of this and #2360 enough to make builds reproducible?

bmwiedemann:
It is for my openSUSE build tests where we have a very normalized build environment (constant user, path, umask, locale, timezone)

PR #2362, opened 2018-03-25
AppVeyor: get zlib from vcpkg

craftwar:
This uses VS2017 zlib and keeps zlib updated.
I don't know about nuget, but I think it's VS2015 lib (v140 string in path). We should use VS2017 lib if possible.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,6 +3,7 @@ image: Visual Studio 2017
 cache:
     - c:\deps -> .appveyor.yml
     - c:\spellcheck-dicts -> .appveyor.yml
+    - c:\Tools\vcpkg\installed
 environment:
     matrix:
         - cmake_build_type: Release
@@ -70,7 +71,6 @@ install:
     - c:\cygwin\bin\ls -l /cygdrive/c/Libraries
     - c:\cygwin\bin\ls -l /cygdrive/c/ProgramData
     - c:\cygwin\bin\env
-    - nuget install zlib -OutputDirectory c:\deps
     - nuget install Gettext.Tools -OutputDirectory c:\deps
     - c:\cygwin\bin\find /cygdrive/c/deps "-type" f
     - ps: |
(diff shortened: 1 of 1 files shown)


Stanzilla:
Finally someone started looking into vcpkg, nice! Nit: Can we keep lowercase folder names, so `tools` instead, please.

.appveyor.yml
```diff
@@ -3,6 +3,9 @@ image: Visual Studio 2017
 cache:
     - c:\deps -> .appveyor.yml
     - c:\spellcheck-dicts -> .appveyor.yml
+    - c:\Tools\vcpkg\installed
+init:
+  - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1')) #rdp debug
```
Stanzilla: What is this used for?

IceN9ne:
A previous commit split our x86 and x64 builds to where we're using 2015 for x86 and 2017 for x64.

Edit: Nevermind, we're using 2017 for everything, except for Qt libs on x86 which are 2015.

.appveyor.yml
craftwar: sorry, i forget to remove it, it's rdp testing. fix soon

craftwar:
We are using zlib 1.2.8 from nuget. There is newer bug fix version 1.2.11 from vcpkg. It's no problem. You can build VS2015 zlib if you are using VS2015.

craftwar:
I think keep the same as AppVeyor is better. There are other uppercase folder names elsewhere. They are all the same as AppVeyor.

.appveyor.yml
```diff
@@ -156,11 +149,11 @@ build_script:
     - c:\cygwin\bin\ls -l release/
     - ps: |
         if ($env:cmake_build_type -eq "Debug") {
-            $zlib = c:\cygwin\bin\find /cygdrive/c/deps/ -path "*v140*/$env:nuget_arch/Debug/zlibd.dll"
-            $zlibpdb = c:\cygwin\bin\find /cygdrive/c/deps/ -path "*v140*/$env:nuget_arch/Debug/zlibd.pdb"
+            $zlib = "c:\Tools\vcpkg\installed\$env:arch-windows\debug\bin\zlibd.pdb"
+            $zlibpdb = "c:\Tools\vcpkg\installed\$env:arch-windows\debug\bin\zlibd1.dll"
```
DarthGandalf: Looks reverse (pdb vs dll).
Also such variables are not necessary anymore.

.appveyor.yml
```diff
@@ -127,22 +127,15 @@ build_script:
     - cd build
     - '"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" %vc_arch%'
     - path
+    - vcpkg remove --outdated
+    - vcpkg install zlib:%arch%-windows
```
DarthGandalf, line 131: Thanks, this is much cleaner than my c:\cygwin\bin\find hack

DarthGandalf (approved):

Stanzilla (approved):

PR #2364, opened 2018-04-07
AppVeyor: grab gettext from vcpkg and clean up a bit

Stanzilla:
* Gettext from vcpkg
* No longer handle OpenSSL ourselves since it's preinstalled in AV (https://www.appveyor.com/docs/build-environment/)
* No longer handle NSIS ourselves since it's preinstalled in AV (https://www.appveyor.com/docs/build-environment/)
* Remove the strawberry caching logic since we never cache it
* Remove a few `ls` so they don't flood the log. I mainly did it while testing, happy to put them back if wanted.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@ version: 0.0.0.0.1-branch-{branch}-build-{build}
 image: Visual Studio 2017
 cache:
     - c:\spellcheck-dicts -> .appveyor.yml
-    - c:\Tools\vcpkg\installed
+    - c:\Tools\vcpkg\installed -> .appveyor.yml
 environment:
     matrix:
         - cmake_build_type: Release
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,6 @@
 version: 0.0.0.0.1-branch-{branch}-build-{build}
 image: Visual Studio 2017
 cache:
-    - c:\deps -> .appveyor.yml
     - c:\spellcheck-dicts -> .appveyor.yml
     - c:\Tools\vcpkg\installed
 environment:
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -60,45 +59,11 @@ environment:
...
-        } else {
-            choco install strawberryperl -y $env:choco_arch
-            # Strawberry perl doesn't support MSVC, but we force the support.
-            c:\cygwin\bin\bash -c "/bin/cat .appveyor.perl >> /cygdrive/c/Strawberry/perl/lib/core/config.h"
-            c:\cygwin\bin\touch /cygdrive/c/Strawberry/kvirc_workaround
-        }
+        choco install strawberryperl -y $env:choco_arch
+        # Strawberry perl doesn't support MSVC, but we force the support.
+        c:\cygwin\bin\bash -c "/bin/cat .appveyor.perl >> /cygdrive/c/Strawberry/perl/lib/core/config.h"
+        c:\cygwin\bin\touch /cygdrive/c/Strawberry/kvirc_workaround
```
DarthGandalf: This file was for the cache; if you don't want cache, this file is not needed.

DarthGandalf (commented):
Probably perl cache was removed at some point...

These `ls` are not strictly necessary, but they can help understand what went wrong when there are issues. I understand they are huge... maybe there is some way to make them downloadable as artifact instead?

.appveyor.yml
```diff
@@ -60,45 +59,11 @@ environment:
...
-    - c:\cygwin\bin\ls -l "/cygdrive/c/Program Files (x86)"
-    - c:\cygwin\bin\ls -l "/cygdrive/c/Program Files"
-    - c:\cygwin\bin\ls -l /cygdrive/c/Tools
-    - c:\cygwin\bin\ls -l /cygdrive/c/Libraries
-    - c:\cygwin\bin\ls -l /cygdrive/c/ProgramData
-    - c:\cygwin\bin\env
-    - nuget install Gettext.Tools -OutputDirectory c:\deps
-    - c:\cygwin\bin\find /cygdrive/c/deps "-type" f
     - ps: |
-        if ((Test-Path "c:\OpenSSL-$env:ssl_arch") -And ((Get-Item "c:\OpenSSL-$env:ssl_arch\bin\ssleay32.dll").VersionInfo.ProductVersion -eq '1.0.2o')) {
```
Dessa: please reinstate the openssl handling entirely. AVs pre-installed one is outdated and the whole check is there so it does nothing if at any point in time the version on AV is the same one as the installed one.

Dessa (changes requested):

Stanzilla:
@DarthGandalf done. yeah I remember we tried to make perl cache work but the free AppVeyor plan does not allow such a big cache so it was removed.

@Dessa done!

Dessa (approved):

PR #2365, opened 2018-04-07
add a changelog generating script

Stanzilla:
Usefulness of a changelog based on `git log` is open to discussion.


Stanzilla:
Alternate version without author names would look something like this https://gist.github.com/Stanzilla/83925a404e102014203e92ebc3c7f8e2

DarthGandalf:
Dunno about the format, but `ChangeLog` file is outdated right now... I'll add this to 5.0 milestone.

PR #2367, opened 2018-04-07
serverdb.kvc: Applied general cleanup and network fixes

IceN9ne:
Removed single server entries leaving only pools/round-robins (when available)
Fixed SSL entries
Added missing pools
Added missing SSL entries to increase default security options
Reorganized entries
Standardized naming scheme for all entries

Stormdancing changed to KaraokeWorld
Terra changed to IRC-Hispano

Removed dead networks:
  SyrolNet
  Telstra
  UnderZ
  Univers
  Warpednet (Merged with freenode)

Removed networks that autoban on join:
  Zurna


Dessa:
maybe put the ssl entries right below their respective "normal" entries?
or was there a reason in the way you did it that i am missing?

IceN9ne:
I did it that way because that's how it had been for other entries, so I tried to keep the style consistent.  KVIrc will sort them by name when displaying in the client. I'm not opposed to sorting them that way and had considered making the change.

wodim:
I very much agree with removing all the specific servers and only leaving the random servers there.

Stanzilla (approved):

Issue #1691, opened 2015-10-23
Windows installer options


Stanzilla:
Adding to https://github.com/kvirc/KVIrc/issues/1517 I think we should add and remove a few options in the installer:
- [ ] 1. Start menu => Keep
- [ ] 2. Desktop shortcut => Keep
- [ ] 3. Quick launch icon => Remove, not used in modern Windows
- [x] 4. Autostart => Remove
- [ ] 5. WinAmp plugin => Remove, if people want those plugins they can get it from the website, no? 
- [ ] 6. Pin to Taskbar => Add, Windows only
- [ ] 7. Selection for https://github.com/kvirc/KVIrc/issues/1517 => Add

Also @Dessa mentioned the possibility of moving to NSIS 3 (https://sourceforge.net/projects/nsis/files/NSIS%203%20Pre-release/3.0b2/)

DarthGandalf:
> does this show up on Linux?

Installer itself doesn't show up on Linux. Installation is done there differently, in the same way how other software is installed on Linux: either via package manager of distro, or via `cmake; make install`

staticfox:
Enumerating my thoughts:
1) ok
2) ok
3) no, it was a feature until 7 and there's not benefit to removing it
4) ok
5) I can see it, it's always been grayed out for me on d00ps builds anyways
6) no, see http://blogs.msdn.com/b/oldnewthing/archive/2003/09/03/54760.aspx

staticfox:
Note: highlighting https://github.com/kvirc/KVIrc/issues/1573 because it's relevent

Stanzilla:
@blindsighttf2 about 6, that blog entry is about XP and from the year 2003. I was actually talking about the Start Menu pinning that is used in Win8+ which is programmatically accessible

staticfox:
Hmm.. I'll look in to it further, but I'm not going to touch anything until at least 2 other devs give feedback.

un1versal:
I dont use windows by default so I cant really speak as a user but here are some thoughts.

:-1:  to removing 3, more Windows versions that still support that option exist than those which dont. and while KVIrc supports them this is particularly hard —unless you know for sure the amount of users on any particular Windows version are— to make a properly informed decision.

:-1: for removing winamp but :+1:  for adding support to other players, I mean the impact of this on installer size doesn't justify removing stuff like this plus only provide it buried in an outdated website which scares users away, not ideal imo at this time.

[...]

PR #2368, opened 2018-04-09
Installer fixes

Stanzilla:
* fix the finish page so it launches KVIrc as the user that ran the install before UAC privilege escalation.
* remove autostart related functions
* remove Winamp installer option

Rebase of https://github.com/kvirc/KVIrc/pull/2147 + additions from https://github.com/kvirc/KVIrc/issues/1691


Dessa:
here is hoping that the plugin actually works with nsis3. if not there is also https://github.com/lordmulder/stdutils but that does only work with 32 bit nsis (and if it works in nsis3 then it is "by luck")

Stanzilla:
stdutils would also help the new items in https://github.com/kvirc/KVIrc/issues/1691 taskbar pinning etc https://stackoverflow.com/questions/24326685/pin-icons-to-taskbar?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

Stanzilla:
bug: shows the "kvirc is installed, want to start it?" page before actually installing it.

Dessa (approved):

PR #2369, opened 2018-04-10
use the StdUtils NSIS plugin

Stanzilla:
The ExecShellAsUser function it provides lets us start KVIrc as a user instead of admin from the last page of setup.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -59,6 +59,10 @@ environment:
 matrix:
     fast_finish: true
 install:
+    # - systeminfo
+    # - c:\cygwin\bin\uname -a
+    # - c:\cygwin\bin\cat /proc/cpuinfo
+    # - c:\cygwin\bin\cat /proc/meminfo
     - ps: |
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -96,6 +100,15 @@ install:
             (New-Object Net.WebClient).DownloadFile('https://sourceforge.net/projects/wordlist/files/speller/2017.08.24/hunspell-en_GB-ise-2017.08.24.zip', 'c:\spellcheck-dicts\zip\en_GB-ise.zip')
             c:\cygwin\bin\bash -lc "cd /cygdrive/c/spellcheck-dicts/unzip; 7z x -y ../zip/en_US.zip; 7z x -y ../zip/en_GB-ize.zip; 7z x -y ../zip/en_GB-ise.zip"
         }
+    - ps: |
+        if (Test-Path "C:\Program Files (x86)\NSIS\Plugins\StdUtils.dll") {
```
Dessa: its not actually cached anyway is it?

Stanzilla: it isn't but I figured why not, maybe it will be in the future. I realize that I am not very consistent :)

Dessa: well then use the right path at least :P

PR #2371, opened 2018-04-11
change dev build installer name

Stanzilla:
it is now based on the last git tag only and more accurately represents the real state.


DarthGandalf (approved):
Probably name of exe should also mention release vs debug... But I like this change

Stanzilla:
@DarthGandalf good idea, `KVIrc-5.0.0-beta1-59-g8c49a773d-x64.Release.exe` or different? `KVIrc-5.0.0-beta1-59-g8c49a773d-x64-Release.exe`?

Issue #1848, opened 2016-01-17
Bug: Theme installing/loading/unistalling of lower or greater engine version themes is utterly broken.


un1versal:
### Bug 1 no consistent and effective way of messaging user on installing / loading themes

for e.g. https://github.com/kvirc/KVIrc/blob/master/src/kvirc/kernel/KviTheme.cpp#L127-L132 **IS BROKEN** you may get 

thisngs like https://github.com/kvirc/KVIrc/blob/master/src/modules/theme/ThemeFunctions.cpp#L213-L218 and https://github.com/kvirc/KVIrc/blob/master/src/modules/theme/ThemeFunctions.cpp#L229-L234 they dont account for both greater/lower 
- A warning pop up dialog message be printed that reflects the fact a user is **installing a lower** engine theme version
- A warning pop up dialog message be printed that reflects the fact a user is **installing a greater** engine theme version
- A warning pop up dialog message be printed that reflects the fact a user is **loading a lower** engine theme version
- A warning pop up dialog message be printed that reflects the fact a user is **loading a greater** engine theme version
#### Bug 2 uninstalling a theme via theme web browser

![capture](https://cloud.githubusercontent.com/assets/3521959/12222788/5547babe-b7be-11e5-81a0-a5d675f2afc5.PNG)

Clicking that uninstall button does nothing here...
### Bug 3

when `KVI_CURRENT_THEME_ENGINE_VERSION "1.0.0"` installing a higher theme version you get no warning and theme is invisible on theme manager

> reference
> https://github.com/kvirc/KVIrc/pull/1832

as far as Im concerned this is a release blocker on its own.

PR #2373, opened 2018-04-19
update version to 5.0.0

Stanzilla:
About time? did I forget something?


DarthGandalf (changes requested):
This is not "Personnel usage"

Stanzilla:
I am pretty sure it is, kvirc is not a business and there is no money made with it.

Dessa:
that doesn't really matter, it is still subject to a license and as such (IANAL) not really personal

Stanzilla:
The license is CC0 Creative Commons which should be fine?

wodim:
The licence is conflicting. CC0 is public domain which should be OK but the text below says it's only for personal use.

Also, being honest it's not clear to me that the licence is ok: all posts in that website look like they are automatically generated, maybe scrapped from other websites online.

We should use a picture with a clear licence and where the author is clearly identified.

Stanzilla:
I mean... https://www.google.com/search?tbs=sbi:AMhZZisR2YE5YOioB1Ld8KdR6Yqx7LpD2PrynnSogRdsiGOLWL3WMJbcsfHqU8N0dSJFaYxjNyoAg3rxr8YoktXwUlKg0iBMSoc-5XmtbTB_1pHcWfeSiVhiy_1cjWqV92V1rMSNRj84ubVxuUaWs4LETT638uxuYur66MkeXk_1bgHP1e6usGUz7Yu7190uc7BSx-wKTpvkfJvv1ZbQB9DAgZd8EU4lmE8hqyvTmGX-OGE_1kMnJNNIsl5IGIHh5_1Y6j204-E4LOIx0doh9teEi4EzKcJ_11-O1WWcEzUTsx4o-bATnDY5X2YHqBQEAaS3ACwFhz5vSAn1kt&width=2560&height=1600

IceN9ne:
I agree that we need to version bump to 5.0.0 and I'm all for it.  I just want these concerns addressed first:

- The rc file is still named `.kvirc4.rc` and `kvirc4.ini`. Should we change it? Should we leave it? (`KVI_HOME_CONFIG_FILE_NAME`)
- KVIrc modules have "4.x.x" in their KVIRC_MODULE register macro, will this cause issues?
- KviIpcSentinel has object name `kvirc4_ipc_sentinel`
- Will scripts/themes have issues loading after the version bump?

I suppose that as long as modules load fine, then everything can be addressed in future commits

wodim:
I say leave the config files and the sentinel with the same name unless you want to write the code to rename the old config files

Dessa:
- that would need a migration but probably. its also not-so-great that it is in C:\Users\Username directly on windows, but that probably(?) needs to stay.
- no idea, never tried, but i *think* there was a version check at some point?
- can't really comment on that
- maybe. there is also still https://github.com/kvirc/KVIrc/issues/1848 and this version introduces a greater theme version. but i don't think it matters much, probably.

Stanzilla:
I'd say we leave everything as is in this PR and deal with potential migrations later. One thing that should be made sure of, is that modules work and so far everything was working fine for me but please install the appveyor builds and confirm.

PR #2374, opened 2018-04-20
remove russian website and mailing list links from help menu

Stanzilla:

### DIFF
--- a/src/kvirc/kernel/KviCoreActions.cpp
+++ b/src/kvirc/kernel/KviCoreActions.cpp
@@ -319,32 +319,10 @@ void register_core_actions(KviActionManager * m)
 	    KVI_SHORTCUTS_HELP);
 
 	SCRIPT_ACTION(
-	    KVI_COREACTION_KVIRCMAILINGLIST,
-	    "openurl http://www.kvirc.net/?id=mailinglist",
-	    __tr2qs("Subscribe to the Mailing List"),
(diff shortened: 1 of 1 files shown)


DarthGandalf:
kvirc.ru is still up...

swarfega:
Could we use the template from .ru on .net?

DarthGandalf:
It's MediaWiki

swarfega:
It looks nicer than the .net website imo.

Stanzilla:
.ru is still up but not updated and "under our control" or is it your site, @DarthGandalf I don't even know

DarthGandalf:
No, it's by another Alexey... I'll contact him.

Stanzilla:
Thanks, I think it's risky to point people to a website we don't have control over. Could change any day and then scam people

auzhva:
Hi everyone:)

I’m that Alexey, that you are looking for:) That created kvirc.ru site and actually Russian community community quite a while ago.

Despite not being updated too often, it still has some traffic. And downloads from there.

So I don’t think shutdowns is a great way to go. 

It would be absolutely great if someone from here wants to contribute. That person doesn’t need permission. It’s wiki. Just do!

If someone trusted wants to have admin permission to make it better - let’s do so as well

auzhva:
Btw, this site might be not updated, but it will never ever scam people;)

PR #2378, opened 2018-04-22
LF all the things

Stanzilla:
the problem is since there never was a gitattributes file in the repo and no editorconfig enforced, every windows commiter potenially changed some files to CRLF, so this changes all crlf files back to lf.

not sure if best practice or not


staticfox:
Is this something that clang format wasn’t set up to do?

Stanzilla:
that is actually something you can't even tell clang-format to do. At least I haven't been able to find out how.

staticfox:
Well, if this means that the repo is consistently LF then +1 here

Stanzilla:
I am no expert on .gitgnore but it is currently set up like this:

```
* text=auto
*.cpp text eol=lf
*.h text eol=lf
```
is this enough?

IceN9ne:
What about utf-8 all the things too? It might be a good opportunity to do that as well if we want that.  We have an assortment of binary, us-ascii, and utf-8 charsets for our .cpp and .h files.

wodim:
>We have an assortment of binary, us-ascii, and utf-8 charsets for our .cpp and .h files.

How is that possible? Remember that `file` is not a good assessment tool for this

DarthGandalf:
Note that ASCII is a subset of UTF-8

wodim:
^ A file that doesn't contain accents or weird stuff could pass as a us-ascii file

Stanzilla:
Can we settle the LF thing first, please? :D My .gitattributes does not cover all files we have in the repo, do we need more? Something like https://github.com/alexkaratarakis/gitattributes

craftwar:
`* text=auto`
You need only this line. Git will push all text files as LF to repo and checkout according to OS.

reference cross-platform project
https://github.com/obsproject/obs-studio

Stanzilla:
@craftwar that is what I thought before, then someone told me otherwise: https://github.com/zeit/hyper/pull/2795 not sure if true or not. What yours will do is let your local Git installation decide on what to do, based on the core.autoCRLF setting, afaik.

craftwar:
@Stanzilla  according to git document https://git-scm.com/docs/gitattributes
I think git will detect if it's a text or binary automatically.

add
`*.nsi		text`
git or github think it's binary file

wodim:
nsi are binary files because they are not written in utf-8. **Don't convert them to utf-8!!!**

Stanzilla:
So what we should do is tell Git which files are binary so it ignores it, like that https://github.com/alexkaratarakis/gitattributes/blob/master/C%2B%2B.gitattributes

wodim:
These should be tested. I am not sure if nsis works with utf-8 now, that's why I insisted on not converting them

Stanzilla:
sure, just waiting for appveyor

staticfox:
This UTF-8 stuff probably should have been left to another PR since this PR was intended to be specific towards line endings. Now it seems like we're dealing with 2 unrelated issues.

Stanzilla:
Yeah agree, but it's done now, I could still split it off if wanted

staticfox:
Eh, it's fine. Moving forward, let's just keep PRs specific towards the topic. Otherwise things will get backlogged due to unrelated debates.

Stanzilla:
Build is green, any last words?

Issue #2376, opened 2018-04-20
adding openssl version to about


NicklaTome:
Hi.
Would KVIrc be adding openssl version informations into about programs into executables informations tabs?
Is very burried and manuals work to find version now. Please.
Thank you very much.

PR #2380, opened 2018-04-28
AboutDialog: OpenSSL infos

Pan7:
Show the OpenSSL informations version, compiler flags and buildon in the about dialog.
See https://github.com/kvirc/KVIrc/issues/2376

### DIFF
--- a/src/modules/about/AboutDialog.cpp
+++ b/src/modules/about/AboutDialog.cpp
@@ -42,6 +42,10 @@
 #include <QEvent>
 #include <QCloseEvent>
 
+#ifdef COMPILE_SSL_SUPPORT
+#include <openssl/crypto.h>
+#endif //COMPILE_SSL_SUPPORT
+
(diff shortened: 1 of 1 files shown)


src/modules/about/AboutDialog.cpp
```diff
@@ -191,6 +195,33 @@ AboutDialog::AboutDialog()
...
+	infoString += "<br>";
+	infoString += __tr2qs_ctx("OpenSSL compiler flags", "about");
+	infoString += ": ";
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L)
+	infoString += SSLeay_version(OPENSSL_CFLAGS);
+#else
+	infoString += OpenSSL_version(OPENSSL_CFLAGS);
+#endif
+	infoString += "<br>";
+	infoString += __tr2qs_ctx("OpenSSL build on", "about");
```
wodim, line 216: built*

PR #2381, opened 2018-04-28
AboutDialog: OpenSSL infos

Pan7:
Show the OpenSSL informations version, compiler flags and builton in the about dialog.


wodim:
You don't need to open another PR to make a change. Just push more commits to your branch and they'll show up here.

PR #2390, opened 2018-06-04
AppVeyor: Use latest Qt

craftwar:
We don't need change Qt version any more. It will use latest release on AppVeyor.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,7 +12,7 @@ environment:
           nuget_arch: x64
           vc_arch: amd64
           choco_arch:
-          qt_ver: 5.9\msvc2017_64
+          qt_ver: latest\msvc2017_64
           MSYSTEM: MINGW64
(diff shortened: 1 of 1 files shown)


DarthGandalf:
I have mixed feelings about this.
One one hand, it requires less changes from us, and keeps the version up to date.
On the other hand, the version changes randomly underneath, so e.g. when KVIrc 5 is going to be released, Appveyor changes the latest Qt version, and KVIrc is released with unknown bugs, because it was never tested on that version of Qt (on windows).

craftwar:
For stable, keep it on 5.9 TLS is good choice.
How about adding another matrix? Use latest for building test only, fix any issue introduced by newer Qt.
con: make build time longer

Stanzilla:
I vote nay because of exactly what DarthGandalf said. Another matrix would be fine but as you already said, longer build times and we are already super slow there.

craftwar:
Close, don't change Qt til next TLS 5.12.

PR #2391, opened 2018-06-04
AppVeyor: Grab openssl from vcpkg and fix vcpkg remove option

craftwar:
- Old openssl is built in VS2013, grab from vcpkg to make more consistent (VS2017).
  Remove dependency on msvcr120.dll
- Fix vcpkg remove option

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,6 @@ environment:
         - cmake_build_type: Release
           arch: x64
           target_arch: x86_64
-          ssl_arch: Win64
           nuget_arch: x64
           vc_arch: amd64
           choco_arch:
@@ -21,7 +20,6 @@ environment:
         - cmake_build_type: Release
           arch: x86
           target_arch: x86
-          ssl_arch: Win32
           nuget_arch: Win32
           vc_arch: amd64_x86  # cross-compile from amd64 to x86
           choco_arch: --x86
@@ -34,7 +32,6 @@ environment:
         - cmake_build_type: Debug
           arch: x64
           target_arch: x86_64
-          ssl_arch: Win64
           nuget_arch: x64
           vc_arch: amd64
           choco_arch:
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -127,9 +113,9 @@ build_script:
         }
     - ps: |
         if ($env:cmake_build_type -eq "Debug") {
-            c:\cygwin\bin\cp -pv "c:/Tools/vcpkg/installed/$env:arch-windows/debug/bin/zlibd.pdb" "c:/Tools/vcpkg/installed/$env:arch-windows/debug/bin/zlibd1.dll" "/cygdrive/c/OpenSSL-$env:ssl_arch/libeay32.dll" "/cygdrive/c/OpenSSL-$env:ssl_arch/ssleay32.dll" "/cygdrive/c/OpenSSL-$env:ssl_arch/bin/msvcr120.dll" "c:/Program Files (x86)/Windows Kits/10/bin/$env:arch/ucrt/ucrtbased.dll" release/
+            c:\cygwin\bin\cp -pv "c:/Tools/vcpkg/installed/$env:arch-windows/debug/bin/zlibd.pdb" "c:/Tools/vcpkg/installed/$env:arch-windows/debug/bin/zlibd1.dll" "c:/Tools/vcpkg/installed/$env:arch-windows/debug/bin/libeay32.*" "c:/Tools/vcpkg/installed/$env:arch-windows/debug/bin/ssleay32.*" "c:/Program Files (x86)/Windows Kits/10/bin/$env:arch/ucrt/ucrtbased.dll" release/
```
DarthGandalf, line 116: What does * mean here? Is it more than .dll?

DarthGandalf (approved):

.appveyor.yml
craftwar: dll and pdb

ras0219-msft: Just as a heads-up; we're looking at adding a `d` suffix to the debug OpenSSL dlls, so you may need to change this in the future to include that.

PR #2394, opened 2018-06-19
Option for lookup local hostname as seen by the IRC server

Pan7:
Option to disable hostname lookup if DCC or any other p2p service isn't required.
No more "Looking up the local hostname as seen by the IRC server " and "Unable to resolve the local hostname as seen by the IRC server: DNS query failed, using previously resolved" ... messages.


wodim:
Need to bump `KVI_NUM_BOOL_OPTIONS`

Pan7:
Maybe it would be better to remove the lookup because IRC servers often change it to protect users.

kadirselcuk (approved):

ctrlaltca:
The lookup can still be useful to others, so removing it is not convenient.
A option already exists to avoid resolving the hostname returned by the server.
Look at the "broken bouncer hack" in DCC -> Advanced

PR #2395, opened 2018-07-18
Cmake: Change MSVC compiler flags

craftwar:
Move common MSVC flags to one place.

use /O2 is better than /Ox (full optimization is misguiding)
https://docs.microsoft.com/en-us/cpp/build/reference/ox-full-optimization
> but the /Ox compiler option enables only a subset of the speed optimization options enabled by /O2.

compile error reason
/GL is causing error (may be compiler bug, I have reported ffmpeg /GL bug to VC team)
remove /GL and leave as comment


IceN9ne:
`/Od` is set in the `WANT_DEBUG` section, which is what you want if you want debugging. If you look in the `else()`, it sets `/Ox` for non-debug builds. I fail to see a needed change.

Edit: This comment applies to the original post prior to being edited.

craftwar:
@IceN9ne /O2 > /Ox as I mention in my comment.
Thanks, I fix errors in config and comments.

From my experience in OBS-Studio, cmake supports different build types. No need to setup most options, I guess it may be backward compatibility. RelWithDebInfo should be release build with debug info. I try to keep the same behavior anyway.

PKEuS:
If you are already changing compiler flags, what about adding /Zc:throwingNew? Or does KVIrc use code that does not have a conforming operator new? In particular, if /GL is removed (which should be necessary for LTCG), this might have an effect.

craftwar:
@PKEuS /GL implies /LTCG (always add /LTCG to linker option or linker needs restart), but /LTCG can be used without /GL. I don't know how to add it in cmake.

Nmake is old, it doesn't use full processor cores to build. It uses one cl process compile one file every invoke. (adding /mp doesn't help in this case, keep it won't hurt any way.)

Stanzilla:
@ctrlaltca could you restart appveyor please? :)

Stanzilla (approved):

CMakeLists.txt
```diff
@@ -228,8 +233,8 @@ else()
 			if(MSVC)
 				# this is the equivalent of -O0 -g according to MSDN
 				# QT4 has /Zc:wchar_t-, QT5 has /Zc:wchar_t
```
DarthGandalf, line 235: This comment is outdated

CMakeLists.txt
```diff
@@ -198,6 +198,11 @@ if(WANT_ENV_FLAGS)
 	set(CMAKE_BUILD_TYPE Undefined)
 	set(CMAKE_STATUS_DEBUG_SUPPORT "Environment-driven")
 else()
+	if(MSVC)
+		# QT4 has /Zc:wchar_t-, QT5 has /Zc:wchar_t
+		set(CMAKE_CXX_FLAGS "/MP /Zc:wchar_t /EHsc /source-charset:utf-8")
+		set(CMAKE_C_FLAGS "/MP /Zc:wchar_t /EHsc /source-charset:utf-8")
+	endif()
 	if(WANT_DEBUG)
```
DarthGandalf, line 206: I don't agree with concept of WANT_DEBUG here at all, because in MSVC Release vs Debug is selected after CMake is done. But that's irrelevant for this PR...

DarthGandalf (approved):

Issue #1274, opened 2015-07-26
Improve SSL Security from Man-In-The-Middle-Attacks


DarthGandalf:
**Reported by da2ce7 on 17 Mar 2012 02:45:10 UTC**
In the current version of KVIrc, the program will blindly connect to the same server with a different SSL certificate.  (Identified by its Fingerprint).

This is a major security issues, as people connecting to their bounces may get their password stolen, or connecting to IRC servers, have their nick-server passwords/identities/channels taken from them.

KVIrc should follow the standard practice used by SSH:
- [ ]   Upon first connection to a server, a warning should be displayed for the user to check the servers SSL fingerprint against a known good source.

(Maybe KVIRC should ship with a list of known good SSL certificate fingerprints, for the popular servers, such as Freenode.org)
- [ ]  When connecting to a server and different (non-whitelisted) SSL certificate is provided, KVIrc MUST refuse to connect until the user has whitelisted the SSL certificate.

If KVIrc just displays a warning however connects anyway, the damage is already done, as the malicious attacker already has the passwords, and identification information.

This is a security related bug for KVIrc, and should be fixed in the next public release.

---

Migrated from: https://svn.kvirc.de/kvirc/ticket/1274

---

Edited **Thu  4 Aug 15:08:40 UTC 2016**

A shortlist of ESSENTIAL SSL support bits.
- [ ]  way to verify the current certificates/block or whitelist  
- [ ]  way to deny self signed certificates optionally  
- [ ]  way to monitor which SSL methods / protocols are when using SDCC methods like SDCC file transfer SDCC voice etc.
- [ ]  warnings and ssl certificates popups a la mIRC
- [ ]  warnings when using low security methods for SDCC
- [ ]  more verbose output related to SDCC in transfers window

in short a better way to manage SSL connections and all related tasks

DarthGandalf:
**Comment by ctrlaltca on 18 Mar 2012 19:53:03**
The basic kvirc support for ssl is meant to provide communication security, not authentication.
If you want to implement authentication, it can be done using scripting: please have a look at [http://www.kvirc.ru/forum/topic,1095.0], ticket #1170, revision r5889
Hardcoding certificate fingerprints inside kvirc is probably a bad idea, due to out slow release cycle.
Adding a dynamic certificate datastore to kvirc, some certificate checking facilities and maybe providing some base certificates is probably a nice idea. Or could provide a way to interface with the system certificate datastore.

DarthGandalf:
[...]

PR #2403, opened 2018-10-13
Add EsperNet

lol768:
* Adds [EsperNet](https://esper.net) to the server list.

### DIFF
--- a/data/config/serverdb.kvc
+++ b/data/config/serverdb.kvc
@@ -203,6 +203,12 @@ NServers=2
 0_Hostname=irc.enterthegame.com
 0_Description=EnterTheGame:%20Random%20server
 NServers=1
+[EsperNet]
+0_Hostname=irc.esper.net
+0_Description=EsperNet:%20Random%20server%20SSL
+0_Port=6697
+0_SSL=true
+NServers=1
 [euIRC]
 0_Hostname=irc.euirc.net
 0_Description=euIRC:%20Random%20server


IceN9ne:
I'll merge it if you add a non-SSL connection as `0` and move the SSL version to `1`.

lol768:
I'm not encouraging users to use insecure transport, and you shouldn't do the same :/

It's no different to moving towards deprecating HTTP and pushing HTTPS as the preferred way to access web applications.  There are [mature proposals](https://ircv3.net/specs/extensions/sts.html) that implement a HSTS parallel on top of IRC for this very reason.

if you're not willing to budge on this, please do close the PR.

IceN9ne:
The use of SSL when there are no verification mechanisms in place is only a false sense of security.  This false sense of security is more detrimental than knowingly using non-SSL as at least then you'll consider your usage as opposed to potentially leaking sensitive information.

lol768:
>The use of SSL when there are no verification mechanisms in place is only a false sense of security

We don't use self-signed certificates, they are issued by an actual CA after verification of domain ownership. Users being dumb and disabling verification mechanisms doesn't sound like an issue with the underlying transport to me.

KVIrc does support verifying the chain of trust of the presented server certificates, right?

IceN9ne:
I never said it was an issue with the underlying transport. I was eluding to an issue with the current implementation. Also, not everything requires SSL. The world is about choices, just like the choice to use a vhost, +x, or nothing.

IceN9ne:
This philosophical debate will take us nowhere.

lol768:
>I was eluding to an issue with the current implementation

I wouldn't like to speculate without being familiar with the codebase, but KVIrc 4.9.3 was quite happy to connect to a server with an expired cert as well as a self-signed generated cert with the completely wrong commonName when I tested. I guess now I see what you meant.
<details>






</details>

If this is indeed the case, sounds like a pretty serious issue unrelated to this PR that should be fixed as a priority.

Edit: I guess https://github.com/kvirc/KVIrc/issues/1274 is related.

>Also, not everything requires SSL. The world is about choices, just like the choice to use a vhost, +x, or nothing.

Users can still specify their own host/port/SSL combination to manually connect to, no? There is choice there for users, who understand the risk, to make the decision to connect in an insecure manner if they wish.

But when we have services that receive account credentials (which are probably reused by most users) and potentially sensitive conversations in private channels, do you not think it's responsible to do things securely by default?

-----

In any case, thank you for merging the request.

PR #2408, opened 2018-10-29
AppVeyor: Workaround for Qt5.9 bug

IceN9ne:


DarthGandalf:
Do you have any details what bug is that? Any reference to upstream bugtracker?

IceN9ne:
@ctrlaltca summed it up enough in his other project: https://github.com/Cockatrice/Cockatrice/issues/3378

This is just bypassing the error for the x86 builds so that we can get x64 debug builds until AppVeyor deploys a fixed Qt5.9.

PR #2410, opened 2018-11-17
SASL: Add support for EXTERNAL method.

IceN9ne:
- Adds dynamic fallback from EXTERNAL to PLAIN if auth fails on the network.

### DIFF
--- a/src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
+++ b/src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
@@ -3039,6 +3039,23 @@ void KviIrcServerParser::parseNumericSaslFail(KviIrcMessage * msg)
 		pOut->output(KVI_OUT_SERVERINFO, __tr2qs("SASL authentication error: %Q"), &szParam);
 	}
 
+	// Handle fallback if possible for SASL auth failure or dump if user mandates SASL
+	// and no fallback is available
+	if(msg->numeric() == 904)
+	{
+		if(msg->connection()->stateData()->sentSaslMethod() == QStringLiteral("EXTERNAL"))
+		{
+			if(!msg->connection()->target()->server()->saslNick().isEmpty() && !msg->connection()->target()->server()->saslPass().isEmpty())
+			{
+				KviWindow * pOut = static_cast<KviWindow *>(msg->console());
+				pOut->output(KVI_OUT_SERVERINFO, __tr2qs("Attempting fallback due to SASL failure."));
+				msg->connection()->sendFmtData("AUTHENTICATE PLAIN");
+				msg->connection()->stateData()->setSentSaslMethod(QStringLiteral("PLAIN"));
+				return;
+			}
+		}
+	}
+
 	if(msg->connection()->stateData()->isInsideAuthenticate())
 		msg->connection()->endInitialCapNegotiation();
 }
--- a/src/kvilib/irc/KviIrcServer.cpp
+++ b/src/kvilib/irc/KviIrcServer.cpp
@@ -94,6 +94,7 @@ KviIrcServer::KviIrcServer(const KviIrcServer & serv)
 	m_bAutoConnect = serv.m_bAutoConnect;
 	m_szSaslNick = serv.m_szSaslNick;
 	m_szSaslPass = serv.m_szSaslPass;
+	m_szSaslMethod = serv.m_szSaslMethod;
 
 	if(serv.m_pAutoJoinChannelList)
 		m_pAutoJoinChannelList = new QStringList(*(serv.m_pAutoJoinChannelList));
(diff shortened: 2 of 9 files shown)


AndrioCelos:
Seems to be working well. It may be worth adding a GUI warning or error message if you select SASL EXTERNAL with no SSL certificate or without enabling SSL.

DarthGandalf:
The following is the kind of fallback I had in mind...

C: AUTHENTICATE EXTERNAL
S: AUTHENTICATE +
C: AUTHENTICATE +
S: :server 904 nick :SASL authentication failed
~~C: aaaaaa, I failed! what to do? disconnect!~~
C: AUTHENTICATE PLAIN
S: AUTHENTCIATE +
...

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
```diff
@@ -3039,6 +3039,30 @@ void KviIrcServerParser::parseNumericSaslFail(KviIrcMessage * msg)
...
 
+	// Handle fallback if possible for SASL auth failure or dump if user mandates SASL
+	// and no fallback is available
+	if(msg->numeric() == 904)
+	{
+		if(msg->connection()->stateData()->sentSaslMethod() == QStringLiteral("EXTERNAL"))
+		{
+			if(!msg->connection()->target()->server()->saslNick().isEmpty() && !msg->connection()->target()->server()->saslPass().isEmpty())
+			{
+				KviWindow * pOut = static_cast<KviWindow *>(msg->console());
```
DarthGandalf, line 3050: It doesn't work without static_cast?

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
```diff
@@ -3039,6 +3039,30 @@ void KviIrcServerParser::parseNumericSaslFail(KviIrcMessage * msg)
 		pOut->output(KVI_OUT_SERVERINFO, __tr2qs("SASL authentication error: %Q"), &szParam);
 	}
 
+	// Handle fallback if possible for SASL auth failure or dump if user mandates SASL
+	// and no fallback is available
+	if(msg->numeric() == 904)
+	{
+		if(msg->connection()->stateData()->sentSaslMethod() == QStringLiteral("EXTERNAL"))
+		{
+			if(!msg->connection()->target()->server()->saslNick().isEmpty() && !msg->connection()->target()->server()->saslPass().isEmpty())
```
DarthGandalf, line 3048: I'd probably refactor this to not do the same several times...

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
```diff
@@ -3039,6 +3039,30 @@ void KviIrcServerParser::parseNumericSaslFail(KviIrcMessage * msg)
...
+				msg->connection()->stateData()->setSentSaslMethod(QStringLiteral("PLAIN"));
+				return;
+			}
+		}
+
+		if(KVI_OPTION_BOOL(KviOption_boolDropConnectionOnSaslFailure))
+		{
+			KviWindow * pOut = static_cast<KviWindow *>(msg->console());
+			pOut->output(KVI_OUT_SERVERINFO, __tr2qs("SASL auth failed. Dropping the connection."));
+			msg->connection()->sendFmtData("QUIT");
```
DarthGandalf: Server isn't guaranteed to disconect on QUIT though, even if it probably will... Can we send QUIT and then disconnect?

src/modules/options/OptionsWidget_connection.cpp
```diff
@@ -182,7 +182,10 @@ OptionsWidget_connectionSocket::OptionsWidget_connectionSocket(QWidget * parent)
 	                        "you want to rely on the DNS server to provide the best choice.",
 	                "options"));
 
-	addRowSpacer(0, 5, 0, 5);
+	b = addBoolSelector(0, 5, 0, 5, __tr2qs_ctx("Drop connection on SASL authentication failure", "options"), KviOption_boolDropConnectionOnSaslFailure);
```
DarthGandalf: Should be in a separate commit

DarthGandalf (approved):
If you tested the fallback, LGTM.

Also it would be good to parse 908 before attempting the fallback, but not necessary.

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
IceN9ne: I literally just copied that from the code that was already there to keep it consistent.

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
IceN9ne: Do what multiple times?

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
IceN9ne: The idea is to not force a disconnect so that it will automatically attempt to connect again, in case services are having issues, but I suppose in most situations where services would be having issues, you wouldn't get a 904 response.

DarthGandalf: I think it depends on server implementation. If server is evil (it's remote untrusted input after all), we don't want it to trick us into continuing connection even if user told us to disconnect in this case.

src/kvirc/sparser/KviIrcServerParser_numericHandlers.cpp
DarthGandalf: Have several pieces of code which check whether PLAIN is valid.
If one will need to be changed later, the other one will be forgotten.

PR #2411, opened 2018-11-17
Do not track Linux kernel version

bmwiedemann:
Do not track Linux kernel version
of build system to make reproducible builds easier.

See https://reproducible-builds.org/ for why this is good.

Needed because otherwise there were variations from
```diff
-set(CMAKE_SYSTEM "Linux-4.19.1-1-default")             
+set(CMAKE_SYSTEM "Linux-4.12.14-lp150.12.25-default")
-set(CMAKE_SYSTEM_VERSION "4.19.1-1-default")           
+set(CMAKE_SYSTEM_VERSION "4.12.14-lp150.12.25-default")
```

Code was simplified by also dropping it on MSWindows

This PR was done while working on reproducible builds for openSUSE.


Stanzilla (approved):

src/kvilib/config/KviBuildInfo.cpp
```diff
@@ -134,11 +138,7 @@ namespace KviBuildInfo
 
 	QString buildSystemVersion()
 	{
-#ifdef COMPILE_ON_WINDOWS
```
DarthGandalf, line 137: While I'm fine with the idea in general... What's the point of a function which returns an empty string? Please remove the function

src/kvilib/config/KviBuildInfo.cpp
```diff
@@ -120,7 +120,11 @@ namespace KviBuildInfo
 
 	QString buildSystem()
 	{
+#ifdef COMPILE_ON_WINDOWS
```
DarthGandalf: Why does this depend on windows?

bmwiedemann: I wasnt sure if it was wanted there, but dropped it now with the rest.

src/kvilib/config/KviBuildInfo.cpp
bmwiedemann: dropped it.

PR #2413, opened 2018-11-28
Attempt to update to C++17

IceN9ne:
Checking to see if Travis and AppVeyor are capable of building modern code.

Remove Q_UNUSED:
The Q_UNUSED macro is a hack to get compilers to not warn about unused variables. With C++17, the attribute `[[maybe_unused]]` was added to allow for indication that there is a possibility that the variable in question will not be used and to not warn about it. As can be seen by 081bcfd, it removes quite a bit of noise and makes it clear where the variable is declared of the intention.

Use override and remove virtual:
The override keyword that has been added to modern C++ allows for the indication that a function should override a virtual function from a base class. An error is generated when signatures do not match, preventing headaches in tracking down bugs. There were three instances that I found so far that were pointed out by the override keyword.

Virtualness of a function is inherited from the base class and is not necessary in the signature of the function, thus changing nothing whether you include it or not. With the usage of the keyword override, it becomes even more unnecessary as indication that it is a virtual override is more obvious now and allows for what should be better documentation where virtual is only used when creating a new virtual function that will be overridden by derived classes and override is only used by derived classes. This removes any confusion or need to search for who is the originator of the virtual function in the hierarchy.

Remove inline where possible:
The keyword inline has had many changes over the years in C++ and compilers no longer care about your opinion on whether you think the function should be inlined or not. This will remove all of the extraneous usages.

Remove C-style typedef structs:
The old C-style typedef structs aren't necessary in modern C++ code. Also, identifiers that begin with an _ followed by a capital letter are reserved and should be avoided to not potentially incur undefined behavior (noting that it's not likely that any of KVIrc's names would ever be chosen by a compiler or standard library).

Use inline static in classes:
Take advantage of the ability to initialize static variables in classes directly in the header instead of separate in a .cpp file.

Move away from C-style casts:
C-style classes are a shotgun approach to casting and can do unintended things when used improperly. C++ and Qt have better and more explicit casts that should be used.

New algorithms:
There are new algorithms in C++17 like std::clamp that simplify redundant code and make it more understandable, faster.

Qt5 signal/slot syntax:
The new Qt5 signal/slot syntax removes the use of SIGNAL and SLOT macros and make it a bit more meaningful when connecting and disconnecting slots.

Remove pre-processor defines in favor of constexpr:
In general, pre-processor anything should be kept to a minimum to prevent unwanted side-effects. Also, by changing to constexpr, we get C++ type safety.


Stanzilla:
Got it almost working, https://github.com/Stanzilla/KVIrc/commit/52e26f40bd13dd57d644de36c113f8574d79a88d just need @AlexanderP to look at the Debian stuff

DarthGandalf:
> (noting that it's not likely that any of KVIrc's names would ever be chosen by a compiler or standard library).

Compiler could generate names based on the names used in the program, e.g. during name mangling. So the set of the used names is not limited to STL.

src/modules/dcc/DccCanvasWindow.h
```diff
@@ -53,15 +53,15 @@ class DccCanvasWindow : public DccWindow
...
-	virtual void ownMessage(const char * text, bool bUserFeedback = true);
-	virtual void ownAction(const char * text);
+	const QString & target() override;
+	void fillCaptionBuffers() override;
+	void getBaseLogFileName(KviCString & buffer) override;
+	QPixmap * myIconPtr() override;
+	void resizeEvent(QResizeEvent * e) override;
+	QSize sizeHint() const override;
+	bool event(QEvent * e) override;
+	void ownMessage(const QString & text, bool bUserFeedback = true) override;
```
DarthGandalf, line 63: Default parameters don't work as expected in virtual functions. I'd also remove =true from here

IceN9ne: It's not wrong to use default arguments in virtual functions. While they may not be inherited, standard convention of using the same default argument in every iteration of the virtual function signature makes it very regular and expected behavior. If you want to get rid of default parameters in virtual functions completely, you'd need to change Qt5 library code as well. Take a look at QTranslator::translate or similar.

IceN9ne:
> Compiler could generate names based on the names used in the program, e.g. during name mangling. So the set of the used names is not limited to STL.

That may be a slim possibility, but I did note that the compiler _might_ use it as well. Which is why I'm in favor of removing reliance on code that could generate such undefined behavior.

Stanzilla:
New version https://github.com/Stanzilla/KVIrc/commit/52e26f40bd13dd57d644de36c113f8574d79a88d

Stanzilla:
@AlexanderP we need your help once again :)

Here is the problem: https://travis-ci.org/kvirc/KVIrc/jobs/473059746

.travis.yml
```diff
@@ -59,10 +59,9 @@ install:
     - CMAKE_ARGS=""
     - |
         if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-            sudo add-apt-repository -y ppa:george-edison55/cmake-3.x # want cmake newer than in ubuntu 14.04
             sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test # for gcc-6
             sudo apt-get update
-            sudo apt-get install -y cmake libperl-dev libenchant-dev qtmultimedia5-dev libqt5webkit5-dev libqt5svg5-dev libqt5x11extras5-dev libphonon4qt5-dev libphonon4qt5experimental-dev g++-6
```
Dessa, line 65: why?

Stanzilla: the xenial image has a current cmake so we no longer need to install it

AlexanderP:
Hi. need to install cmake. I can only fix it in git in the evening

сб, 29 дек. 2018 г., 8:08 Benjamin Staneck notifications@github.com:

.travis.yml
AlexanderP: cmake is not installed via dpkg...

Stanzilla: this page says it is? https://docs.travis-ci.com/user/reference/xenial/#compilers-and-build-toolchain

AlexanderP:
I think this will fix the error:

```
diff --git a/dist/build_debian.sh b/dist/build_debian.sh
index d78ee50fc..8a159150a 100644
--- a/dist/build_debian.sh
+++ b/dist/build_debian.sh
@@ -32,7 +32,7 @@ VERSION='4:5.0.0'
 VERSION1='5.0.0'
 TMPFILE=$(mktemp)
 TMPGPG=$(mktemp)
-DIST_PPA="trusty xenial artful bionic"
+DIST_PPA="xenial bionic cosmic"
 PPANAME=kvirc
 
 dchppa_pkg(){
@@ -66,13 +66,6 @@ incoming = ~kvirc/kvirc/ubuntu/
 login = anonymous
 allow_unsigned_uploads = 0
 
-[kvirc-qt5.5]
-fqdn = ppa.launchpad.net
-method = ftp
-incoming = ~kvirc/kvirc-qt5.5/ubuntu/
-login = anonymous
-allow_unsigned_uploads = 0
-
 EOF
 }
 
@@ -121,16 +114,13 @@ dch -a "Date: $datct"
 
 if [ -z "$PPA" ]
 then
+    sed -i '/^ cmake \(>= 3.1\),/d' debian/control
     debuild --no-lintian -us -uc
 else
     gpgkey
     tmpgpg
     test -f ~/.dput.cf || dputcf
     dchppa_pkg
-    rm -f ../*ppa*
-    DIST_PPA="xenial"
-    PPANAME=kvirc-qt5.5
-    dchppa_pkg
 fi
 
 
diff --git a/dist/debian/control b/dist/debian/control
index 1cee7f995..43a261e60 100644
--- a/dist/debian/control
+++ b/dist/debian/control
@@ -14,8 +14,8 @@ Build-Depends: debhelper (>= 9),
  pkg-config,
  libxss-dev,
  python-dev,
- libqt5webkit5-dev, libenchant-dev
-Standards-Version: 4.1.4
+ libenchant-dev
+Standards-Version: 4.2.1
 Homepage: http://www.kvirc.de/
 
 Package: kvirc

```

AlexanderP:
hi. deb package build fixed

.travis.yml
AlexanderP: debuild clears variables when building, for example `PATH`

PR #2415, opened 2018-12-02
Implement mIRC extended colors (16-98)

CyberShadow:
These are described here:

https://www.mirc.com/colors.html
https://modern.ircdocs.horse/formatting.html#colors-16-98


src/kvilib/irc/KviControlCodes.cpp
```diff
@@ -306,4 +306,102 @@ namespace KviControlCodes
...
+			0x4d4d4d,
+			0x656565,
+			0x818181,
+			0x9f9f9f,
+			0xbcbcbc,
+			0xe2e2e2,
+			0xffffff,
+		};
+		if (index < minColor || index > maxColor)
+			return (kvi_u32_t)-1;
```
IceN9ne, line 404: It would probably be better to return the default color.

src/kvilib/irc/KviControlCodes.cpp
```diff
@@ -306,4 +306,102 @@ namespace KviControlCodes
 		}
 	}
 #endif
+
+	// Get extended (16-98) mIRC color.
+	// Unlike the 0-15 ones, these are not configurable.
+	// https://modern.ircdocs.horse/formatting.html#colors-16-98
+	// Returns (kvi_u32_t)-1 if index is out of bounds.
+	kvi_u32_t getExtendedColor(int index)
```
IceN9ne, line 314: Could we just extend the original color array and include them all in one place?  Seems like wasted operations to have to determine if it's above or below 16 with special indexing to account for missing the first 16 values.

src/kvilib/irc/KviControlCodes.cpp
CyberShadow: Generally, signaling an error or returning an invalid value is better when handling invalid input (which should have been caught / filtered out by the caller).

Note that this code is never actually currently executed, and is there just for future-proofing.

I just noticed that `KVI_ASSERT` exists, so I can add that if you think so.

src/kvilib/irc/KviControlCodes.cpp
CyberShadow: The original color array is a variable user-configurable array of `KviColorOption`. That doesn't look like an appropriate place for hard-coded, immutable color values.

IceN9ne: Why can't these be user-configurable as well?

CyberShadow: I don't think they should; unlike the 0-15 colors, which have a much stronger semantic meaning (especially values 0 and 1), these are just points along the hue/luminosity gradient plus some grayscales. It doesn't make more sense to configure these any more than making RGB color codes somehow configurable (if we were to ever add those).

Stanzilla: @IceN9ne agree or disagree with that?

IceN9ne: It's fine either way to me; I was just suggesting an idea.

Stanzilla:
@CyberShadow status?

CyberShadow:
I'm generally happy with the implementation. Have been using a build with this PR since it was open.

src/kvilib/irc/KviControlCodes.cpp
Stanzilla: @IceN9ne ?

Stanzilla:
@IceN9ne let's do it?

PR #2423, opened 2019-04-17
URL List fixes

CyberShadow:
I never used the URL list feature in mIRC and Kvirc, but always assumed that it existed so that you can open URLs with only the keyboard. Apparently I was wrong?

### DIFF
--- a/src/modules/url/libkviurl.cpp
+++ b/src/modules/url/libkviurl.cpp
@@ -140,7 +140,7 @@ UrlDialog::UrlDialog(std::unordered_set<KviUrl *>)
 	labels << __tr2qs("URL") << __tr2qs("Window") << __tr2qs("Count") << __tr2qs("Timestamp");
 	m_pUrlList->setHeaderLabels(labels);
 
-	connect(m_pUrlList, SIGNAL(itemDoubleClicked(QTreeWidgetItem *, int)), SLOT(dblclk_url(QTreeWidgetItem *, int)));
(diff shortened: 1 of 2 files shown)


Stanzilla:
It compiles and I have no problems with it, merge? @IceN9ne @DarthGandalf

IceN9ne:
I don't see anything wrong with the PR and it looks fine, but I couldn't reproduce the "Fix showing configuration dialog after closing it with Esc" issue. Maybe Linux wasn't affected by this.

My only comment would be that I'd rather see the Qt5 signal/slot connection syntax used, but it's fine as not all of them have been fixed anyways.

Merging it is fine.

CyberShadow:
> I don't see anything wrong with the PR and it looks fine, but I couldn't reproduce the "Fix showing configuration dialog after closing it with Esc" issue. Maybe Linux wasn't affected by this.

Hmm, I'm on Linux. Maybe it's specific to something else like the Qt version.

Probably a better fix is possible (intercepting what causes Esc to make Qt hide the dialog instead of closing it, and destroying it there), but it's still an improvement over not being able to summon it again for the session.

Stanzilla:
Yeah but that could be in a follow-up PR

IceN9ne:
> Hmm, I'm on Linux. Maybe it's specific to something else like the Qt version.

Well, if it fixed some builds without negatively affecting the ones that were working, that's a win in my book.

PR #2425, opened 2019-05-05
Add an option so a sunken frame is not drawn on our custom widgets.

wodim:
Stanzilla's request. Helps with dark Qt themes.


ctrlaltca:
Is this still relevant? Maybe it needs its option in OptionsWidget_theme?

ctrlaltca:
I'd close this by now. If it's still desiderable, it needs to be reworked anyway.

PR #2456, opened 2019-12-01
options/windowList: Add event names to alert level tips

staticfox:
This should make it more clear what each alert level will trigger on

https://i.imgur.com/QXP1MjY.png

### DIFF
--- a/src/modules/options/OptionsWidget_windowList.cpp
+++ b/src/modules/options/OptionsWidget_windowList.cpp
@@ -29,6 +29,9 @@
 #define _WANT_OPTION_FLAGS_
 #include "KviOptions.h"
 
+#include <string>
+#include <vector>
+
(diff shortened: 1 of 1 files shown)


AndrioCelos:
What if you reconfigure them?

src/modules/options/OptionsWidget_windowList.cpp
```diff
@@ -29,6 +29,9 @@
 #define _WANT_OPTION_FLAGS_
 #include "KviOptions.h"
 
+#include <string>
+#include <vector>
```
AndrioCelos, line 33: Does this compile?

staticfox: Why wouldn't it?

staticfox:
AV & Travis passed

PR #2461, opened 2020-02-17
KVS quit command aborts delayed reconnect

AndrioCelos:
This branch changes the KVS `quit` command so that it will abort a pending delayed reconnect if it exists, as the toolbar button does.
It also updates the documentation for `$context.state`, which can be used to check whether a reconnect is pending.

### DIFF
--- a/src/kvirc/kernel/KviIrcContext.cpp
+++ b/src/kvirc/kernel/KviIrcContext.cpp
@@ -828,7 +828,7 @@ void KviIrcContext::abortReconnect()
 	if(m_eState == KviIrcContext::PendingReconnection)
 	{
 		delete m_pReconnectTimer;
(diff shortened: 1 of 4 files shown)


src/kvirc/kernel/KviIrcContext.cpp
```diff
@@ -832,6 +823,19 @@ void KviIrcContext::terminateConnectionRequest(bool bForce, const QString & szQu
 	}
 }
 
+void KviIrcContext::abortReconnect()
+{
+	if(m_eState == KviIrcContext::PendingReconnection)
+	{
+		delete m_pReconnectTimer;
+		m_pReconnectTimer = 0;
```
IceN9ne: nullptr

PR #2485, opened 2020-08-22
Batch export logs

Leodmanx2:
I wanted to export all the logs from a channel so I could analyze data from one of the bots. When I went to do so, however, I was greeted with approximately 3000 save file dialogs. I dug into the code and found the following TODO on LogViewWindow.cpp:454.

`// TODO: probably should allow to specify a directory instead of asking for file path on each log file`

This pull request fulfills that TODO.

As it's possible to be exporting megabytes of data in the triple digits and there is I/O work that could hold up processing, I opted to use the QtConcurrent API to parallelize the work and avoid possibly delaying event processing as might happen if using QTimer timeouts.

This commit adds a standard, non-modal progress dialog as pictured below. The user can interact with the client and close the log viewer without interrupting the export.

kvircexport


src/modules/logview/LogViewWindow.cpp
```diff
@@ -386,10 +391,11 @@ void LogViewWindow::filterNext()
 	new LogListViewLog(m_pLastGroupItem, pFile->type(), pFile);
 
 filter_next:
-	pFile = m_logList.next();
+	++m_currentLog;
+	pFile = *m_currentLog;
```
ctrlaltca: This actually segfaults when at the last element.
In std::vector, end() returns the not allocated "past the last" element, that can't be dereferenced 
Since pFile is not used later in the code, the line can be safely removed.

ctrlaltca: fixed in https://github.com/kvirc/KVIrc/pull/2485/commits/31cc120f3ba988fbd7f31a4c5db684c641c3de2e

ctrlaltca:
Thank you very much for your contribution!

Issue #2494, opened 2021-01-26
Errors in manual or code in class painter


solmon4devx:
I found in the manual errors in the order of the parameters of four functions of the painter class.
In the function $setFont the help says from line 302 of file src/modules/objects/KvsObject_painter.cpp  
!fn: $setFont(<family:string>,<size:integer>[,<style:enum>,<style:enum>,..])[br]
		Set the font's family, size and style, valid flag for style are:[br]
the same as in class widget but in the code the order says:

	KVSO_PARAMETERS_BEGIN(c)
	KVSO_PARAMETER("size", KVS_PT_INTEGER, 0, iSize)
	KVSO_PARAMETER("family", KVS_PT_STRING, 0, szFamily)
	KVSO_PARAMETER("style", KVS_PT_STRINGLIST, KVS_PF_OPTIONAL, szListStyle)
	KVSO_PARAMETERS_END(c)

this was introduced in dbaac33

Maybe the correct solution is to change the order of the parameters to this function, so that the two classes widget and painter would have a setFont function with the same order of parameters.

The same occurs with the following three functions:
$drawPie(<x:integer>,<y:integer>,<w:unsigned integer>,<h:unsigned integer>,<angle:integer>,<alen:integer>)
$drawArc(<x:integer>,<y:integer>,<w:unsigned integer>,<h:unsigned integer>,<angle:integer>,<alen:integer>)
$drawChord(<x:integer>,<y:integer>,<w:unsigned integer>,<h:unsigned integer>,<angle:integer>,<alen:integer>)

in the code the order of the three is:
  (<angle:integer>,<alen:integer>,<x:integer>,<y:integer>,<w:unsigned integer>,<h:unsigned integer>)

I don't know when the order of the parameters of these last three was changed, the Grifisx/Noldor scripts used the order indicated in the manual, so the code was changed at some point.

wodim:
drawChord for example has been buggy since forever as far as I can see

wodim:
The correct solution indeed is to fix the code, not the docs, as the code adheres to the Qt api generally. For example:

>void drawPie(int x, int y, int width, int height, int startAngle, int spanAngle)

PR #2495, opened 2021-01-29
Fix for #2494

solmon4devx:
Restores parameters order as indicated in the manual for the functions of painter class:
   setFont, drawRoundRect, drawArc, drawChord and drawPie

### DIFF
--- a/src/modules/objects/KvsObject_painter.cpp
+++ b/src/modules/objects/KvsObject_painter.cpp
@@ -754,6 +754,66 @@ KVSO_END_CONSTRUCTOR(KvsObject_painter)
 		}                                                                                                                    \
 	}
 
+#define KVSO_6OR4ARRAYPARAMETERS(__pXOrArray,__iY,__iW,__iH,__iA1,__iA2)                                                                    \
+	if(__pXOrArray->isArray())                                                                                                              \
+	{                                                                                                                                       \
+		if(__pXOrArray->array()->size() < 4)                                                                                                \
+		{                                                                                                                                   \
+			c->error(__tr2qs_ctx("The array passed as parameter must contain at least 4 elements","objects"));                              \
+			return false;                                                                                                                   \
+		}                                                                                                                                   \
+		if(__pXOrArray->array()->size() < 6)                                                                                                \
+		{                                                                                                                                   \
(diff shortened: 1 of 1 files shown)


src/modules/objects/KvsObject_painter.cpp
```diff
@@ -754,15 +754,75 @@ KVSO_END_CONSTRUCTOR(KvsObject_painter)
...
+			c->error(__tr2qs_ctx("The array passed as parameter must contain at least 4 elements","objects"));                              \
+			return false;                                                                                                                   \
+		}                                                                                                                                   \
+		if(__pXOrArray->array()->size() < 6)                                                                                                \
+		{                                                                                                                                   \
+			if(c->params()->count() < 3)                                                                                                    \
+			{                                                                                                                               \
+				QString error=function+__tr2qs_ctx(" requires either an array as first parameter and 2 integers or six integers","objects");\
```
Arfrever: `__tr2qs_ctx` is for translation of message into other languages.
You are incorrectly assuming that subject of sentence (in this case, name of function) is at the beginning of sentence in all languages.

src/modules/objects/KvsObject_painter.cpp
```diff
@@ -754,15 +754,75 @@ KVSO_END_CONSTRUCTOR(KvsObject_painter)
...
+		{                                                                                                                                   \
+			c->error(__tr2qs_ctx("One of the geometry array parameters didn't evaluate to an integer","objects"));                          \
+			return false;                                                                                                                   \
+		}                                                                                                                                   \
+	} else {                                                                                                                                \
+		if(c->params()->count() < 6)                                                                                                        \
+		{                                                                                                                                   \
+			QString error=function+__tr2qs_ctx(" requires either an array as first parameter or six integers","objects");                   \
```
Arfrever: Same here.

src/modules/objects/KvsObject_painter.cpp
solmon4devx: I know how to use __tr2qs_ctx I was just trying to stick as close as possible to the style of the rest of the file, but you are right.

solmon4devx:
I think this will not be accepted, so I close the pull request.

PR #2511, opened 2021-06-12
Document the AVATAR negotiation protocol

SoniEx2:
This could be better solved by somehow being able to add user/custom message tags to the WHO output, but the legacy protocol should still be documented.

### DIFF
--- a/src/kvilib/irc/KviAvatar.cpp
+++ b/src/kvilib/irc/KviAvatar.cpp
@@ -57,9 +57,9 @@
 		are two users: A and B.[br]
 		When user A wants to see the B's avatar he simply sends a CTCP AVATAR request
 		to B (the request is sent through a PRIVMSG IRC command). The presence of the avatar
-		is announced by the ASCII sequence ETX, '0'-'7', SI, which doesn't show up on IRC
-		clients, at the start of B's realname, where bit 2 (the fours place) of the ASCII
-		digit is set when an avatar is present. (Bits 0 and 1 are unrelated to avatar
+		is announced by the ASCII sequence ETX, one of '0'-'7', SI, which doesn't show up
+		on IRC clients, at the start of B's realname, where bit 2 (the fours place) of the
+		ASCII digit is set when an avatar is present. (Bits 0 and 1 are unrelated to avatar
 		negotiation and should be left unset if not being used.)[br]
 		User B replies with a CTCP AVATAR notification (sent through a NOTICE IRC command)
 		with the name or URL of his avatar.[br]
--- a/src/kvilib/irc/KviAvatar.cpp
+++ b/src/kvilib/irc/KviAvatar.cpp
@@ -57,8 +57,9 @@
 		are two users: A and B.[br]
 		When user A wants to see the B's avatar he simply sends a CTCP AVATAR request
 		to B (the request is sent through a PRIVMSG IRC command). This is negotiated by
-		hiding an U+0003 '4' U+000F sequence at the start of B's realname, which doesn't
-		show up on IRC clients.[br]
+		hiding an U+0003 '4' U+000F, U+0003 '5' U+000F, U+0003 '6' U+000F or
+		U+0003 '7' U+000F sequence (with values other than '4' being deprecated) at the
+		start of B's realname, which doesn't show up on IRC clients.[br]
 		User B replies with a CTCP AVATAR notification (sent through a NOTICE IRC command)
 		with the name or URL of his avatar.[br]
 		The actual syntax for the notification is:[br]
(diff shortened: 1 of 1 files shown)


src/kvilib/irc/KviAvatar.cpp
```diff
@@ -56,7 +56,9 @@
 		Every IRC user has a client-side property called AVATAR. Let's say that there
 		are two users: A and B.[br]
 		When user A wants to see the B's avatar he simply sends a CTCP AVATAR request
-		to B (the request is sent through a PRIVMSG IRC command).[br]
+		to B (the request is sent through a PRIVMSG IRC command). This is negotiated by
+		hiding an U+0003 '4' U+000F sequence at the start of B's realname, which doesn't
```
DarthGandalf: '4' is not always there. The exact number depends on the gender set. See https://github.com/DarthGandalf/znc/blob/dgmods/modules/kvirc.cpp#L40-L43 where I reimplemented parts of this

SoniEx2: The "gender" idea, especially with no privacy/mutability options, needs to be deprecated. But okay, providing compatibility with those isn't too bad we guess.

DarthGandalf: Sorry but I'm not merging this. If you don't want to use gender, don't, but the feature is not going anywhere until there is a valid replacement.

DarthGandalf:
If you want to document this, please do, just don't try to sneak in the deprecation of a feature which you disagree with.

Re privacy: this is opt-in. If user doesn't want others to learn the gender, just don't configure one.

SoniEx2:
Hmm...

Y'know, fair enough, this pull request isn't about the lower bits. So let's not make it about them.

They can be documented later. Elsewhere. This is about avatars.

wodim:
There are several problems with this change:

1. It's written very awkwardly
2. It's not a bitfield and they are not bits
3. You should not use unicode notation because they are not unicode codepoints but bytes

src/kvilib/irc/KviAvatar.cpp
```diff
@@ -56,7 +56,11 @@
 		Every IRC user has a client-side property called AVATAR. Let's say that there
 		are two users: A and B.[br]
 		When user A wants to see the B's avatar he simply sends a CTCP AVATAR request
-		to B (the request is sent through a PRIVMSG IRC command).[br]
+		to B (the request is sent through a PRIVMSG IRC command). This is negotiated by
+		the ASCII sequence ETX, '0'-'7', SI, which doesn't show up on IRC clients, at the
+		start of B's realname, where bit 2 of the ASCII digit is set when an avatar is
```
DarthGandalf: Should it be clarified that bit 2 is the one which does "+4", or is it clear enough as is? I sometimes confuse their numeric order, because little endian/big endian.

src/kvilib/irc/KviAvatar.cpp
```diff
@@ -56,7 +56,11 @@
 		Every IRC user has a client-side property called AVATAR. Let's say that there
 		are two users: A and B.[br]
 		When user A wants to see the B's avatar he simply sends a CTCP AVATAR request
-		to B (the request is sent through a PRIVMSG IRC command).[br]
+		to B (the request is sent through a PRIVMSG IRC command). This is negotiated by
```
DarthGandalf: The presence of the avatar is announced by ...

src/kvilib/irc/KviAvatar.cpp
DarthGandalf: bit 2 of the number represented by the decimal ASCII digit

SoniEx2: The convention is that the LSB is bit 0, regardless of byte or bit order. But guess it doesn't hurt to clarify anyway.

src/kvilib/irc/KviAvatar.cpp
SoniEx2: Done.

src/kvilib/irc/KviAvatar.cpp
```diff
@@ -56,7 +56,11 @@
 		Every IRC user has a client-side property called AVATAR. Let's say that there
 		are two users: A and B.[br]
 		When user A wants to see the B's avatar he simply sends a CTCP AVATAR request
-		to B (the request is sent through a PRIVMSG IRC command).[br]
+		to B (the request is sent through a PRIVMSG IRC command). The presence of the avatar
+		is announced by the ASCII sequence ETX, '0'-'7', SI, which doesn't show up on IRC
```
DarthGandalf: On a new reread, this looks like ETX 0 1 2 3 4 5 6 7 SI. Maybe clarify that it's a single digit?

PR #2518, opened 2021-11-26
Fixed up to build on MacOS 11.6 and restore SSL and Python support

horsten:
This fixes the build on MacOS 11.9 with current cmake 3.20.5 and latest Homebrew as of 20211126.

The problems were:
1) A workaround for a previous bug in cmake was breaking SSL builds
2) Python madness - see https://gitlab.kitware.com/cmake/cmake/-/issues/20165

Now builds and works beautifully, thanks for a nice IRC client (the only one on Mac?)


horsten:
Doesn't look like the autobuilder is working. My build is working fine but I've only tested Mac. Should be the same for all others with recent cmake. To build on MacOS I configure it with:

cmake -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5/ -DPython3_ROOT_DIR=/usr/local/opt/python

I'm releasing the resulting DMG on my fork repo in case it's useful for someone until this gets merged.

DarthGandalf:
Thanks for fix!

> Doesn't look like the autobuilder is working

Travis is now broken (by design), but noone took time to migrate the configs to github actions or some other CI

Appveyor also broken (not by design), but noone took time to figure out what's wrong with the windows build and fix it

PR #2526, opened 2022-06-26
QString::null is deprecated

nmariusp:


ctrlaltca:
This is quite safe and long overdue, thank you

PR #2540, opened 2023-06-17
appveyor: Try to fix builds

staticfox:
Experimenting, cygwin's last 32 bit build was 3.3.6. let's see what happens


DarthGandalf:
I think we can just drop 32 bit builds... Or rewrite this to not use cygwin at all. I only used cygwin here because I know bash better than powershell

ctrlaltca:
Superseded by #2542

PR #2541, opened 2023-07-06
Port to (and require) Qt 5.15. Drop KDE4 support, fix KDE5

ctrlaltca:
This PR removes the usage of all deprecated classes and methods in Qt5, with a single exception.
Only KvsObject_xmlreader is still using a deprecated Qt class, but it needs to be completely rewritten or deprecated aswell.
KDE4 support has been removed, and KDE5 support has been integrated where missing (eg. the terminal widget and the filetransfer "open file" menu).
This actually requires Qt 5.15, any recent (decent) distro is shipping it since years.
It also requires a quite recent version of KDE Frameworks is KDE support is enabled. Please report any problem compiling, i'll be glad to implement workarounds where possible.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -389,7 +389,7 @@ endif()
 # Check for Qt5
 ############################################################################
 
-find_package(Qt5Widgets)
+find_package(Qt5Widgets 5.15)
 if(Qt5Widgets_FOUND)
 	set(CMAKE_STATUS_QT_VERSION "${Qt5Widgets_VERSION_STRING}")
 	set(qt5_kvirc_modules)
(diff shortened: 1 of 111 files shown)


DarthGandalf:
This does build for me. However, something broke in the tree window list, it doesn't sort channels by name anymore.

DarthGandalf:
And it looks like it reset my fonts to default from my custom settings

DarthGandalf:
Not just fonts, but also message decorations

ctrlaltca:
Thank you, i'll look at these regressions

ctrlaltca:
@DarthGandalf both problems should be sorted now, can you please give it another test?

DarthGandalf:
Confirmed, those 2 are fixed.

Next issue: mouse scroll up/down broke in the channel tree list. It used to switch selected channel.

ctrlaltca:
Mouse scrolling fixed. By mistake I changed it to use horizontal scrolling instead of vertical.

PR #2552, opened 2023-08-09
Port the mediaplayer mpris interface to mpris 2.2; fix #1912

### LINKED ISSUE #1912: Features: MPRIS v2.2
Hi! 
KVIrc currently supports the MPRIS 1 API but is not MPRIS 2 capable.

More desktop applets (MPRIS Applet in GNOME and Xfce, xfce4-soundmenu-plugin in Xfce, Ayatana Sound Menu) are able to control only players supporting the latest version of the MPRIS standard (like Amarok, Banshee, Clementine, Guayadeque, Pragha, Rhythmbox, Spotify, VLC, xnoise, etc) and are not able to control MPRIS 1 capable players like Audacious.

I think it would make sense to implement an MPRIS 2 interface in KVIrc.

References:
https://specifications.freedesktop.org/mpris-spec/latest/

ctrlaltca:
This PR updates the mediaplayer module's MPRIS interface from version 1 to version 2.2 (https://specifications.freedesktop.org/mpris-spec/latest/)
Version 1 of the specification is considered obsolete and broken; all media players already updated to the new spec years ago.
Two dead mediaplayers got removed (Mozilla Songbird and BMPx).
This PR has been tested working with VLC, Clementine and Audacity.

A flaw has been found in the position/length handling: the variable holding the milliseconds value is an int, so it will report broken values for long files; from my tests it breaks after around 15 minutes. The fix is not contained in this PR since it requires a data type change in all the other interfaces, too.

### DIFF
--- a/src/modules/mediaplayer/MpMprisInterface.cpp
+++ b/src/modules/mediaplayer/MpMprisInterface.cpp
@@ -26,44 +26,9 @@
 #if(defined(COMPILE_DBUS_SUPPORT) && !defined(COMPILE_ON_WINDOWS) && !defined(COMPILE_ON_MAC) && !defined(COMPILE_ON_MINGW))
 #include "KviLocale.h"
 
-/*
-	according to MPRIS 1.0, GetStatus returns struct of 4 integers.
-
-	First integer: 0 = Playing, 1 = Paused, 2 = Stopped.
-	Second integer: 0 = Playing linearly, 1 = Playing randomly.
-	Third integer: 0 = Go to the next element once the current has finished playing, 1 = Repeat the current element
-	Fourth integer: 0 = Stop playing once the last element has been played, 1 = Never give up playing
-*/
-
-struct MPRISPlayerStatus
-{
-	int Play;
-	int Random;
-	int RepeatCurrent;
-	int RepeatPlaylist;
-};
-Q_DECLARE_METATYPE(MPRISPlayerStatus)
-
-QDBusArgument & operator<<(QDBusArgument & argument, const MPRISPlayerStatus & status)
-{
-	argument.beginStructure();
-	argument << status.Play << status.Random << status.RepeatCurrent << status.RepeatPlaylist;
-	argument.endStructure();
-	return argument;
-}
-
-const QDBusArgument & operator>>(const QDBusArgument & argument, MPRISPlayerStatus & status)
-{
-	argument.beginStructure();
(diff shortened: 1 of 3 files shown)


src/modules/mediaplayer/MpMprisInterface.cpp
```diff
@@ -83,8 +48,8 @@ int MpMprisInterface::detect(bool)
 }
 
 #define MPRIS_SIMPLE_CALL(__path, __action)                                           \
```
DarthGandalf: identifiers with double `_` ar reserved for compiler's own needs, so this may randomly break

src/modules/mediaplayer/MpMprisInterface.cpp
```diff
@@ -597,6 +375,6 @@ MP_IMPLEMENT_DESCRIPTOR(
 MpClementineInterface::MpClementineInterface()
     : MpMprisInterface()
 {
-	m_szServiceName = "org.mpris.clementine";
+	m_szServiceName = "org.mpris.MediaPlayer2.clementine";
```
DarthGandalf, line 378: I'm not familiar with this API (and too lazy to check), but shouldn't there be a way to check a generic media player, without listing all of possibilities? E.g. what KDE's lock screen or KDE's tray icon are doing

Currently at least `org.mpris.MediaPlayer2.strawberry` is missing from this list

ctrlaltca: yes, we could enumerate all services named `org.mpris.MediaPlayer2.*` and use them, since they are supposed to expose the same exact dbus interfaces.
In the past we had to use different service named for each player, that's why we have a different class for each of them. Now we could just implement a "generic mpris" class and check the available players at runtime.

src/modules/mediaplayer/MpMprisInterface.cpp
ctrlaltca: The new commit removes the double __

src/modules/mediaplayer/MpMprisInterface.cpp
ctrlaltca: The new commit adds 2 new interfaces. The first one is for strawberry, the second one is a generic fallback for any other player implementing a mpris2 interface.
At first i thought about removing the player-specific interfaces, but this would remove the ability to switch from one player to another from the options.

PR #2554, opened 2023-08-11
Replace QtWebKit with QtWebEngine; fix #1854

### LINKED ISSUE #1854: Port to qtwebengine for Qt-5.6

ctrlaltca:
Three items are affected by this change:

1) the Help Widge: it has been completely ported to QtWebEngine .
It can also work without it, falling back to QTextBrowser
2) the Addon/Theme package download  dialog:
It has been ported, and it's just as broken as it was before (the links on the website points to not-existing ftp.kvirc.de, uninstall never worked)
3) the webview object:
It has been ported, but all the functions used to manipulate elements on the DOM have been removed since QtWebEngine doesn't include them. Everything must be ported to use plain javascript.

Dropping QtWebkit is past due, since you don't want to browse the web with a 2016 browser.
QtWebEngine is still "old" on Qt5, and probably KVIrc would be better compiled with WANT_WEBKIT=no anyway.


ctrlaltca:
This is needed for Qt6 (qtwebkit doesn't exists anymore).

Issue #2529, opened 2022-07-30
Build for macOS >= 12.4


silv2k:
I have a MacBook M1 and used KVIrc until Apple decided to remove Python 2.7 with the update to macOS 12.4.

Is there a possibilty to run KVIrc without Python 2.7? 
Disabling SIP and installing Python 2.7 to the previous path is not an option.

I tried to make a build with homebrew and Python 3.10.5 by following the instructions for macOS, but the app crashs everytime at starting.
The report says it is a codesigning error.
I tried serveral command to resign the app, but nothing worked.

Can somebody make a build for macOS without Python 2.7 or help me to troubleshoot my bugs?

Thank you

```
-------------------------------------

Translated Report (Full Report Below)

-------------------------------------

Process: kvirc [473]

Path: /Users/USER/Downloads/*/KVIrc.app/Contents/MacOS/kvirc

Identifier: com.kvirc.kvirc

Version: 5.0.1 (???)

Code Type: ARM-64 (Native)

Parent Process: launchd [1]

User ID: 501

Date/Time: 2022-07-31 00:07:30.8322 +0200

OS Version: macOS 12.5 (21G72)

Report Version: 12

Anonymous UUID: 668A36F9-98A4-4E28-B055-A6509AB4E253

Time Awake Since Boot: 36 seconds

System Integrity Protection: enabled

Crashed Thread: 0 Dispatch queue: com.apple.main-thread

Exception Type: EXCBADACCESS (SIGKILL (Code Signature Invalid))

Exception Codes: UNKNOWN_0x32 at 0x0000000103904000

Exception Codes: 0x0000000000000032, 0x0000000103904000

Exception Note: EXCCORPSENOTIFY

Termination Reason: Namespace CODESIGNING, Code 2

**Comment by emarsk** [comment]
I installed Python2 with homebrew via [this tap](https://github.com/kamilturek/homebrew-python2), then I linked the framework to a location where the pre-built KVIrc is expecting it:

```
sudo ln -svfn /usr/local/Cellar/python@2/2.7.18/Frameworks/Python.framework /Library/Frameworks/
```

`/Library` isn't locked by SIP, so this should work for you as well.

---

I also tried a build with Python3, following [the instructions here](https://github.com/kvirc/KVIrc/wiki/Compiling-KVIrc-on-macOS) (note that the path to qt5 is now `/usr/local/opt/qt@5/`). It works for me, no errors.

silv2k:
Thank you for the advice.
[...]

PR #2556, opened 2023-08-30
Macos: manually create dmg to fix icon position

ctrlaltca:
Reference: https://github.com/kvirc/KVIrc/issues/2529#issuecomment-1675798351
Creating this PR to check if the build succeeds


ctrlaltca:
Confirmed working

PR #2557, opened 2023-10-05
[RFC] Add support for Qt6

ctrlaltca:
This PR contains all the needed bits to get KVIrc to compile with both Qt5 and Qt6.
Big changes:
 * QRegExp usage has been ported to QRegularExpression. To avoid disruptive (and future) changes, a new KviRegExp class has been created that mimics the old interface already used in KVIrc;
 * QChar and QString are more picky when casting from an integer/char, a lot of explicit casts had to be added;
 * obligatory class and method renames to follow qt's quest for the perfect word;
 * QKeySequence (used for shortcuts) doesn't like anymore the classic Qt::Modifier + Qt::Key style, so a bit of rework had to be done to keep the definitions compatible with both Qt5 and Qt6
 
Changes that will need some rework (last commit):
 * QFont changed the weight scale in Qt6, in order to use the new values we'd need to fix all configuration files in all themes. By now, it uses the legacy deprecated values.
* QTextCodec is depecated, and the new QStreamConverter only supports Utf8/16/32 and Latin1 before Qt6.5. By now we stick to qt5compat's QTextCodec.

Other warnings that needs fixing:
 * all QMessageBox calls needs to be rewritten using a different ctor (the one with StandardButtons);
 * QVariant::type() is deprecated

The plain is to get this working on CI, collect any comment and maybe start getting the good commits one by one in master.


CMakeLists.txt
```diff
@@ -386,18 +386,21 @@ if(ZLIB_FOUND)
...
+# Check for Qt
 ############################################################################
 
-find_package(Qt5Widgets 5.15)
-if(Qt5Widgets_FOUND)
-	set(CMAKE_STATUS_QT_VERSION "${Qt5Widgets_VERSION_STRING}")
-	set(qt5_kvirc_modules)
+# first check if Qt6 or Qt5 has to be used
+find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core)
+# second, detect available modules for the specific Qt version
```
DarthGandalf, line 394: Why not do both components in the same find_package? `COMPONENTS Core Widgets`

src/modules/str/libkvistr.cpp
```diff
@@ -1721,41 +1721,109 @@ static bool str_kvs_fnc_split(KviKvsModuleFunctionCall * c)
...
+	if(szSep.isEmpty())
 	{
-		a->append(new KviKvsVariant{szStr});
+		a->set(0, new KviKvsVariant(szStr));
 		return true;
 	}
 
-	bool bWild = szFla.contains('w', Qt::CaseInsensitive);
-	bool bContainsR = szFla.contains('r', Qt::CaseInsensitive);
+	if(iMaxItems == 0)
```
DarthGandalf, line 1735: Why did you move this condition down?

src/modules/upnp/XmlFunctions.cpp
```diff
@@ -112,7 +112,11 @@ QString XmlFunctions::getNodeValue(const QDomNode & rootNode, const QString & pa
 QString XmlFunctions::getSource(const QDomNode & node, int indent)
 {
 	QString source;
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 	QTextStream textStream(&source, QIODevice::WriteOnly);
+#else
+	QTextStream textStream(&source, QIODeviceBase::WriteOnly);
```
DarthGandalf: QIODevice::WriteOnly should still work in Qt6, because QIODevice inherits from QIODeviceBase

src/modules/str/libkvistr.cpp
```diff
@@ -1721,41 +1721,109 @@ static bool str_kvs_fnc_split(KviKvsModuleFunctionCall * c)
...
+				}
+
+				iMatch += len;
+				iBegin = iMatch;
+			}
+		}
+	}
 	else
-		list = szStr.splitRef(szSep, splitBehavior, sensitivity);
+	{
```
DarthGandalf, line 1783: Lots of changes in this function. I don't see in the PR description what this is for, so I don't know how to review this change

DarthGandalf (commented):
We can easily require Qt6.5+ for the Qt6 build, that shouldn't be a problem, I think

CMakeLists.txt
ctrlaltca: The first call is only used to determine if Qt5 or Qt6 will be used, and define `${QT_VERSION_MAJOR}` accordingly.
All the other calls force the specific Qt version using `find_package(Qt${QT_VERSION_MAJOR} ..)`.

This way of support both Qt5 and Qt6 without mixing them is documented here: https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html

src/modules/str/libkvistr.cpp
ctrlaltca: IceN9ne's rewrote  $str.split() in 2017 making use of QStringRef, that unfortunately doesn't exists anymore in Qt6. 
I reverted a part of this commit: https://github.com/kvirc/KVIrc/commit/9e8a9d9e0d38993e145dddab64b42114efc4d3f6#diff-2c8a6d9d89531ab98f04a8dce7a8d6670d4661adac0b78edd4b5a38941890befR1717

src/modules/str/libkvistr.cpp
ctrlaltca: Sorry, I missed it in the PR description. 

IceN9ne's rewrote $str.split() in 2017 making use of QStringRef / QString::splitRef(), but they unfortunately doesn't exists anymore in Qt6.
I reverted a part of this commit: https://github.com/kvirc/KVIrc/commit/9e8a9d9e0d38993e145dddab64b42114efc4d3f6#diff-2c8a6d9d89531ab98f04a8dce7a8d6670d4661adac0b78edd4b5a38941890befR1717
The previous version of $str.split() wasn't making use of those Qt features and could be ported without further changes

src/modules/upnp/XmlFunctions.cpp
ctrlaltca:
```
src/modules/upnp/XmlFunctions.cpp:115:52: error: incomplete type ‘QIODevice’ used in nested name specifier
  115 |         QTextStream textStream(&source, QIODevice::WriteOnly);
```
You are right, this could be fixed with just an `#include <QIODevice>`

ctrlaltca: Done in https://github.com/kvirc/KVIrc/pull/2557/commits/4e922196f409363439c65facfaef3605671716db

PR #2561, opened 2023-11-11
Fix text[idx] out of bounds assertion in KviTopicWidget::paintColoredText

i-garrison:
Just restructure the loop so index is always in bounds + extract `isKviControlCode()` helper to make it look nicer.


ctrlaltca (approved):
Looks good to me

ctrlaltca:
Merged, thank you!

PR #2562, opened 2023-11-11
Fix KviTopicWidget label shifted down and clipped

i-garrison:
`KviThemedLabel` has margin set to 4. While sizing topic widget this is not taken into account. This causes rendered topic label text to be shifted down and clipped at the bottom.

Fix this by setting `KviThemedLabel` margin to zero and using stylesheet left and right margin properties instead.

While at it, simplify management of editor widget lifecycle and widget size. Instead of doing that by hand on events, create `QStackedWidget` to operate visibility and size of label and editing widgets. Create all objects at once in `KviTopicWidget` constructor, attach label widget at index 0 and a `KViTaiHbox` with line editor and 3 buttons at index 1. Only operate visibility by switching current stacked widget index and let `QStackedWidget` resize it's contents as needed.


ctrlaltca (approved):
Looks good to me

ctrlaltca:
Merged, thank you!

PR #2563, opened 2023-11-11
Do not leak each created KviInputEditorSpellCheckerBlock

i-garrison:
Objects allocated by `splitTextIntoSpellCheckerBlocks` into buffer vector are never freed. Fix this by turning vector of pointers into vector of objects.


ctrlaltca (approved):
Nice catch. Looks good to me

ctrlaltca:
Merged, thank you!

Issue #2564, opened 2023-11-19
ASAN report


freedom1b2830:
execute 
```sh
LD_PRELOAD=asan.so kvirc
```

# Report 1
execute cmd:
```
/chanserv help 
```
see 
```
=================================================================
==230325==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200021e01f at pc 0x7f130a6818df bp 0x7ffcbde201b0 sp 0x7ffcbde1f958
READ of size 16 at 0x60200021e01f thread T0
    #0 0x7f130a6818de in __interceptor_strlen /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:461
    #1 0x5638202c9521 in KviIrcConnection::sendData(char const*, int) (/usr/bin/kvirc+0x320521) (BuildId: b59d19250aa912da724fd622d8ec81dd7a3e775b)
    #2 0x56382020b4d8 in KviKvsCoreSimpleCommands::raw(KviKvsRunTimeContext*, KviKvsVariantList*, KviKvsSwitchList*) (/usr/bin/kvirc+0x2624d8) (BuildId: b59d19250aa912da724fd622d8ec81dd7a3e775b)
    #3 0x563820271322 in KviKvsTreeNodeCoreSimpleCommand::execute(KviKvsRunTimeContext*) (/usr/bin/kvirc+0x2c8322) (BuildId: b59d19250aa912da724fd622d8ec81dd7a3e775b)
    #4 0x56382022c45c in KviKvsScript::executeInternal(KviKvsRunTimeContext*) (/usr/bin/kvirc+0x28345c) (BuildId: b59d19250aa912da724fd622d8ec81dd7a3e775b)
    #5 0x56382022c59f in KviKvsScript::execute(KviWindow*, KviKvsVariantList*, KviKvsVariant*, int, KviKvsExtendedRunTimeData*) (/usr/bin/kvirc+0x28359f) (BuildId: b59d19250aa912da724fd622d8ec81dd7a3e775b)
    #6 0x56382026ceb1 in KviKvsTreeNodeAliasSimpleCommand::execute(KviKvsRunTimeContext*) (/usr/bin/kvirc+0x2c3eb1) (BuildId: b59d19250aa912da724fd622d8ec81dd7a3e775b)
[...]

PR #2502, opened 2021-03-19
Fix #2478

Leodmanx2:
My client was crashing every day or two. Putting it through a debugger immediately revealed an access violation. I gave that a quick patch and then let it run again until it crashed. When it eventually did, the problem turned out to be #2478.

I fixed what appears to be the cause, but I can't be 100% sure it is as I'm not currently free to disable WAN access to reproduce the issue. Firewalling the client on my local machine turned out to be insufficient and I didn't want to bother trying to trigger it in the code. That said, I was able to run the client for a full week without another crash, so I'm pretty confident. And either way, both commits fix things that need to be fixed.

PR #2565, opened 2023-11-20
Avoid sending a spurious null byte to the server at the end of raw commands

ctrlaltca:
While investigating #2564 it was noticed that raw commands always add a spurious null byte to the message sent to the server.
This null byte was added in #2502  to fix a possible crash during a strlen() call used to create a QString used by the socketspy.
This PR reworks the original fix: instead of appending a null byte, create the QString with an explicit length.

### DIFF
--- a/src/kvirc/kernel/KviIrcConnection.cpp
+++ b/src/kvirc/kernel/KviIrcConnection.cpp
@@ -832,14 +832,12 @@ bool KviIrcConnection::sendData(const char * pcBuffer, int iBuflen)
 			m_pConsole->outputNoFmt(KVI_OUT_SOCKETWARNING, __tr2qs("[LINK WARNING]: Socket message truncated to 512 bytes."));
 	}
 
-	KviDataBuffer * pData = new KviDataBuffer(iBuflen + 3);
+	KviDataBuffer * pData = new KviDataBuffer(iBuflen + 2);
 	KviMemory::move(pData->data(), pcBuffer, iBuflen);
 	*(pData->data() + iBuflen) = '\r';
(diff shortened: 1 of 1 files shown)


src/kvirc/kernel/KviIrcConnection.cpp
```diff
@@ -832,14 +832,12 @@ bool KviIrcConnection::sendData(const char * pcBuffer, int iBuflen)
...
-	KviDataBuffer * pData = new KviDataBuffer(iBuflen + 3);
+	KviDataBuffer * pData = new KviDataBuffer(iBuflen + 2);
 	KviMemory::move(pData->data(), pcBuffer, iBuflen);
 	*(pData->data() + iBuflen) = '\r';
 	*(pData->data() + iBuflen + 1) = '\n';
-	*(pData->data() + iBuflen + 2) = '\0';
 
-	QString szMsg = (const char *)(pData->data());
-	szMsg.truncate(iBuflen);
+	QString szMsg = QString::fromUtf8((const char *)(pData->data()), iBuflen);
```
DarthGandalf, line 840: Is it guaranteed to be utf8?

DarthGandalf (approved):

src/kvirc/kernel/KviIrcConnection.cpp
ctrlaltca: It's a raw string (byte sequence) coming from the user (/quote, /raw), we can't be 100% sure of its encoding.
The  implicit ctor used before internally called QString::fromUtf8(), so there's no change in behavior: https://doc.qt.io/qt-5/qstring.html#QString-7

PR #2568, opened 2023-11-22
Settings: include styleName in font; fix #2567

### LINKED ISSUE #2567: Font selection issue
Gentoo Linux amd64, version 4a51c0e82d066349c0450a70d010c4e4e1d6ec00
Qt 5.15.11

Выделение_488

I was trying to report for #2566 which font I'm using , and clicked the Verdana button on the screenshot. The button has the same font as what I'm actually using. However, the font selection window claims that I have different font settings - the window claims Bold, while I'm using Italic.

ctrlaltca:
KviStringConversion: is used to convert a font from/to a string for settings storage.
This PR adds "styleName" to the font properties saved in the settings file.
This property is needed by QFontDialog to show the correct font style (eg. italic, bold, etc) for certain fonts.
One example of this font is MS' Verdana, that includes localized version for the style names.

### DIFF
--- a/src/kvilib/ext/KviStringConversion.cpp
+++ b/src/kvilib/ext/KviStringConversion.cpp
@@ -226,13 +226,6 @@ namespace KviStringConversion
 	void toString(const QFont & font, QString & szBuffer)
 	{
 		QString szFamily(font.family());
-		szBuffer = QString::asprintf("%s,%d,%d,%d", szFamily.toUtf8().data(), font.pointSize(), font.styleHint(),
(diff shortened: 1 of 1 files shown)


DarthGandalf:
This does help... as long as user doesn't change the system locale. Because the localized string "Курсив"/"Italic" is saved in settings. When I flip between different LC_ALL, I see the same behavior as without this fix.

I guess that's still better than what we have now :(

ctrlaltca:
Correct, as long as the font uses localized style names there's no way we can support switching locales.

PR #2570, opened 2023-11-24
Remove QWebEngine usage in core. Rewrite web package manager

### LINKED ISSUE #1733: Let KVIrc's Addon web repository draw from Github
Instead of checking http://www.kvirc.de/app/ (because you know, that gets updated way too much so we should really give it a break :smile:), let's just generate a static list and put it in the root directory of https://github.com/kvirc/kvirc-scripts and have KVIrc download the addons from the repository... I'm not sure if Github has a policy on letting people use them as a file server though. Just an idea. That or I could set up a mirror on my box of the repository and we can have clients download off of that. Either way, the addons manager could use some love.

ctrlaltca:
This PR introduces 2 big modifications.

The first one is related to the help module, that right now can use as its backend Qt's QTextBrowser (a simple html browser) or QWebEngine (based on Chromium, supports javascript etc, usually quite outdated and with a bad security record).
This PR removes the QWEbEngine backend from the help module, since we don't need it features, it's more bloated, and it never appeared in any windows release anyway.

The second one is related to WebPackageManager, aka the "get more themes" and "get more addons" features.
Right now they use QWebEngine to show a page from www.kvirc.net and download files via ftp.
This feature is broken since ages because ftp downloads aren't supported anymore and sometimes because the www.kvirc.net website has a broken ssl cert.
This PR rewrites the WebPackageManager to use plain http requests to a json file containing the information about themes/addons. Right now the file is loaded from http://127.0.0.1/ for local testing, but the target is to host the files in a github repo (implementing #1733) so that contributors can use issues and pull requests.

Once these changes are in, the only usage of QWebEngine left is in the "object" module.

Prerequisite to fix #1733
Note: this PR needs an additional commit to fix the theme/addon urls once the repositories are created.

### DIFF
--- a/src/kvirc/ui/KviWebPackageManagementDialog.cpp
+++ b/src/kvirc/ui/KviWebPackageManagementDialog.cpp
@@ -26,8 +26,6 @@
 
 #include "KviWebPackageManagementDialog.h"
 
-#ifdef COMPILE_WEBENGINE_SUPPORT
-
 #include "KviLocale.h"
 #include "KviApplication.h"
 #include "KviMessageBox.h"
@@ -39,12 +37,9 @@
 
 #include <QScreen>
 #include <QToolButton>
-#include <QLineEdit>
 #include <QToolTip>
-#include <QTimer>
 #include <QClipboard>
 #include <QHBoxLayout>
-#include <QWebEngineView>
 #include <QVBoxLayout>
 #include <QToolBar>
 #include <QFile>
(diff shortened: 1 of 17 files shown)


src/kvirc/ui/KviWebPackageManagementDialog.cpp
```diff
@@ -39,19 +37,89 @@
...
+				setIcon(pixImage);
+			}
+		}
+	});
+}
+
+void KviWebPackageListItem::showPopupImage()
+{
+	QDialog dlg;
+	QHBoxLayout *l = new QHBoxLayout(&dlg);
```
DarthGandalf: Can you use consistent way of declaring pointers (wrt spaces)?

Probably whatever clang-format does

src/kvirc/ui/KviWebPackageManagementDialog.cpp
```diff
@@ -60,125 +128,162 @@ KviWebPackageManagementDialog::KviWebPackageManagementDialog(QWidget * pParent)
...
+	m_pItemDelegate->setIconSize(QSize(64, 64));
+	m_pListWidget->setItemDelegate(m_pItemDelegate);
+	m_pListWidget->setMinimumHeight(400);
+	m_pListWidget->setMinimumWidth(520);
+
+	m_pListWidget->setSelectionMode(QAbstractItemView::SingleSelection);
+	m_pListWidget->setSortingEnabled(true);
+	//connect(m_pListWidget, SIGNAL(itemDoubleClicked(QListWidgetItem *)), this, SLOT(applyTheme(QListWidgetItem *)));
+
+	connect(m_pListWidget, SIGNAL(customContextMenuRequested(const QPoint &)),
```
DarthGandalf, line 173: Above you used &pointer to function to connect to signal?

DarthGandalf (commented):
What do I need to serve on 127.0.0.1 to test this?

src/kvirc/ui/KviWebPackageManagementDialog.cpp
ctrlaltca: Done, thank you

src/kvirc/ui/KviWebPackageManagementDialog.cpp
ctrlaltca: Sorry, I don't get the question.
The signature of the signal is https://doc.qt.io/qt-6/qwidget.html#customContextMenuRequested
It's a const reference (&) to a QPoint object

DarthGandalf: I meant using `SIGNAL(bar())` vs using `&Foo::bar`

ctrlaltca: The different connect() style is needed to be able to use a lambda function a the slot

PR #2571, opened 2023-11-25
Test adding a linter to github actions

ctrlaltca:


ctrlaltca:
Lol it fails at its own default config

PR #2575, opened 2023-12-04
SSL updates

ctrlaltca:
This PR includes some updates to SSL:
* disable old protocols SSLv2, SSLv3, TLS 1.0 and TLS 1.1.
* disable some unsecure features on old OpenSSL version
* fix the use of ADH for DCC with recent openssl
* permit the use of anonymous DH only on DCC, not with servers
* set the server hostname inside the hello packet to support SNI (a server with multiple certificates)

### DIFF
--- a/src/kvilib/net/KviSSL.cpp
+++ b/src/kvilib/net/KviSSL.cpp
@@ -378,16 +378,43 @@ bool KviSSL::initContext(Method m)
 		SSL_CTX_set_verify(m_pSSLCtx, SSL_VERIFY_PEER, verify_clientCallback);
 	}
 
+	SSL_CTX_set_options (m_pSSLCtx,
+		// disable old, unsecure protocols
+		SSL_OP_NO_SSLv2|SSL_OP_NO_SSLv3
+		|SSL_OP_NO_TLSv1|SSL_OP_NO_TLSv1_1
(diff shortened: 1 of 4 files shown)


src/kvilib/net/KviSSL.cpp
```diff
@@ -378,11 +378,23 @@ bool KviSSL::initContext(Method m)
...
+		|SSL_OP_NO_COMPRESSION
+#endif
+#ifdef SSL_OP_SINGLE_DH_USE
+		|SSL_OP_SINGLE_DH_USE
+#endif
+#ifdef SSL_OP_SINGLE_ECDH_USE
+		|SSL_OP_SINGLE_ECDH_USE
+#endif
+	);
 	// we want all ciphers to be available here, except insecure ones, orderer by strength;
```
DarthGandalf, line 396: This comment isn't true anymore

DarthGandalf (approved):

src/kvilib/net/KviSSL.cpp
ctrlaltca: the following line disables insecure ciphers and sorts them

DarthGandalf: oh right, I misread the comment as "even insecure ones"...

PR #2579, opened 2023-12-08
Fix regression in some shortcuts 

ctrlaltca:
Eg. alt+enter to open the multi line editor.
Reported by @DarthGandalf on IRC

Regression introduced in cba0ebedf3f0cc4da6f8729dde10282b23eb18e6


ValeZAA:
You do realise this app is unusable on Windows? It crashes in 10 minutes after log in always...

ctrlaltca:
@ValeZAA while i appreciate your feedback, why are you commenting on this PR, that's related to shortcuts?
We just fixed a crash bug that's possibly what you are experiencing, we are currently building a new release with the fix included. It will be available in less than one hour, please give a test run at https://github.com/kvirc/KVIrc/releases/tag/5.2.0-beta2 .

PR #2585, opened 2023-12-25
Fix build with KDE Frameworks 6

alebastr:
Tested on a real Plasma 6.0 Beta 2, notifications *mostly* work (with exception of "View" being unable to focus the window).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -538,22 +538,29 @@ if(WANT_KDE)
 	find_package(ECM 1.0.0 QUIET NO_MODULE)
 	if(DEFINED ECM_VERSION_MAJOR)
 		set(CMAKE_MODULE_PATH ${ECM_MODULE_PATH} ${ECM_KDE_MODULE_DIR} ${CMAKE_MODULE_PATH})
+		set(KDE_COMPONENTS
+			CoreAddons      # KAboutData
+			I18n            # KLocalizedString
+			XmlGui          # KMainWindow
+			WindowSystem    # KWindowSystem
+			Notifications   # KNotification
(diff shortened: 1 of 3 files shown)


ctrlaltca:
Thank you!
I'm currently building plasma 6 beta 2 (well.. since this morning, to be honest). Once i get it up running I'll test the PR.

ctrlaltca:
Tested, it fixed compilation and works fine. Thanks again!

alebastr:
As for the window activation, that's how NeoChat does the trick:

[`NotificationsManager::postNotification`](https://invent.kde.org/network/neochat/-/blob/29321aeaf3361becdf5bf65aa9b01a721f4307c1/src/notificationsmanager.cpp#L182),
[`WindowController::showAndRaiseWindow`](https://invent.kde.org/network/neochat/-/blob/29321aeaf3361becdf5bf65aa9b01a721f4307c1/src/windowcontroller.cpp#L57)

A quick port of the code works on X11 and Wayland, but requires KF6 and Qt6 >= 6.5.0. Not sure if you want that much ifdefs :)

ctrlaltca:
KDE 6 alpha required Qt 6.5.0, and I saw some components of beta2 explicitly requiring Qt 6.6.1.
I think we can assume that an user compiling against KF6 will have a Qt version >= 6.6.
That's probably the way we'll need to take anyway, so if you want to propose a PR with your port of the code i'll be glad to give it a look, thank you!

DarthGandalf:
Until we drop support of Qt 5, I'd be fine to require new enough Qt 6 for users who want 6 instead of 5

PR #2586, opened 2023-12-26
Activation support for KDE (KF6) notifications

alebastr:
Use xdg-activation tokens when available. This should work with KF6 and `xcb` and `wayland` platform backends on Plasma 6 (Wayland).

Results may vary with other Wayland compositors: there's a really weird interaction with `FlashQueryWindowOnNewMessages` on Sway that should not be permitted by the compositor. GNOME should be fine though.

It's also not worth extracting the lambda as a named function: the whole xdgActivationToken dance is specific to KNotification and not reusable.

### DIFF
--- a/src/kvirc/kernel/KviApplication.cpp
+++ b/src/kvirc/kernel/KviApplication.cpp
@@ -126,6 +126,10 @@
 #ifdef COMPILE_KDE_SUPPORT
 #include <QStandardPaths>
 #include <KNotification>
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+#include <KStartupInfo>
+#include <KWindowSystem>
+#endif
 #endif
 
 #ifdef COMPILE_QX11INFO_SUPPORT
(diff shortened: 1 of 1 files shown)


ctrlaltca:
Confirmed working just fine, thank you

DarthGandalf:
Should this also bump the required Qt6 version in CMakeLists.txt?

alebastr:
> Should this also bump the required Qt6 version in CMakeLists.txt?

As @ctrlaltca noted in #2585, KDE Frameworks already does Qt version check for us. It won't be possible to have Qt6, `COMPILE_KDE_SUPPORT` and Qt version below 6.5.
There's also no direct dependency on any of the new Qt APIs.

DarthGandalf:
> It won't be possible to have Qt6, COMPILE_KDE_SUPPORT and Qt version below 6.5.

That's fine. What I mean is a good error message during configure time, instead of a random build-time error.

But it looks like I confused Qt6 version and KF6 version

PR #2587, opened 2023-12-28
update dicts, a clang-format config change and perl updates

Dessa:
only real visible change is likely the dicts, lets wait for CI and see if this still works.


ctrlaltca:
Looks good, I tested the resulting windows artifact

PR #2589, opened 2024-01-08
Fix nickserv identification (fix wildcard regexp on strings containing a slash)

ctrlaltca:
Reported on irc by @Arisha 
KviRegExp has been ported to QRegularExpression, including the "wildcard mode" handled with QRegularExpression::wildcardToRegularExpression().
This method normally assumes that the pattern is a path, so it doesn't consider the slash "/" character as valid.
Eg. the `*IDENTIFY*` pattern used by default by nickserv identification gets converted to the regexp `[^/]*IDENTIFY[^/]*`
Qt 6.6 introduced a specific flag to disable this behavior.


ctrlaltca:
Confirmed working on both Qt5 and Qt6

PR #2592, opened 2024-01-18
Fix build with Qt5 and WANT_DCC_VIDEO

alebastr:
```
src/modules/dcc/DccVideoWindow.cpp: In member function ‘virtual const QString& DccVideoWindow::target()’:
src/modules/dcc/DccVideoWindow.cpp:660:40: error: cannot convert ‘QString’ to ‘QString*’ in assignment
```
Tested with `/eval echo $target` in the DCC Video window, looks correct.

### DIFF
--- a/src/modules/dcc/DccVideoWindow.cpp
+++ b/src/modules/dcc/DccVideoWindow.cpp
@@ -396,7 +396,6 @@ DccVideoWindow::DccVideoWindow(DccDescriptor * dcc, const char * name)
 {
 	m_pDescriptor = dcc;
 	m_pSlaveThread = nullptr;
-	m_pszTarget = nullptr;
 
 	m_pButtonBox = new KviTalHBox(this);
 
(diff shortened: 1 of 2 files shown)


src/modules/dcc/DccVideoWindow.cpp
```diff
@@ -654,11 +647,8 @@ void DccVideoWindow::connectionInProgress()
 const QString & DccVideoWindow::target()
 {
 	// This may change on the fly...
-	if(!m_pszTarget)
-		m_pszTarget = new QString();
-
-	m_pszTarget = QString::asprintf("%s@%s:%s", m_pDescriptor->szNick.toUtf8().data(), m_pDescriptor->szIp.toUtf8().data(), m_pDescriptor->szPort.toUtf8().data());
-	return *m_pszTarget;
+	m_szTarget = QString::asprintf("%s@%s:%s", m_pDescriptor->szNick.toUtf8().data(), m_pDescriptor->szIp.toUtf8().data(), m_pDescriptor->szPort.toUtf8().data());
```
alebastr, line 650: Pretty sure this should be
```c++
QLatin1String("%1@%2:%3").arg(m_pDescriptor->szNick, m_pDescriptor->szIp, m_pDescriptor->szPort);
```
to avoid extra charset conversions and for type safety.
But then I'll have to change all ~30 occurrences of `asprintf` for consistency, and that's beyond the scope of a simple build fix.

ctrlaltca:
Merged, thank you!

PR #2599, opened 2024-01-27
Add support for sanitizers at compile time (and fix some issues)

ctrlaltca:
This PR adds the WANT_SANITIZER cmake flag, only available on debug builds.
The option accepts 4 possible values: off (default), address, memory and thread.
Building with this flag will make the executable slower (but still usable).
The resulting executable will log memory related problems in the terminal.

It already found a few issues, for which the fix are included here as separate commits:
 * Some singleton classes were dereferencing nullptrs on init. Simplify them and avoid the bug.
 * A possible fix for a previous perl related bug made symbols exported by modules global. This may cause a crash if multiple modules exported the same symbols (hint: they did). Revert this change.
 * KviTalIconAndRichTextItemDelegate assumed its parent was a QListWidget and happily casted to it. Then it started getting used inside QTreeWidgets, where the casts were wrong. I guess this couldn't cause a crash since we only used methods from their common ancestor QAbstractItemView, but better fix it.

 There are still a couple of things i'd like to investigate before merging this:
 * KviAction::actionDestroyed() wrongly assumes the sender is a QAction, but it's a QObject instead
 * ~Kvs scripting seems to leak memory around here and there, definitely needs some further investigation~ Fixed
 * ~KviSignalHandler leaks the whole object~ Fixed
 * ~KviLocale's ctor instantiates a KviMessageCatalogue; KviMessageCatalogue's ctor requires the KviLocale singleton to already exist~ Fixed by requiring UTF-8 for .po files.
 * ~KviInputEditor leaks every typed text from undo/redo stacks~ Fixed

Finally, a debug message warning when a single message is longer than 512 bytes has been removed.
As per https://ircv3.net/specs/extensions/message-tags.html#size-limit , message tags can prepend a whooping 8kbs of "tags" (header data) to each message.


CMakeLists.txt
```diff
@@ -233,6 +234,11 @@ else()
 				set(CMAKE_C_FLAGS "-O0 -g")
 			endif()
 		endif()
+		if(WANT_SANITIZER)
```
DarthGandalf: This is specifically address sanitizer. There are other sanitizers - TSan, UBSan, MSan, etc.

I don't mind being able to enable ASan using a special option, though I usually set CXXFLAGS directly in such cases. But I suggest at least making it clear somehow which sanitizer is used: either `WANT_ADDRESS_SANITIZER=1` or `WANT_SANITIZER=address`

src/kvirc/module/KviModuleManager.cpp
```diff
@@ -175,8 +175,6 @@ bool KviModuleManager::loadModule(const QString & modName)
 	}
 
 	QLibrary * pLibrary = new QLibrary(tmp);
-	pLibrary->setLoadHints(QLibrary::ExportExternalSymbolsHint);
```
DarthGandalf, line 178: Won't this break perl again?

DarthGandalf (changes requested):

CMakeLists.txt
ctrlaltca: Nice idea for improvement, will implement

src/kvirc/module/KviModuleManager.cpp
ctrlaltca: Perl actually worked for everybody but some gentoo boxes. I've tested reverting the change and it works for me on both linux and windows,but i didn't test gentoo specifically.
To be honest i'm not even sure the fix worked, since there's no feedback about it.

DarthGandalf: I use gentoo, and I remember kvirc crashing a lot for me while I tried to use perl scripting there :) I don't remember whether it was before or after that change.

CMakeLists.txt
ctrlaltca: Done in https://github.com/kvirc/KVIrc/pull/2599/commits/962c93a66e15bc8748e77a4b53d6f06cb4c725e8
Btw gcc unfortunately doesn't support the memory sanitizer, only clang does.

src/kvirc/module/KviModuleManager.cpp
ctrlaltca: The original issue was about the perlcore module not loading at all, so as long as an hello world kvs/perl script works, it should be ok.

CMakeLists.txt
```diff
@@ -234,10 +235,21 @@ else()
 				set(CMAKE_C_FLAGS "-O0 -g")
 			endif()
 		endif()
-		if(WANT_SANITIZER)
+		string(TOUPPER "${WANT_SANITIZER}" WANT_SANITIZER )
+		if(WANT_SANITIZER STREQUAL ADDRESS)
 			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer")
-			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer")
-			set(CMAKE_STATUS_DEBUG_SUPPORT "Yes, with sanitizers")
+			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined -fsanitize=memory -fsanitize=address -fno-omit-frame-pointer")
```
DarthGandalf: undefined, memory, and address at the same time?

ctrlaltca: Undefined and address can be used together. Memory is a copy/paste error, it can only be used alone.
I didn't notice because KVIrc only uses the c++ compiler, so only CXX_FLAGS was checked.
Fixed in https://github.com/kvirc/KVIrc/pull/2599/commits/38b91c6c9dfeb94a31e23c54b92011a3b2bba14c, thank you

CMakeLists.txt
```diff
@@ -233,6 +235,22 @@ else()
...
+			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer")
+			set(CMAKE_STATUS_DEBUG_SUPPORT "Yes, with address,leak and undefined behavior sanitizers")
+		elseif(WANT_SANITIZER STREQUAL MEMORY)
+			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=memory -fPIE -pie -fno-omit-frame-pointer")
+			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=memory -fPIE -pie -fno-omit-frame-pointer")
+			set(CMAKE_STATUS_DEBUG_SUPPORT "Yes, with memory sanitizer")
+		elseif(WANT_SANITIZER STREQUAL THREAD)
+			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=thread -fno-omit-frame-pointer")
+			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=thread -fno-omit-frame-pointer")
+			set(CMAKE_STATUS_DEBUG_SUPPORT "Yes, with memory sanitizer")
```
Dessa: that should probably be thread

ctrlaltca: Fixed in https://github.com/kvirc/KVIrc/pull/2599/commits/42009c2d1eefaebf63b5d860b3cc3366537ee838, thank you

DarthGandalf (approved):

PR #2606, opened 2024-02-12
Switch windows CI builds to Github Actions and Qt6

ctrlaltca:
Required a lot of small code fixes.


.github/workflows/build_windows.yml
```diff
@@ -0,0 +1,185 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup devcmd
+        uses: ilammy/msvc-dev-cmd@v1
+
+      # https://learn.microsoft.com/en-us/vcpkg/users/binarycaching#gha
+      - name: Set variables for vcpkg
```
DarthGandalf, line 28: What's the point of this? This step takes values from environment, and puts them to... environment?

cmake/KVIrc.nsi.cmake
```diff
@@ -19,9 +19,11 @@ Name "@NICENAME@"
 !define PUBLISHER 'Szymon Stefanek and The KVIrc Development Team'
 
 ; Git release, e.g. KVIrc-4.3.2-dev-git-5465-ge352609.exe
-OutFile KVIrc-${KVIRC_BUILD_REVISION}.exe
+;OutFile KVIrc-${KVIRC_BUILD_REVISION}.exe
 ; Stable version, e.g. KVIrc-4.2.0-Equilibrium.exe
 ;OutFile KVIrc-${VERSION}-${RELEASE_NAME}.exe
+; Github CI
+OutFile KVIrc.exe
```
DarthGandalf, line 26: What was wrong with ${KVIRC_BUILD_REVISION}?

Does this file still need to be passed through cmake then?

.github/workflows/build_windows.yml
ctrlaltca: It sounds obscure to me, too.
I found this way to enable caching for vcpkg at the link https://learn.microsoft.com/en-us/vcpkg/users/binarycaching#gha , and it works.

cmake/KVIrc.nsi.cmake
ctrlaltca: I did this for 2 reasons:
 1. the CI for macos already created a name for the resulting setup file based on the git commit/tag, and i wanted to use the same name also for the windows CI
 2. when i need to upload the artifact from inside the CI i need to know its full name, and i wanted to avoid having to add a workaround like executing "dir *.exe" and filtering its output.
So, the easiest way i foudn was to make NSIS output a fixed name and the rename the resulting setup file from inside the CI action.

PR #2610, opened 2024-02-24
Fix crash on multiline editor; Fix 2609

ctrlaltca:
The m_pStartTimer pointer was not initialized to nullptr.
Added some guards to the pointer usage.
While we're at it, make the contextual help work again

### DIFF
--- a/src/modules/editor/ScriptEditorImplementation.cpp
+++ b/src/modules/editor/ScriptEditorImplementation.cpp
@@ -94,6 +94,7 @@ static bool bCompleterReady = false;
 ScriptEditorWidget::ScriptEditorWidget(QWidget * pParent)
     : QTextEdit(pParent)
 {
+	m_pStartTimer = nullptr;
 	m_pSyntaxHighlighter = nullptr;
 	setTabStopDistance(48);
 	setAcceptRichText(false);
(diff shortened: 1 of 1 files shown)


src/modules/editor/ScriptEditorImplementation.cpp
```diff
@@ -457,8 +457,9 @@ bool ScriptEditorWidget::contextSensitiveHelp() const
 	QTextCursor cur = cursorForPosition(QPoint(r.x(), r.y()));
 	cur.select(QTextCursor::WordUnderCursor);
 	QString szText = cur.selectedText();
-	QString szTmp = szText;
-
+	KviQString::escapeKvs(&szText);
+	QString szParse = QString("timer -s (help,0){ help.open %1; }").arg(szText);
```
DarthGandalf, line 461: Why does it need to go through timer?

ctrlaltca: That's how it worked in the past:
https://github.com/kvirc/KVIrc/commit/5f0292c32b336be0bb279a7511a9ee37ba9df3fa#diff-ee6389e2178d7b15aa4a4ccc1699f64becf9a3e25feedcfde4e4ea6d0f2905eaL539

I guess the idea was to make it async

PR #2612, opened 2024-03-03
Switching the mac CI to qt6 

ctrlaltca:
Remove the cmake hardcoded deploy code, use macdeployqt instead.

Also update qt version on windows and linux builds


ctrlaltca:
This is finally working, merging

PR #2613, opened 2024-03-08
Numerical 301: respect the value of KviOption_boolWhoisRepliesToActiveWindow

ctrlaltca:
Numerical 301 is the handler of "user is away" messages.
Other whois reply handlers respect the value of this variable even if a query is open.


ctrlaltca:
@Mattoje this should fix the problem you reported on IRC

Mattoje:
yep it's fixed indeed

PR #2622, opened 2024-03-21
More qt6 porting

ctrlaltca:
This PR fixes all the remaining warnings in code when compiled with Qt6.
The code is still Qt5 and Qt6 compatible.
Most of the changes are boring, with the single exception of 6f5ad1f5d33aabfeff4e7cb266480e268ae9c84a that actually changes the way font weights are saved in the configuration files. Old values are migrated automatically.

### DIFF
--- a/src/modules/dialog/libkvidialog.h
+++ b/src/modules/dialog/libkvidialog.h
@@ -31,6 +31,7 @@
 
 #include <QDialog>
 #include <QMessageBox>
+#include <QPushButton>
 #include <QString>
 
 class KviKvsCallbackMessageBox : public QMessageBox, public KviKvsCallbackObject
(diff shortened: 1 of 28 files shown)


DarthGandalf (approved):

src/modules/dialog/libkvidialog.h
```diff
@@ -48,6 +49,11 @@ class KviKvsCallbackMessageBox : public QMessageBox, public KviKvsCallbackObject
 	    KviKvsVariantList * pMagicParams,
 	    KviWindow * pWindow, bool modal = false);
 	~KviKvsCallbackMessageBox();
+
+protected:
+	QPushButton * yesButton;
```
DarthGandalf: shouldn't this be m_ ?

ctrlaltca: Fixed in a54359db22a8a65d8018fd76bef5ff5b19ba63f6

Issue #1895, opened 2016-03-03
Initial whitespace in first line of input unnecessarily stripped


Arfrever:
Initial whitespace in first line of input is unnecessarily stripped.

If I write

```
   a
```

in single-line input mode and press Enter,

```
a
```

will be sent.

If I write

```
   a
   b
```

in multi-line input mode and press Ctrl+Enter,

```
a
   b
```

will be sent.

un1versal:
@Arfrever for the record can you post this info as well please https://github.com/kvirc/KVIrc/wiki/Submitting-a-bug-report#information-required-when-making-a-bug-report-via-github-issues

Arfrever:
```
Executable Information

KVIrc 4.9.1 'Aria'

Runtime Info:
System name: Linux
Architecture: x86_64
Qt version: 5.5.1
Qt theme: fusion

Build Info:
Build date: 2016-02-25 03:59:53 UTC
Sources date: 20160102
Revision number: fb3d7d44b0a90afea71c1775d4511e3807ab2c63
System name: Linux
CPU name: x86_64
Build command: /usr/bin/cmake
Build flags: 
   MANDIR=share/man
   CMAKE_INSTALL_PREFIX=/usr
   VERBOSE
   LIB_SUFFIX=64
   Threads=POSIX
Compiler name: /usr/bin/x86_64-pc-linux-gnu-g++
Compiler flags: -march=native -O2 -fno-ident -frecord-gcc-switches -pipe -Wall -fno-strict-aliasing 
Qt version: 5.5.1
```

PR #2627, opened 2024-03-29
Strip leading spaces on command; Fix #2626

### LINKED ISSUE #2626: [Feature Request] Strip leading spaces on command
My niche problem: I copy a command from some DCC channel but I included the prepending space because I did not properly 'target' the slash before because it's hard for me to properly see it: `  /MSG user text`.   
Result: "Unknown command is an unknown server command"

Can it be changed that when kvirc encounters this error, instead of showing the error and aborting it will strip leading spaces and issue the command again - and only if that fails show the error?

Thank you

ctrlaltca:
KVIrc used to strip any space at the beginning of a message/command.
This behavior was reported as a bug in #1895 and changed in eed2e4dd7779d38e336834c7054e5be67b97094e.
While it's good to keep any space at the beginning of a text message, it's good to strip them if the message contains a command.

### DIFF
--- a/src/kvirc/kernel/KviUserInput.cpp
+++ b/src/kvirc/kernel/KviUserInput.cpp
@@ -45,6 +45,8 @@ namespace KviUserInput
 		if(!c->unicode())
 			return true; // empty
 
+		// check if it's a KVS command. Skip any initial space
+		while(c->isSpace()) c++;
 		if(c->unicode() == '\\')
 		{
 			c++;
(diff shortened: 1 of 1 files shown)


DarthGandalf:
To clarify: this affects both user-friendly and KVS modes? Should it?

Code-wise, this seems fine.

ctrlaltca:
> To clarify: this affects both user-friendly and KVS modes? Should it?

Yes, it affects both user-friendly and KVS modes.
I guess it should, seems like a regression introduced in eed2e4dd7779d38e336834c7054e5be67b97094e to fix another usecase.

PR #2628, opened 2024-04-01
Port dcc video to qt6

ctrlaltca:
Dcc video has never been in an usable state, and it still isn't.
This PR adapts the code to the Qt6 API, dropping Qt5 support.
It should now be working on every OS supported by Qt.
There are still a lot of things that would need to be done, eg:
 * adaptive frame rate and resolution
 * real video compression
 * muxing audio, video and text in a single tcp stream
 * creating a standard / actually making it compatible with other clients

To be honest i don't use any DCC anymore, let alone dcc video. In a world of NAT-ted devices, better technologies exists to accomplish this. No need to reinvent the wheel this time.
Still, we either drop dcc video completely or keep it working enough to justify its existence.

### DIFF
--- a/src/modules/dcc/DccVideoWindow.cpp
+++ b/src/modules/dcc/DccVideoWindow.cpp
@@ -46,6 +46,10 @@
 #include <QToolTip>
 #include <QByteArray>
 #include <QBuffer>
+#include <QMessageBox>
+#include <QVideoFrame>
+#include <QVideoSink>
+#include <QVideoWidget>
 
 #ifdef COMPILE_CRYPT_SUPPORT
 #include "KviCryptController.h"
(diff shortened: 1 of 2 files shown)


ctrlaltca:
@DarthGandalf i don't expect a real review on this PR, but id' be curious to know your opinion on dropping dcc video vs keeping it minimally upgraded but disabled by default.

DarthGandalf:
I stopped using DCC many years ago. There are better FOSS videoconferencing solutions. If you want to drop it, I won't stop you.

VVD:
> This PR adapts the code to the Qt6 API, dropping Qt5 support.

IMHO, check for WANT_DCC_VIDEO must be with QT6 only:
```
--- CMakeLists.txt.orig 2024-05-11 11:44:18 UTC
+++ CMakeLists.txt
@@ -938,6 +938,7 @@ endif()
 # DCC video support + extended webcam support (LibV4L2)
 ############################################################################

+if(${QT_VERSION_MAJOR} EQUAL 6)
 option(WANT_DCC_VIDEO "Compile DCC video support" OFF)

 if(WANT_DCC_VIDEO)
@@ -955,6 +956,11 @@ if(WANT_DCC_VIDEO)
 else()
        set(COMPILE_DISABLE_DCC_VIDEO 1)
        set(CMAKE_STATUS_DCC_VIDEO_SUPPORT "No")
+endif()
+
+else()
+       set(COMPILE_DISABLE_DCC_VIDEO 1)
+       set(CMAKE_STATUS_DCC_VIDEO_SUPPORT "Not available (Qt 6 required)")
 endif()

 ############################################################################
```

PR #2630, opened 2024-04-08
Make D-Bus adaptor actually useful and use it as an IPC backend

ctrlaltca:
KviDbusAdaptor exists since long time, but it's currently unused/quite useless.
This PR adds the necessary changes to make it working/possibly useful.

IPC is currently broken on Linux/Wayland, as the existing backend is using XAtom.
This PR also adds a D-Bus based backend for IPC, so that kvirc can detect on startup if another instance is already running and eventually pass messages to it (using kvirc -e / -x / -r).

As a bonus, you can now go full [inception](https://wikipedia.org/wiki/Inception) mode by running KVS such as:
```
/$system.dbus("net.kvirc.KVIrc", "/kvirc", "net.kvirc.KVIrc","ipcMessage", "session", "QString=echo ciao")
```

### DIFF
--- a/src/kvirc/kernel/KviDbusAdaptor.cpp
+++ b/src/kvirc/kernel/KviDbusAdaptor.cpp
@@ -28,7 +28,7 @@
 
 #ifdef COMPILE_DBUS_SUPPORT
 
-#include <QtGlobal> //for qWarning()
+#include <QtDebug> //for qWarning()
 #include <QDBusConnection>
 
 KviDbusAdaptor::KviDbusAdaptor(QObject * pObj)
--- a/src/kvirc/kernel/KviDbusAdaptor.cpp
+++ b/src/kvirc/kernel/KviDbusAdaptor.cpp
@@ -28,6 +28,8 @@
 
 #ifdef COMPILE_DBUS_SUPPORT
 
+#include <QDBusConnection>
+
 KviDbusAdaptor::KviDbusAdaptor(QObject * pObj)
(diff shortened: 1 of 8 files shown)


src/kvirc/kernel/KviDbusAdaptor.cpp
```diff
@@ -24,11 +24,37 @@
...
+    }
+}
+
+#ifndef COMPILE_NO_IPC
+void KviDbusAdaptor::ipcMessage(const QString &message)
+{
+    if(!g_pApp)
+        return;
+
+    g_pApp->ipcMessage(message.toUtf8().data());
```
DarthGandalf, line 56: I'd prefer if the dbus interface was more structured than a random binary blob which might or might not be a KVS string

DarthGandalf: And it doesn't have any way to return values

ctrlaltca: This just had to play nice with the other ipcMessage implementations, that works exactly by throwing the random binary blob.
Exporting a more structured/descriptive interface via DBus would be indeed nicer.
Maybe something including all open networks/channels/queries, with at least some working getInfo / sendMessage / signalMsgReceived

ctrlaltca: I've created #2632 to keep track of it

PR #2633, opened 2024-04-13
Add an option to disable vertical margin between lines; fix #2464

### LINKED ISSUE #2464: [Request] Adjustable Line Spacing in Output
It would be excellent if we could adjust the amount of whitespace between lines in the output window. This would be very useful for ascii art communities.

Here's an example:

http://lowtrucks.net/Forums/Pictures/mIRC-IRC-ANSIart1.png

You'll see that full-height characters touch at the full-height characters above and below it since there is not a defined number of pixels separating lines.

ctrlaltca:
This PR adds an option to disable vertical margin between lines.

Inside the ircview the m_iFontDescent variable, used primarily for painting, was also used as an additional margin between lines. Split that usage into a new m_iFontLineVMargin.
An unused member variable "m_iFontLineWidth" has been removed from the ircview.


DarthGandalf:
Sorry, I somehow missed this. I don't have much comments, I just hope you tested it :)

PR #2636, opened 2024-04-30
Spellchecker: migrate to enchant++ 

ctrlaltca:
To avoid use of private API in enchant-provider.h, no more available since enchant 2.7


ctrlaltca:
Merging this, since it's breaking CI.

PR #2638, opened 2024-05-04
Set portable when it is needed on windows

### LINKED ISSUE #2637: Windows: Portable file check inside ctor not working as intended
Current usage of isWritable on Windows will result in directory being always read only (if user installs portable installation on NTFS volume)

Here info on how to enable NTFS permission checking actually https://doc.qt.io/qt-6/qfileinfo.html#ntfs-permissions
But I would suggest just to disable it all together on windows and let it fail if user actually marks folder as read only

Code in question:
https://github.com/kvirc/KVIrc/blob/master/src/kvirc/kernel/KviApplication_setup.cpp#L97C35-L97C45

DoumanAsh:
This moves `portable` file check outside of application ctor to avoid `QApplication` weird behavior

Previously on my system (Windows 10) with version 5.2.2, KVirc was unable to detect `portable` file correctly
With this change, it is now able to correctly determine it

The most likely reason is that you can only call `applicationDirPath` outside of ctor

Fixes #2637

### DIFF
--- a/src/kvirc/kernel/KviApplication.cpp
+++ b/src/kvirc/kernel/KviApplication.cpp
@@ -227,9 +227,6 @@ KviApplication::KviApplication(int & argc, char ** argv)
 	kvi_socket_flushTrafficCounters();
 	// don't let qt quit the application by itself
 	setQuitOnLastWindowClosed(false);
-#if defined(COMPILE_ON_WINDOWS) || defined(COMPILE_ON_MINGW)
-	m_bPortable = KviFileUtils::fileExists(QString("%1%2%3").arg(g_pApp->applicationDirPath().arg(KVI_PATH_SEPARATOR_CHAR).arg("portable")));
-#endif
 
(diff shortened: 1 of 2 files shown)


DoumanAsh:
I'm not sure what is wrong with mac os build:

```
[100%] Building CXX object src/modules/spellchecker/CMakeFiles/kvispellchecker.dir/libkvispellchecker.cpp.o
1717
/Users/runner/work/KVIrc/KVIrc/src/modules/spellchecker/libkvispellchecker.cpp:29:10: fatal error: 'enchant-provider.h' file not found
1718
#include <enchant-provider.h>
1719
         ^~~~~~~~~~~~~~~~~~~~
```

ctrlaltca:
Don't worry, it's a known problem, the fix is in https://github.com/kvirc/KVIrc/pull/2636 but i'm waiting for an ACK before merging it

ctrlaltca:
Merged, thank you!

Issue #2647, opened 2024-06-09
Can't Launch Binary on Windows 7 with v5.2.2 onward


zellchristensen:
While the installer works fine, I am unable to launch kvirc.exe in version 5.2.2 or 5.2.4 under Windows 7 (x64). I just get an error "The application was unable to start correctly (0xc0000005). Click OK to close the application."

![2024-06-08_21-19-43](https://github.com/kvirc/KVIrc/assets/38737447/f2043a19-d372-4095-bb2d-b41659328785)

It seems to launch fine in version 5.2.0 and under. 

I'm assuming this is most likely related to a dependency that dropped support for Windows 7.

If Windows 7 support is discontinued, I wasn't able to find any documentation on the website or github to that effect.

ctrlaltca:
I just tested on Window 7 and i can confirm the application is not starting, as it's requiring some components (eg. d3d12.dll) not available on Windows 7.
We are using Github CI to build KVIrc packages, and unfortunately it looks like they don't support Windows 7 anymore out of the box.
I tried switching to an old image in https://github.com/kvirc/KVIrc/pull/2648 , if the resulting build works on Windows 7 i'll publish it as official.

> If Windows 7 support is discontinued, I wasn't able to find any documentation on the website or github to that effect.

Yep, that was not intentional, sorry.

ctrlaltca:
So, a windows 7 compatible build has been added to https://github.com/kvirc/KVIrc/releases/tag/5.2.4
Unfortunately it's based on Qt5 instead of Qt6, since Qt6 doesn't support Windows 7.
We'll try to continue target windows 7 as long as it's possible.

PR #2648, opened 2024-06-09
Test windows2019 CI image to see if the resulting binary is still compatible with windows 7

ctrlaltca:
See #2647


ctrlaltca:
It looks like Qt6 is not supporting Windows 7 anymore, so this is a no-go.

PR #2651, opened 2024-06-12
update azzurra.org -> azzurra.chat

sorcio:
Updating the address of Azzurra IRC Network, which is now `irc.azzurra.chat`. I also updated an example which used the old address.

Context: Azzurra IRC Network is a long-running (since 1997) major Italian network. It had used the `azzurra.org` domain name since 2001, but it has been recently forced to renounce it due to the domain owner being uninvolved with network operations. On June 8th, the entire Azzurra network [transitioned to azzurra.chat](https://www.azzurra.chat/?mod=news) as the official domain name. The only link between the former name and the new one is a CNAME on `irc.azzurra.org`, which is only a temporary measure granted by the domain owner, and might not last long.

### DIFF
--- a/data/config/serverdb.kvc
+++ b/data/config/serverdb.kvc
@@ -47,7 +47,7 @@ NServers=2
 1_SSL=true
 NServers=2
 [Azzurra]
-0_Hostname=irc.azzurra.org
+0_Hostname=irc.azzurra.chat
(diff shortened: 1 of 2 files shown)


ctrlaltca:
Hello @sorcio , thanks for the heads-up!

PR #2661, opened 2024-07-11
Force app_id to be set in more places

ReillyBrogan:
While testing the new version of kvirc I noticed that the setup window was using the `net.kvirc.kvirc` app_id. It seems my previous fix was incomplete since that window is started before the other window where we set the app_id.

Fix it by adding it to the setup window as well, and also just add it to a bunch more locations so that the chances of it not being set elsewhere are lower.


ctrlaltca:
Merged, thank you!

Issue #2117, opened 2016-08-20
GTKSTYLE: Not as broke as advertised - Conditions may apply - Update INSTALL doc.


un1versal:
> Note: This issue is a discussion now linked to from INSTALL doc general  future reference and to have a place to discuss it and track issues.
## Intro

When looking at [`-DWANT_GTKSTYLE=` text in our INSTALL doc](https://github.com/kvirc/KVIrc/blob/master/doc/INSTALL.txt#L446-L461) thers all this doom and gloom blaming it on GTK insanity.

[Following this comment by @minj](https://github.com/kvirc/KVIrc/issues/2113#issuecomment-241187426) I just had to test it at this time.
### INSTALL DOC claims
- Menus do not appear
- Scrollbars are invisible
- Cant show icons on menus
- Label background is fixed see https://github.com/kvirc/KVIrc/issues/2117#issuecomment-241208158
- interferes with KVIrc internal theming (unreadable text white text on white background)

Also see https://github.com/kvirc/KVIrc/issues/2117#issuecomment-241208158 for clarification on the original documented issues.
### Observations after testing
- Menus _appear_ and render just fine (2 exceptions)
- Scrollbars _are visible_
- Icons _can and are shown in menus_
- Label background (seems fine ~~idk what fixed or INSTALL doc meant~~ with the style tested.)
- Theres no unreadable text  on my tests (exception for interference with theme is selection bars and some tooltips hardcoded colors.)
- Some dark Icons arent quite visible (exception but needs improving)
#### Exceptions of which some should be fixed.
- [ ] 1.  Server connect button is unstyled
- [ ] 2.  Groupbox boundaries on settings menus aren't visible,
- [ ] 3.  Menu separators and other similar type separators aren't visible.
- [ ] 4.  Overrides KVIrc theme selected color (exception which seems fine anyway)
- [ ] 5. Some Tooltips look weird and some text hard to read **(because we customize them)** this is expected then but looks bad, some unreadable text depending on style.
- [ ] 6. DCC transfer progress indicator is unstyled. (probably customized also).
- [ ] 7. Some dark program Icons arent quite visible in a dark setting/

Things could be better but hardly critical or not a reason to disable GTKSTYLE  **_however_**

I _**suspect**_ this is heavily dependent on desktop used, perhaps even distro or version and themes used how well GTK style is implemented there.
### Test Environment  / Install

[...]

PR #2663, opened 2024-07-14
Add an option to let the user choose the qt style used by the app

### LINKED ISSUE #2662: Does KVIrc have a dark mode on Windows?
Hi everyone.
Is it possible on Windows to have the KVIrc window and all its elements displayed in dark colors (not only the channel/chat window contents)?
Thanks in advance!

ctrlaltca:
Fix #2662

### DIFF
--- a/src/modules/options/OptionsWidget_theme.cpp
+++ b/src/modules/options/OptionsWidget_theme.cpp
@@ -31,7 +31,9 @@
 #include "KviTalToolTip.h"
 
 #include <QLayout>
+#include <QStyleFactory>
 
+#define DEFAULT_QT_STYLE "Default"
 OptionsWidget_theme::OptionsWidget_theme(QWidget * parent)
     : KviOptionsWidget(parent)
 {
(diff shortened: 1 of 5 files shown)


DarthGandalf:
So the -style command line flag will be ignored?

Any way to only change the default, but still honor the explicit user's config?

ctrlaltca:
> So the -style command line flag will be ignored?

Unfortunately yes.

> Any way to only change the default, but still honor the explicit user's config?

Should we add an option? This way we could also remove the hardcoded style for gtk3 (#2117).
Hopefully we can get the style from options before the mainwindow get shown to avoid a sudden style change

AbraxasDMG:
Is the command line switch the only way to apply different style at the moment (no possibility to do it through some config file or environment var)?

If so, KVIrc should also be able to recognize -style cmdline argument and avoid treating it as a server name to which it should connect automatically.

Just realized that this can be prevented if the style argument is provided as follows.
`-style=Fusion`

Then a new standalone server is not added to the list and automatically attempted to connect to.

ctrlaltca:
@DarthGandalf please have a look at the second iteration of this patch, I've changed approach adding a new option in theme->general

@AbraxasDMG once we get this working you won't need to provide the commandline parameter anymore

src/modules/options/OptionsWidget_theme.cpp
```diff
@@ -44,12 +46,38 @@ OptionsWidget_theme::OptionsWidget_theme(QWidget * parent)
 	    KVI_OPTION_BOOL(KviOption_boolUseGlobalApplicationFont));
 	connect(b, SIGNAL(toggled(bool)), f, SLOT(setEnabled(bool)));
 
-	addRowSpacer(0, 3, 1, 3);
+	addLabel(0, 3, 0, 3, __tr2qs_ctx("Qt Style:", "options"));
+	m_pQtStyle = new QComboBox(this);
+	addWidgetToLayout(m_pQtStyle, 1, 3, 1, 3);
+	m_pQtStyle->addItem(DEFAULT_QT_STYLE);
+	foreach(QString key, QStyleFactory::keys()) {
```
DarthGandalf: From Qt docs: New code should prefer C++11 range-based loops.
```suggestion
	for (const QString& key : QStyleFactory::keys()) {
```

DarthGandalf:
Haven't tested, but seems fine, I guess

PR #2664, opened 2024-07-15
Remove GTK Style flag; fix #2117. Default to Fusion theme on Windows

### LINKED ISSUE #2117: GTKSTYLE: Not as broke as advertised - Conditions may apply - Update INSTALL doc.

ctrlaltca:
This is a followup of #2663


src/kvilib/config/KviBuildInfo.cpp
```diff
@@ -88,11 +88,11 @@ const QStringList feature_list{
 #endif
 #ifdef COMPILE_ENCHANT_SUPPORT
 	"Enchant",
-#endif
-#ifdef COMPILE_ENABLE_GTKSTYLE
-	"GTK",
 #endif
 	"Qt5",
```
DarthGandalf, line 92: maybe put Qt5 to #else branch?

ctrlaltca: I kept Qt5 since some scripts may need it, eg. `/echo $features(Qt5)` to determine if  some features only available since Qt5 are present

DarthGandalf: Okay, but a comment which explains that would be useful

DarthGandalf: Done in 03a8df16588ab42b4f4a070fc665e7ccf512e89d

PR #2669, opened 2024-07-24
Misc fixes to sound module related to autodetect

### LINKED ISSUE #2667: Settings are changed even when clicking Cancel
5.2.4, linux

1. phonon is selected as sound system
2. open settings -> general -> sound -> sound system
3. click Autodetect
4. see that the dropdown was changed from phonon to qt
5. click Cancel to close the window (or X button)
6. open settings again
7. it's qt

### LINKED ISSUE #2668: Automatically autodetect sound system if old one doesn't exist
5.2.4, linux

1. compile with phonon support
2. use phonon happily
3. compile without phonon support

Actual result:

"null" sound system is selected, and there's no sound

Expected result:

the sound system should be autodetected in this case, which would select "qt", since "phonon" is not available anymore

https://bugs.gentoo.org/934767

ctrlaltca:
1) Sound options: don't immediately save changes when sound system gets autodetected.
Instead of changing the value of `KVI_OPTION_STRING(KviOption_stringSoundSystem)` inside `KviSoundPlayer::detectSoundSystem`, return the detected value and let the caller handle it
fix #2667

2) If the config-saved sound system is not available anymore, force an autodetect
fix #2668

3) Fixed the "qt" sound system not working under qt6: the QSoundEffect object was method-local and deleted before the sound actually played.

4) ~Prefer the "phonon" sound system if available over the "qt" sound system, since the latter can only play wav files.~ changed in point 6

5) Nuke support for Arts (KDE2/3 sound server) and ESD (Gnome sound server before 2009)

6) Replace QSoundEffect with QMediaPlayer to support more file formats. Adjust the priority of "qt" sound system in autodetect


DarthGandalf:
https://doc.qt.io/qt-6/videooverview.html#supported-media-formats says it's not only wav supported

ctrlaltca:
That doc is valid for QMediaPlayer. The doc for QSoundEffect is: https://doc.qt.io/qt-6/audiooverview.html#low-latency-sound-effects
It it doesn't add further dependencies, we  could switch to QMediaPlayer anyway

DarthGandalf:
>  It it doesn't add further dependencies, we could switch to QMediaPlayer anyway

Yeah, that would be closer to phonon experience

I'll try this locally before approving

ctrlaltca:
@DarthGandalf : switched to QMediaPlayer, if you want to give it a try

src/modules/snd/libkvisnd.h
```diff
@@ -146,7 +149,10 @@ class KviSoundPlayer : public QObject
 	Phonon::MediaObject * m_pPhononPlayer;
 #endif //!COMPILE_PHONON_SUPPORT
 #ifdef COMPILE_QTMULTIMEDIA_SUPPORT
-	QSoundEffect * m_pSoundEffect;
+	QMediaPlayer * m_pMediaPlayer;
```
DarthGandalf: I think new code shouldn't be using new/delete anymore, but unique_ptr

src/modules/snd/libkvisnd.cpp
```diff
@@ -397,22 +397,26 @@ bool KviSoundPlayer::play(const QString & szFileName)
...
-		    (!KviQString::equalCI(KVI_OPTION_STRING(KviOption_stringSoundSystem), "unknown")) &&
-		    (!KviQString::equalCI(KVI_OPTION_STRING(KviOption_stringSoundSystem), "null")))
-		{
-			qDebug(
-			    "Sound system '%s' is not valid, you may want to re-configure it in the options dialog...",
-			    KVI_OPTION_STRING(KviOption_stringSoundSystem).toUtf8().data());
-			return false; // detection already attempted (and failed?)
-		}
+		qDebug(
+		   "Sound system '%s' is not valid, attempting autodetect...",
```
DarthGandalf, line 378: I've tested it, and it somewhat works. I was able to use "/play" command, but if you go to settings, it shows "null" there in sound system, which is then changed to qt after using /play.

However, this is a different "null" from the one set explicitly - "/play" didn't work after explicit setting of system to null, but it worked (and autodetected qt) when I did /play without explicitly setting system to null first

DarthGandalf: Would it make more sense to do the autodetection on startup instead when the old sound system is invalid? Let's not touch "null" to make it easy to disable sounds altogether

ctrlaltca: > However, this is a different "null" from the one set explicitly - "/play" didn't work after explicit setting of system to null, but it worked (and autodetected qt) when I did /play without explicitly setting system to null first

I guess the first time KviOption_stringSoundSystem was still set to "phonon"m but your kvirc binary didn't support it, so the "phonon" option dialog was missing and the default "null" was selected.
If you set "null" explicitly, kvirc won't force an autodetect.

> Would it make more sense to do the autodetection on startup instead when the old sound system is invalid?

kvirc startup or snd module startup? Both seems unfair to me, because they would force change an option even if the user doesn't want/need to, eg  `echo $snd.isMuted`

src/modules/snd/libkvisnd.h
ctrlaltca: I would prefer a separate commit updating the whole source file code to modern c++ standard

DarthGandalf: How do you imagine writing or even reviewing such commit which gets rid of all the legacy code and replace everything with something more modern?

That's basically rewriting the whole kvirc from scratch :)

Gradually it's at least possible to do

src/modules/snd/libkvisnd.cpp
DarthGandalf: > kvirc startup or snd module startup? Both seems unfair to me, because they would force change an option even if the user doesn't want/need to, eg echo $snd.isMuted

Yeah, good point, not sure.

> so the "phonon" option dialog was missing and the default "null" was selected.

Maybe then make this option more clear between the two "null"s - one is "null", another is "invalid, will autodetect", or something like that?

src/modules/snd/libkvisnd.h
ctrlaltca: Gotcha. Done in [386e46e](https://github.com/kvirc/KVIrc/pull/2669/commits/386e46e3f6ce9d4c0d30ab1a19221e75a2966940)

src/modules/snd/libkvisnd.cpp
ctrlaltca: That would mean adding a fake option to the dropdownlist, adding some code to detect if the user chose that fake option, filtering it out to avoid it getting in the config file, etc..
I see no gain for the user.

DarthGandalf: Ok, what's more confusing?

1) changing the option during `echo $snd.isMuted`

or

2) having a fake null in options which looks indistinguishable from the real null, but behaves differently - this is the current situation

I think I'd prefer 1) over 2)

ctrlaltca: So, I've tried to add fake option to the settings:

immagine

This worked, but pressing the "test" button would call snd.play with an invalid sound system, forcing an autodetect that would play the sound but keep the dummy option selected. So it was even more confusing :)

At the end, in [1b045ba](https://github.com/kvirc/KVIrc/pull/2669/commits/1b045baabfdfd608023b42c022cd693647dbf078) the sound options dialog was changed to force an autodetect if the sound system defined in the options is invalid.

DarthGandalf: Thanks!

DarthGandalf (approved):

PR #2672, opened 2024-07-30
Add support for "Monospace" control code (ASCII 0x11). 

ctrlaltca:
Idling in some channels i received some message that our ircview was not able to cope with.
Turns out these messages contained ASCII 0x11, that's supposed to start a text section using monospace font.
https://modern.ircdocs.horse/formatting#monospace
I guess the idea is to replicate the `preformatted text` used eg. by markdown or html's `<pre>`, `<tt>`, `<code>` tags.

This PR add support for detecting this control code, so that the text get printed correctly, but by now the control code gets ignored.
Adding support for rendering different fonts in KviIrcView would require some heavy work and probably a few new options.

### DIFF
--- a/src/kvilib/irc/KviControlCodes.cpp
+++ b/src/kvilib/irc/KviControlCodes.cpp
@@ -45,6 +45,7 @@ namespace KviControlCodes
 				case KviControlCodes::Underline:
 				case KviControlCodes::Bold:
 				case KviControlCodes::Italic:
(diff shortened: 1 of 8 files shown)


DarthGandalf (approved):
To clarify: there's no way to input this character yet?

ctrlaltca:
> To clarify: there's no way to input this character yet?

Nope, outside of hacks like `$char(17)`

PR #2677, opened 2024-09-07
update windows versions and target windows 10 for Qt6

Dessa:
i have a local change for the included manifest but that is not in here since i have yet to figure out the actual impact.


ctrlaltca:
Looks good to me

PR #2689, opened 2024-11-17
KviShortcut: add new ctor to support  QKeySequence::StandardKey with more than one QKeySequence

### LINKED ISSUE #2685: shift+delete no longer working to 'cut'?
Hullo!

Recently upgraded my system to Windows 11 and installed latest version of KVirc, but I'm finding that "shift+delete" no longer "copies and deletes" like has been standard Windows behaviour since forever.  (This behaviour is documented on https://www.kvirc.net/doc/doc_keyboard.html too)

The shift+delete shortcut still works in other programs I've tested (Chrome, VSCode -- although frustratingly it has also not worked on Facebook for a while now!).

Is there a config option or a fix for this behaviour?

Thank!

ctrlaltca:
On Qt6 the cast from  QKeySequence::StandardKey to QKeySequence will create the shortcut only on the first sequence.
From https://doc.qt.io/qt-6/qkeysequence.html#QKeySequence-1 :
```
The resulting object will be based on the first element in the list of key bindings for the key.
```
Fix #2685

### DIFF
--- a/src/kvilib/core/KviShortcut.h
+++ b/src/kvilib/core/KviShortcut.h
@@ -48,6 +48,7 @@ public:
 #else
 	static QShortcut * create(const char * key, QObject * parent, const char * member = nullptr, const char * ambiguousMember = nullptr, Qt::ShortcutContext context = Qt::WindowShortcut);
 	static QShortcut * create(const QKeySequence & key, QObject * parent, const char * member = nullptr, const char * ambiguousMember = nullptr, Qt::ShortcutContext context = Qt::WindowShortcut);
+	static QShortcut * create(const QKeySequence::StandardKey & key, QObject * parent, const char * member = nullptr, const char * ambiguousMember = nullptr, Qt::ShortcutContext context = Qt::WindowShortcut);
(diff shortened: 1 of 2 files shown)


src/kvilib/core/KviShortcut.h
```diff
@@ -48,6 +48,7 @@ class KVILIB_API KviShortcut
 #else
 	static QShortcut * create(const char * key, QObject * parent, const char * member = nullptr, const char * ambiguousMember = nullptr, Qt::ShortcutContext context = Qt::WindowShortcut);
 	static QShortcut * create(const QKeySequence & key, QObject * parent, const char * member = nullptr, const char * ambiguousMember = nullptr, Qt::ShortcutContext context = Qt::WindowShortcut);
+	static QShortcut * create(const QKeySequence::StandardKey & key, QObject * parent, const char * member = nullptr, const char * ambiguousMember = nullptr, Qt::ShortcutContext context = Qt::WindowShortcut);
```
DarthGandalf, line 51: It's declared but undefined in case of Qt 5?

ctrlaltca: It's defined and implemented only for Qt6. On a second view it seems correct to me, maybe I'm missing something?

DarthGandalf: Right, I misread placement of ifdefs

PR #2700, opened 2025-03-16
Support sending authorization ID in SASL EXTERNAL

DarthGandalf:
The use case is provide network name and client ID to ZNC
See https://github.com/znc/znc/pull/1936


ctrlaltca (approved):
LGTM

PR #2706, opened 2025-07-15
Adapt to Qt 6.9 deprecations

ctrlaltca:


src/modules/objects/KvsObject_pixmap.cpp
```diff
@@ -463,7 +463,20 @@ KVSO_CLASS_FUNCTION(pixmap, mirrored)
 		return false;
 	}
 	m_currentType = Image;
+
+#if (QT_VERSION < QT_VERSION_CHECK(6, 9, 0))
```
DarthGandalf, line 467: As small performance optimization, perhaps skip this altogether, if both directions are false?

ctrlaltca: When an user calls the "mirrored" method on an image he `[[likely]]` wants to flip it either horizontally or vertically. If both parameters are false this is just a noop (still, we could detect this at the beginning in the code and bail out early)

PR #2715, opened 2025-10-04
Fix DBUS notifier arguments

Bartkk0:
Fixes `org.freedesktop.Notifications.Notify()` being called with wrong signedness for notification_id


ctrlaltca:
Hi, thank you for the correction. Anyway, if you want to define an unsigned integer with value 0, you don't need an explicit cast, just use `0u`

ctrlaltca:
Thanks again, merged!

PR #2718, opened 2025-10-10
Fix error handling of QSslSocket

ctrlaltca:
The signal name changed since qt5.

### DIFF
--- a/src/kvilib/net/KviHttpRequest.cpp
+++ b/src/kvilib/net/KviHttpRequest.cpp
@@ -213,7 +213,7 @@ void KviHttpRequest::closeSocket()
 
 	QObject::disconnect(m_p->pSocket, SIGNAL(connected()), this, SLOT(slotSocketConnected()));
 	QObject::disconnect(m_p->pSocket, SIGNAL(disconnected()), this, SLOT(slotSocketDisconnected()));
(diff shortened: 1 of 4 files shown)


src/kvilib/net/KviHttpRequest.cpp
```diff
@@ -213,7 +213,7 @@ void KviHttpRequest::closeSocket()
 
 	QObject::disconnect(m_p->pSocket, SIGNAL(connected()), this, SLOT(slotSocketConnected()));
 	QObject::disconnect(m_p->pSocket, SIGNAL(disconnected()), this, SLOT(slotSocketDisconnected()));
-	QObject::disconnect(m_p->pSocket, SIGNAL(error(QAbstractSocket::SocketError)), this, SLOT(slotSocketError(QAbstractSocket::SocketError)));
+	QObject::disconnect(m_p->pSocket, SIGNAL(errorOccurred(QAbstractSocket::SocketError)), this, SLOT(slotSocketError(QAbstractSocket::SocketError)));
```
DarthGandalf, line 216: doesn't this need #ifdef for the Qt version?

ctrlaltca: Nope, `error()` was used in qt4, `errorOccurred()` is used in qt5 and qt6. This is broken since ages.

PR #2743, opened 2025-12-19
Add a spellchecker backend based on windows >=8 integrated spellchecker

ctrlaltca:
And use it in place of enchant+hunspell on windows build.


src/modules/spellchecker/libkvispellchecker.cpp
```diff
@@ -191,8 +293,18 @@ static bool spellchecker_kvs_reload_dictionaries(KviKvsModuleCommandCall * c)
 
 static bool spellchecker_module_init(KviModule * m)
 {
+#ifdef COMPILE_ENCHANT_SUPPORT
 	g_pEnchantBroker = new enchant::Broker();
 	g_pEnchantDicts = new KviPointerList<enchant::Dict>(/* bAutoDelete = */ false);
+#endif
+
+#ifdef COMPILE_WINSPELLCHECKER_SUPPORT
+	if (FAILED(CoCreateInstance(__uuidof(SpellCheckerFactory), nullptr,
```
DarthGandalf, line 302: I think this won't work on mingw?

ctrlaltca: It should, it's all from standard win32 headers, but honestly I didn't try

ctrlaltca: I can confirm the needed headers are in place in mingw64:
https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/include/spellcheck.h
https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/include/winerror.h#L3382

DarthGandalf: I meant the `__uuidof` operator

But looks like it's supported on mingw too: https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/CAKk_kQMbXBj4W_%3D2bB3%3DPM3WMbQeD%2BeChWopTjJA5k_JCa4M4g%40mail.gmail.com/#msg36395238

Issue #2745, opened 2026-01-06
[Feature Request] load history in buffers


logrusx:
I've been using Quassel up until now and I find it extremely useful it loads the history in the buffer. I'm not sure if it's due to it still being Qt5 and this is a drawing bug or it's designed to work that way but when I scroll up say in a channel after logging in, the history from previous sessions appears.

ctrlaltca:
Settings -> IRC -> Channel -> Paste last channel log
Settings -> IRC -> Query -> Paste last query log

Is this what you are looking for?

logrusx:
Not exactly. I'd like it to be automatic and not limited in time or lines.

DarthGandalf:
That would be https://ircv3.net/specs/extensions/chathistory

PR #2746, opened 2026-01-09
Add support for batch and draft/chathistory capabilities

ctrlaltca:
As suggested in https://github.com/kvirc/KVIrc/issues/2745
This PR will enable the capabilities if supported by the server (use irc.unrealircd.org for tests)
A new `OnBatch` event has been added
Messages will contain a new `@batch` message tag when relevant
No UI/Option have been added right now, but a simple event handler for `OnMeJoin` can make such a request:
```
chathistory LATEST #channel * 50
```


ctrlaltca:
@DarthGandalf can you have a look at this please?

DarthGandalf (commented):
1. I see only batch here, not draft/chathistory
2. it doesn't actually do anything with BATCH other than just not complaining about unknown message?
3. if server sends too many BATCH + with random ids, it'll eventually OOM the client

PR #2758, opened 2026-05-07
Windows spellchecker: avoid crash on double deletion

### LINKED ISSUE #2757: Spell checker dictionaries are no longer loaded in 5.2.10
Using both installers at https://github.com/kvirc/KVIrc/releases/tag/5.2.10, dictionaries are installed to `C:\Program Files\KVIrc\share\hunspell\`, but the _Spell Checker Dictionaries_ list is empty, like #2404.

I also notice that `C:\Program Files\KVIrc\lib\enchant-2\enchant_hunspell.dll` is not loaded, since the file can be deleted while kvirc is running.

On the same system 5.2.8 works fine.

ctrlaltca:
Fix #2757

### DIFF
--- a/src/modules/spellchecker/libkvispellchecker.cpp
+++ b/src/modules/spellchecker/libkvispellchecker.cpp
@@ -237,7 +237,6 @@ static void spellchecker_reload_dicts()
 	while(!g_pWinSpellDicts->isEmpty()) {
 		auto pDict = g_pWinSpellDicts->takeFirst();
 		pDict->Release();
-		delete pDict;
 	}
 #endif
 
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Could ISpellChecker be wrapped with RAII instead?

ctrlaltca:
> Could ISpellChecker be wrapped with RAII instead?

windows COM objects use refcounting, so they delete themselves automatically after the last Release().
Do you mean use a smart pointer? I'm not sure how to mix that without avoiding another double free.

lhmouse:
> > Could ISpellChecker be wrapped with RAII instead?
> 
> windows COM objects use refcounting, so they delete themselves automatically after the last Release(). Do you mean use a smart pointer? I'm not sure how to mix that without avoiding another double free.

There's `CComPtr` and `ComPtr`.

